PR #102, opened 2023-11-15
fix(ci): pipeline logic improvements

c-ryan-k:
This PR fixes some `continue-on-error` and `approval` logic.
- Added `continue-on-error` value passthrough to pre-check workflows (security, tox, code-coverage, azdev_linter)
- tox / code-coverage is set to always continue-on-error (code coverage report shouldn't fail a test)
- Changed "deployment scenario" input strings to individual boolean inputs:
  - [ ] continue release if pre-checks fail
  - [ ] draft github release
  - [ ] upload wheel
- Writes approval confirmation (and deployment steps) to step summary output
- Simplifies conditional input logic in release workflow
---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/.github/workflows/release_workflow.yml
+++ b/.github/workflows/release_workflow.yml
@@ -37,7 +37,8 @@ jobs:
   #   uses: ./.github/workflows/smoke_test.yml
   approval:
     needs: [security, build, test, linter]
-    if: ${{ (success() || inputs.continue-on-fail == 'true') && inputs.release_strategy != 'none' }}
+    # as long as build worked, our break-glass can skip other checks
+    if: github.event.inputs.release_strategy != 'none' && (success() || (needs.build.result == 'success' && github.event.inputs.continue-on-fail == 'true'))
     environment: production
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/release_workflow.yml
```diff
@@ -37,28 +37,29 @@ jobs:
...
             echo "Wheel will be uploaded to storage account."
-          elif [ "${{ inputs.release_strategy }}" == "prod-release"]; then
+          elif [ "${{ inputs.release_strategy }}" == "prod-release" ]; then
             echo "Wheel will be uploaded to storage account after drafting a github release"
           fi
   # occurs when release_strategy is prod-release
   draft_github_release:
     needs: [approval]
-    if: needs.approval.result == 'success' && inputs.release_strategy == 'prod-release'
+    if: success() || (needs.approval.result == 'success' && github.event.inputs.release_strategy == 'prod-release')
```
digimaun, line 57: here `success()` could mean `github.event.inputs.release_strategy == 'upload-wheel'` right?

c-ryan-k: Sorry, this is a complicated one so bear with me 😆 

In this conditional, `success()` allows this step to continue in the happy path scenario (all previous jobs are reporting non-failed, non-cancelled status), which means approval can not be skipped.

One issue I ran into debugging these pipelines was that github treats `if: ${{ condition }}` as `if: ${{ success() && condition }}` if you don't explicitly add one of the pipeline status checks (`success()`, `always()`, `failure()`, etc) in your conditional logic.

So, in this statement, I'm explicitly allowing the step to succeed if everything before it has succeeded, but adding a top-level `OR` to evaluate `approval.result` and `release_strategy` - which otherwise won't evaluate because `success() &&` could be false if the pre-checks fail regardless of the break-glass input 🙃 

https://github.com/orgs/community/discussions/26885#discussioncomment-3253802
https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

> You can use the following status check functions as expressions in if conditionals. A default status check of success() is applied unless you include one of these functions.

digimaun: Here is the scenario I am thinking of, doing an upload-wheel only where draft_github_release should not run.

In the happy path success() should evaluate true, so draft_github_release still runs.

c-ryan-k: Gotcha - understood. Let me pull this one back into draft for a bit, now that we're not doing private index updates perhaps our `inputs` could be more boolean:
 - [ ] bypass failed pre-checks
 - [ ] draft github release
 - [ ] upload wheel to extension storage
 
...rather than scenario-based, which would help clean up the logic a bit and make our options a bit more flexible. It would essentially deconstruct our `upload-wheel` (option 3) and `prod-release` (options 2 and 3) "scenarios" into individual boolean checks, which also gives us the ability to perform a github release without a storage account wheel upload.

digimaun (approved):

PR #103, opened 2023-11-16
feat: --mq-traces in support bundle operation

digimaun:
* This PR introduces the optional `--mq-traces` flag to `az iot ops support create-bundle`.
* This is a convenience/shortcut, as the `stats` command supports both fetching traces by `--trace-ids` or `--trace-dir`. 
* Also smooths out some tcp socket byte handling.

### DIFF
--- a/azext_edge/edge/providers/support/mq.py
+++ b/azext_edge/edge/providers/support/mq.py
@@ -66,10 +66,7 @@ def fetch_diagnostic_traces():
                         }
                     )
 
-        except Exception as e:
-            import pdb
-
-            pdb.set_trace()
+        except Exception:
             logger.debug(f"Unable to process diagnostics pod traces against namespace {namespace}.")
 
     return result
--- a/azext_edge/edge/providers/support/mq.py
+++ b/azext_edge/edge/providers/support/mq.py
@@ -5,14 +5,14 @@
 # ----------------------------------------------------------------------------------------------
 
 from functools import partial
-from typing import Iterable
+from typing import Iterable, Optional
 
 from knack.log import get_logger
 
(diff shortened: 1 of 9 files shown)


azext_edge/edge/providers/support/mq.py
```diff
@@ -47,7 +48,28 @@ def fetch_diagnostic_metrics(namespace: str):
...
+        logger.debug(f"Unable to process diagnostics pod metrics against namespace {namespace}.")
+
+
+def fetch_diagnostic_traces():
+    namespaces = get_mq_namespaces()
+    result = []
+    for namespace in namespaces:
+        try:
+            traces = get_traces(namespace=namespace, trace_ids=["!support_bundle!"])
+            if traces:
```
c-ryan-k, line 60: nit - now that `get_traces` can return either a list of dicts or a list of tuples, do we need to typecheck its return value? Might even just be here, as the other reference to it is a straight `return get_traces(...`

c-ryan-k (approved):

azext_edge/edge/providers/support/mq.py
digimaun: Added test coverage for the return value in the stats tests area.

If the wrong type gets returned, and the later accessing code here causes an exception it should get caught, and also, we have a backup exception handler at the support bundle runner level.

digimaun: Also thinking about the type check it would per list item since the outer most type is either list or None

digimaun: Let's discuss in subsequent PR.

PR #121, opened 2024-01-06
feat: Use ensure_azure_namespace_path() + add placeholder SPC. 

digimaun:
- When importing azure namespace packages, ensures extension packages are first on namespace pkg path and sys.path.
- Add placeholder SPC for `aio-opc-ua-broker-issuer-list`.

### DIFF
--- a/azext_edge/edge/util/common.py
+++ b/azext_edge/edge/util/common.py
@@ -142,13 +142,14 @@ def url_safe_hash_phrase(phrase: str):
 
 
 def ensure_azure_namespace_path():
-    import os
-    import sys
-
     """
     Run prior to importing azure namespace packages (azure.*) to ensure the
     extension root path is configured for package import.
     """
+
+    import os
+    import sys
+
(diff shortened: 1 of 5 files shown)


c-ryan-k (approved):
:shipit:

azext_edge/edge/util/common.py
```diff
@@ -139,3 +139,34 @@ def url_safe_hash_phrase(phrase: str):
...
+        if getattr(azure, "__path__", None) and ext_azure_dir not in azure.__path__:  # _NamespacePath /w PEP420
+            if isinstance(azure.__path__, list):
+                azure.__path__.insert(0, ext_azure_dir)
+            else:
+                azure.__path__.append(ext_azure_dir)
+
+    if sys.path and sys.path[0] != ext_path:
+        sys.path.insert(0, ext_path)
+
+    return
```
vilit1, line 172: nit - no need for return. Also awkward spacing/ordering of imports vs function description

vilit1 (approved):

PR #122, opened 2024-01-08
feat: Apply resource name filter to edge service checks

Elsie4ever:
This PR contains:
1. Apply resource name filter to existed edge service checks
    - The name filter can have wildcard, case insensitive
    - Note: when there is name filter applied, the check status will be `skipped`

3. Update the behavior of check evaluation to be consistent across services when there is no resource found
image

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/check/base.py
```diff
@@ -73,7 +75,43 @@ def check_post_deployment(
...
+    if resource_name:
+        resource_name = resource_name.lower()
+        # check resource for wildcarded name
+        if "*" in resource_name:
+            resource_name = resource_name.replace("*", ".*")
+            # add beginning and end of string regex
+            resource_name = f"^{resource_name}$"
+            resources = [resource for resource in resources if re.match(resource_name, get_resource_name(resource))]
+        else:
+            resources = [resource for resource in resources if resource_name == get_resource_name(resource)]
```
c-ryan-k, line 101: I did a bit of testing, and I think we can improve performance here for wildcard searches with a small tweak.
We can use `re.compile()` to speed up multiple regex matches (so you don't need to compile the regex again for each item in the list):
```suggestion
        if "*" in resource_name:
            regex = re.compile(f"^{resource_name.replace('*', '.*')}$")
            resources = [resource for resource in resources if regex.match(get_resource_name(resource))]
        else:
            resources = [resource for resource in resources if resource_name == get_resource_name(resource)]
```

azext_edge/edge/providers/check/lnm.py
```diff
@@ -302,13 +314,24 @@ def _process_lnm_pods(
...
     pods = get_namespaced_pods_by_prefix(prefix=prefix, namespace=namespace, label_selector=label_selector)
 
+    if resource_name:
+        pods = filter_resources_by_name(resources=pods, resource_name=resource_name)
+
+        if not pods:
+            check_manager.add_target(target_name=target)
+            check_manager.add_display(target_name=target, display=Padding("Unable to fetch pods.", (0, 0, 0, padding + 2)))
+
+            return pods
```
c-ryan-k, line 331: If the return type is `None` - no need to return a value.
Even if it looks like it's `None`, it could be `[]`, we can just use `return` to ensure no value is returned.

Edit: it looks like further down this method is also returning `pods` - is the return type incorrect?
Should we get the pods first and then pass them into this method to process them, or is there a reason it's fetching them and returning them?

azext_edge/edge/providers/check/lnm.py
```diff
@@ -302,13 +314,24 @@ def _process_lnm_pods(
     conditions: Optional[List[str]] = None,
     namespace: Optional[str] = None,
     detail_level: int = ResourceOutputDetailLevel.summary.value,
+    resource_name: str = None,
 ) -> None:
     def _get_lnm_pods_namespace(pod: V1Pod) -> str:
```
c-ryan-k, line 319: Similar comments elsewhere, but this could be a common method to avoid re-implementing it everywhere

azext_edge/edge/providers/check/mq.py
```diff
@@ -279,15 +287,21 @@ def evaluate_broker_listeners(
         "status",
     ]
 
-    all_listeners = MQ_ACTIVE_API.get_resources(MqResourceKinds.BROKER_LISTENER).get("items", [])
+    # all_listeners = MQ_ACTIVE_API.get_resources(MqResourceKinds.BROKER_LISTENER).get("items", [])
```
c-ryan-k, line 290: commented code - safe to remove?

c-ryan-k (commented):
Chunk of review

azext_edge/edge/providers/check/base.py
```diff
@@ -73,7 +75,43 @@ def check_post_deployment(
...
+def filter_resources_by_name(
+    resources: List[dict],
+    resource_name: str,
+) -> List[dict]:
+    # decide if to use resource["metadata"]["name"] or resource.metadata.name
+    def get_resource_name(resource: dict) -> str:
+        # get lowercase name from resource
+        if isinstance(resource, dict):
+            return resource.get("metadata", {}).get("name", "").lower()
+        return resource.metadata.name.lower()
```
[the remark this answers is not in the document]
Elsie4ever: yep I will! this happened before the akri pr change😁

azext_edge/edge/providers/check/dataprocessor.py
```diff
@@ -71,36 +74,35 @@ def check_dataprocessor_deployment(
...
+    resource_name: str = None,
 ) -> Dict[str, Any]:
     check_manager = CheckManager(check_name="evalInstances", check_desc="Evaluate Data processor instance")
 
     instance_namespace_conditions = ["len(instances)==1", "provisioningStatus"]
-    instance_all_conditions = ["instances"]
 
-    instance_resources: dict = DATA_PROCESSOR_API_V1.get_resources(DataProcessorResourceKinds.INSTANCE)
     target_instances = generate_target_resource_name(api_info=DATA_PROCESSOR_API_V1, resource_kind=DataProcessorResourceKinds.INSTANCE.value)
-    check_manager.add_target(target_name=target_instances, conditions=instance_all_conditions)
```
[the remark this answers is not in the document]
Elsie4ever: yeah, I was updating the to keep consistent with mq and other services as well, the add_target for _all_ namespace only needed when `if not all_instances` triggers, all others will be add_target with namespaces

azext_edge/edge/providers/check/lnm.py
Elsie4ever: yep it should be gone once I finish fixing the merge conflicts😋

azext_edge/edge/providers/check/cloud_connectors.py
```diff
@@ -56,6 +58,31 @@ def process_cloud_connector(
...
+        excluded_connector_names = [
+            get_resource_name(connector) for connector in excluded_connectors
+        ]
+
+        all_topic_maps = [
+            map
+            for map in all_topic_maps
+            if map.get("spec", {}).get(topic_map_reference_key)
+            not in excluded_connector_names
+        ]
```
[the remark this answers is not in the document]
Elsie4ever: Thanks this is really valuable info, will think about it more!

azext_edge/edge/providers/check/lnm.py
Elsie4ever: thanks for the good catch! there is no need to return for anything😋

azext_edge/edge/providers/check/cloud_connectors.py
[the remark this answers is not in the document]
Elsie4ever: Hi Ryan, now I updated to filter with both name and namespace info, let me know if you have more concern, thanks!

azext_edge/edge/providers/check/cloud_connectors.py
```diff
@@ -64,7 +91,7 @@ def process_cloud_connector(
             message=f"No {connector_display_name} resources detected",
             padding=connector_padding,
         )
-        if detail_level != ResourceOutputDetailLevel.summary.value:
+        if detail_level != ResourceOutputDetailLevel.summary.value and not connector_resource_name:
```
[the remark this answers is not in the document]
Elsie4ever: after changing I think this still cover the case? we only show invalid topic maps with no connector found when there is no name filter apply?

azext_edge/edge/providers/check/cloud_connectors.py
```diff
@@ -56,6 +58,38 @@ def process_cloud_connector(
...
+        # filter out topic maps with both excluded connector name and namespace
+        all_topic_maps = [
+            map
+            for map in all_topic_maps
+            if (
+                map.get("spec", {}).get(topic_map_reference_key),
+                get_resource_metadata_property(map, prop_name="namespace"),
+            )
+            not in excluded_connector_properties
+        ]
```
c-ryan-k, line 91: Can we add some unit tests specifically for this chunk of logic?
I followed through with some sample data and it looks like it's including / excluding the right things, but would be great to have a test where we can throw a bunch of connector names, namespaces, and topic maps into a function and make sure we get the right ones out at the other end.

azext_edge/edge/providers/check/base.py
```diff
@@ -701,30 +731,25 @@ def add_display_and_eval(
...
-    return resource.get("metadata", {}).get("name")
-
-
-def get_pod_namespace(pod: V1Pod) -> Union[str, None]:
-    return pod.metadata.namespace
+# get either name or namespace from resource that might be a object or a dict
+def get_resource_metadata_property(resource: Union[dict, Any], prop_name: str) -> Union[str, None]:
+    if isinstance(resource, dict):
+        return resource.get("metadata", {}).get(prop_name)
+    return getattr(resource.metadata, prop_name, None)
```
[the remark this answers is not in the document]
Elsie4ever: got this, I changed to `hasattr(resource, "metadata")` since object like V1Pod is not iteratable

azext_edge/edge/providers/check/cloud_connectors.py
Elsie4ever: thanks for testing! I will update with unit test

Elsie4ever: Hi Ryan, in order to test this logic I put them in a new function, since eval_function only return things in display manager and does not have detail of topic maps, please let me know if there's more concern to the tests!

azext_edge/edge/providers/check/base.py
```diff
@@ -735,21 +735,20 @@ def add_display_and_eval(
...
         return resource.get("metadata", {}).get(prop_name)
-    return getattr(resource.metadata, prop_name, None)
+    return getattr(resource.metadata, prop_name, None) if hasattr(resource, "metadata") else None
 
 
-def resources_grouped_by_namespace(resources: List[dict]):
-    resources.sort(key=lambda resource: get_resource_metadata_property(resource, prop_name="namespace"))
-    return groupby(resources, lambda resource: get_resource_metadata_property(resource, prop_name="namespace"))
+def key_func(resource):
+    return get_resource_metadata_property(resource, prop_name="namespace")
```
c-ryan-k, line 742: My previous comment was assuming we'd define the `key_func` (just a suggested name) inside the `x_grouped_by_y` function:

```
def resources_grouped_by_namespace(resources):
    def by_namespace(item):
        return item.namespace
    resources.sort(key=by_namespace)
    ...
```

But if there are multiple functions that use this, we should keep it at the top level but rename it something a little more readable like "get_namespace"

azext_edge/edge/providers/check/cloud_connectors.py
```diff
@@ -219,6 +195,40 @@ def process_cloud_connector(
     return check_manager.as_dict(as_list)
 
 
+def filter_connector_and_topic_map_resources(
+    connector_resource_name: str,
+    topic_map_reference_key: str,
+    all_connectors: List[Dict[str, Any]],
+    all_topic_maps: List[Dict[str, Any]],
+):
```
c-ryan-k, line 203: nit: typing - especially for something with a complex return type like `tuple[list, list]`

azext_edge/edge/params.py
```diff
@@ -135,7 +135,14 @@ def load_iotops_arguments(self, _):
             "Choose 0 for a summary view (minimal output), "
             "1 for a detailed view (more comprehensive information), "
             "or 2 for a verbose view (all available information).",
-        )
+        ),
+        context.argument(
+            "resource_name",
+            options_list=["--resource-name", "--rn"],
```
vilit1, line 141: this param should have filter in its name like: `--resource-name-filter` and `--rnf`

Otherwise, a customer could use this param (not read the param help) and then assume that other params should support filtering too

Elsie4ever: yeah good point, will update!

azext_edge/edge/providers/check/base.py
Elsie4ever: got it! I moved it upper level since the function defined inside is a bit repetitive, will update the name!

azext_edge/edge/params.py
Elsie4ever: actually after carefully thinking, I don't think filter is necessary since all other params in this command is actually working as filters as well, it's a bit strange to only add filter in the param?

azext_edge/edge/providers/check/base.py
```diff
@@ -86,19 +87,24 @@ def filter_resources_by_name(
...
+    # if only alphanumeric characters and hyphens, check resource for exact name
+    if "?" not in resource_name and "*" not in resource_name:
+        resources = [resource for resource in resources if resource_name == get_resource_metadata_property(
+            resource, prop_name="name").lower()
+        ]
+    else:
+        regex = re.compile(f"{resource_name.replace('*', '.*')}")
+        resources = [resource for resource in resources if regex.fullmatch(
+            get_resource_metadata_property(resource, prop_name="name").lower()
+        )]
```
c-ryan-k, line 108: If we're already ensuring that the filter name does *not* include anything except alphanumeric, `*`, and `?` - we don't need to split the cases based on whether they have a glob pattern in them or not.

We can use the [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) module to match glob-style strings to run `fnmatch(name, pattern)` for case-insensitive matches, or `fnmatchcase` for case-sensitive matching:

```
from fnmatch import fnmatch
resources = [
    resource
    for resource in resources
    if fnmatch(get_resource_metadata_property(resource, prop_name="name"), resource_name)
]
```

Elsie4ever: got it, will test it out!

azext_edge/edge/providers/check/base.py
```diff
@@ -86,19 +87,24 @@ def filter_resources_by_name(
...
-        else:
-            resources = [resource for resource in resources if resource_name == get_resource_metadata_property(
-                resource, prop_name="name").lower()
-            ]
+    if not resource_name:
+        return resources
+    
+    # validate resource_name that should only contain alphanumeric characters, hyphens, ? and *
+    if not re.fullmatch(r"[a-zA-Z0-9\-?*]+", resource_name):
+        raise InvalidArgumentValueError("Invalid resource name. Only alphanumeric characters, hyphens, ? and * are allowed.")
```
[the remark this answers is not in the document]
Elsie4ever: I see, so in this case I should add a validate resource name method in [_validators.py](https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/_validators.py) and do validator=validate_resource_name in param.py?

c-ryan-k (approved):

PR #146, opened 2024-02-21
feat: Update notification

digimaun:
* Pattern to notify users of update availability when using `az iot ops *` commands.
* `init` as the basis of compatibility with any other commands, has the new option `--ensure-latest` which will throw an error if a new CLI version exists. This is to support users who want to make sure they are on the latest.

### DIFF
--- a/azext_edge/__init__.py
+++ b/azext_edge/__init__.py
@@ -6,20 +6,38 @@
 
 from azure.cli.core import AzCommandsLoader
 from azext_edge.constants import VERSION
+from knack.events import EVENT_INVOKER_POST_PARSE_ARGS
+
+
+def version_check_handler(cli_ctx, **kwargs):
+    from .edge.util.version_check import check_latest
+
+    command: str = kwargs.get("command")
+    if command:
+        if command.startswith("iot ops"):
+            if command == "iot ops init":
+                ensure_latest = kwargs["args"].ensure_latest
+                check_latest(cli_ctx=cli_ctx, force_refresh=ensure_latest, throw_if_upgrade=ensure_latest)
+            else:
+                check_latest(cli_ctx)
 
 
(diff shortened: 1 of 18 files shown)


azext_edge/__init__.py
```diff
@@ -6,20 +6,38 @@
...
+
+
+def version_check_handler(cli_ctx, **kwargs):
+    from .edge.util.version_check import check_latest
+
+    command: str = kwargs.get("command")
+    if command:
+        if command.startswith("iot ops"):
+            if command == "iot ops init":
+                ensure_latest = kwargs["args"].ensure_latest
```
c-ryan-k, line 19: Just a thought - should we also check `'ensure_latest' in kwargs['args']` before this accessor, in case we rename/move/remove this param in `params.py`?

c-ryan-k (approved):

azext_edge/__init__.py
```diff
@@ -6,20 +6,38 @@
...
+from knack.events import EVENT_INVOKER_POST_PARSE_ARGS
+
+
+def version_check_handler(cli_ctx, **kwargs):
+    from .edge.util.version_check import check_latest
+
+    command: str = kwargs.get("command")
+    if command:
+        if command.startswith("iot ops"):
+            if command == "iot ops init":
```
vilit1, line 18: I thought we were supposed to care if a person has an old deployment too (as in encourage them to upgrade + redo init)... if so then wouldn't spamming this check for every iot ops call be better? Or does the service have an answer to that already?

vilit1 (approved):

azext_edge/__init__.py
digimaun: Good thought. I think we should punt for a subsequent iteration. As you mentioned in another setting, init should should throw if the arg changes.

PR #156, opened 2024-03-01
fix: make `az iot ops check` find nested errors

vilit1:
image

Also found a nested error (it resolved itself by now) so added some flavor color:
image

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/providers/check/base.py
+++ b/azext_edge/edge/providers/check/base.py
@@ -793,7 +793,14 @@ def process_dict_resource(
                     namespace=namespace,
                     display=Padding(display_text, (0, 0, 0, padding))
                 )
-                display_text = f"[cyan]{value}[/cyan]"
+                if "error" in key.lower():
+                    display_text = f"[red]{value}[/red]"
+                    check_manager.set_target_status(
+                        target_name=target_name,
+                        status=CheckTaskStatus.error.value
+                    )
+                else:
+                    display_text = f"[cyan]{value}[/cyan]"
                 check_manager.add_display(
                     target_name=target_name,
                     namespace=namespace,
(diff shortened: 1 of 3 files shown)


azext_edge/edge/providers/check/base.py
```diff
@@ -785,29 +785,43 @@ def process_dict_resource(
...
+                value=value
+            )
+            check_manager.add_display(
+                target_name=target_name,
+                namespace=namespace,
+                display=Padding(display_text, (0, 0, 0, value_padding))
+            )
+
+
+def process_value_color(
```
Elsie4ever, line 811: The only concern I can think of so far for report error when key contains "error" text in `process_dict_resource` is: would there be possibility that key contains "error" that is not actually error? for example if there is a kvp like `"error_logging_disabled": True`. But again, this is a good solution for capturing nested error since we never know when error can happen and in **this specific case**, there is no hint in CRD that an error can happen here

vilit1: For now, we can assume that if there is an error in the key, then it is talking about an error message more often than not. We can always add in exceptions (ex: value is NoError means that this is not an error).

I have made this into a helper function so it is easier to track and edit.

azext_edge/edge/providers/check/base.py
```diff
@@ -785,29 +785,43 @@ def process_dict_resource(
...
+    check_manager: CheckManager,
+    target_name: str,
+    key: Any,
+    value: Any,
+) -> str:
+    value = value if value else "N/A"
+    if "error" in str(key).lower() and str(value).lower() not in ["null", "n/a", "none", "noerror"]:
```
c-ryan-k, line 818: Are these keys all service-defined? Or is there a chance a user would define a custom key that uses the word `error` for whatever reason?

vilit1: I am assuming they are service defined... 

If a user is modifying the service (that it gets caught in the check) then I would say this can at least show them they are doing wacky things

digimaun (approved):
Approved based off our internal discussion.

PR #170, opened 2024-03-18
refactor: follow up on support bundle

Elsie4ever:
1. Investigation for init container: have not found specific documentation state that init container do not have previous instance log, but I'm also not able to capture any previous log even init container restarts multiple times, will get error like "Error from server (BadRequest): previous terminated container "init-reader" in pod "aio-dp-reader-worker-0" not found". Therefore just in case, I will still try capture the previous log, but it's guarded with exception handler
2. Refactored to capture init container and rest of the containers in one method.
3. Fix the issue that add_pod_to_mocked_pods was not applied when call
4. Remove code that set capture_previous_logs=True, since now the function is default to True

Verified support bundle still capture the same resources as before
---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/tests/edge/support/conftest.py
```diff
@@ -31,17 +31,21 @@ def add_pod_to_mocked_pods(
...
+            expected_pod_map[namespace][pod_name].update({container_name: mock_log})
+        else:
+            expected_pod_map[namespace][pod_name] = {container_name: mock_log}
 
     pods_list = V1PodList(items=pod_list)
     mocked_client.CoreV1Api().list_pod_for_all_namespaces.return_value = pods_list
     mocked_client.CoreV1Api().read_namespaced_pod_log.return_value = mock_log
 
-    yield expected_pod_map
+    return expected_pod_map
```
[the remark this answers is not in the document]
Elsie4ever: oh I see, even deleting the return will have the same effect to auto apply the update to mock_pod_list?

azext_edge/tests/edge/support/conftest.py
```diff
@@ -15,12 +15,12 @@ def add_pod_to_mocked_pods(
...
     expected_pod_map,
     mock_names: List[str] = None,
-    mock_init_containers: bool = False
+    mock_init_container: bool = False
 ):
     from kubernetes.client.models import V1PodList, V1Pod, V1PodSpec, V1ObjectMeta, V1Container
 
     current_pods = mocked_client.CoreV1Api().list_pod_for_all_namespaces.return_value
-    pod_list = current_pods.items()
+    pod_list = current_pods.items
```
vilit1, line 23: is current_pods supposed to be a dict? make sure that items does not need the ()

azext_edge/tests/edge/support/conftest.py
```diff
@@ -31,17 +31,21 @@ def add_pod_to_mocked_pods(
         spec = V1PodSpec(containers=[V1Container(name=container_name)])
         pod = V1Pod(metadata=V1ObjectMeta(namespace=namespace, name=pod_name), spec=spec)
 
-        if mock_init_containers:
+        if mock_init_container:
             pod.spec.init_containers = [V1Container(name="mock-init-container")]
-            expected_pod_map[namespace][pod_name]["mock-init-container"] = mock_log
+            expected_pod_map[namespace][pod_name] = {"mock-init-container": mock_log}
```
vilit1, line 36: make sure you are not losing values here (since this is overwriting a dict rather than value

azext_edge/tests/edge/support/conftest.py
Elsie4ever: yep. items() actually report errors when called(the function wasn't triggered before so didn't found the error), V1pod only has items property

azext_edge/tests/edge/support/conftest.py
```diff
@@ -31,17 +31,21 @@ def add_pod_to_mocked_pods(
...
         pod_list.append(pod)
-        expected_pod_map[namespace][pod_name] = {container_name: mock_log}
+
+        if pod_name in expected_pod_map[namespace]:
+            expected_pod_map[namespace][pod_name].update({container_name: mock_log})
```
vilit1, line 40: if it is only one key, then use the [] notation. Update is more for updating multiple keys in a dict

azext_edge/tests/edge/support/test_support_unit.py
```diff
@@ -460,12 +461,24 @@ def assert_list_pods(
 
     for namespace in mocked_list_pods:
         for pod_name in mocked_list_pods[namespace]:
-            for container_name in mocked_list_pods[namespace][pod_name]:
+            init_data = ""
+            import copy
```
vilit1, line 465: honestly just push this import to the top of the file

azext_edge/tests/edge/support/conftest.py
Elsie4ever: yeah I double checked, at this point pod_name under expected_pod_map[namespace] is not yet exists

azext_edge/edge/providers/support/base.py
```diff
@@ -494,37 +470,45 @@ def default_bundle_name(system_name: str) -> str:
...
-            )
-        except ApiException as e:
-            logger.debug(e.body)
+    capture_previous_log_runs = [False]
+
+    if capture_previous_logs:
+        capture_previous_log_runs.append(True)
+
+    for container in pod_containers:
+        for capture_previous in capture_previous_log_runs:
```
vilit1, line 490: so here you check each container twice if capture previous log is true?

Elsie4ever: right, we would need to do that because setting previous to True and False will get different log traces and they are not overlapping

azext_edge/edge/providers/support/base.py
```diff
@@ -47,10 +47,10 @@ def process_crd(group: str, version: str, kind: str, plural: str, api_moniker: s
 
 def process_v1_pods(
     resource_api: EdgeResourceApi,
-    label_selector=None,
+    capture_previous_logs: bool = True,
+    include_metrics: bool = False,
     since_seconds: int = 60 * 60 * 24,
-    include_metrics=False,
-    capture_previous_logs=True,
+    label_selector: Optional[str] = None,
```
vilit1, line 53: nit: be careful when modifying positions of arguments since it could affect anything that calls it without specifying the exact param

azext_edge/tests/edge/support/conftest.py
Elsie4ever: I see, will update!

azext_edge/edge/providers/support/base.py
vilit1: and there is no way to get everything in one call? sadness...

azext_edge/tests/edge/support/conftest.py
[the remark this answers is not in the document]
vilit1: since you are modifying the object - yes it will. If you want to google a term, search fruitful functions (idk what the alternative is called - void?)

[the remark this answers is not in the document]
Elsie4ever: looks like it works! will update thanks!

azext_edge/edge/providers/support/base.py
Elsie4ever: Yeah that's right, it's sad when I read this from their doc😥

azext_edge/edge/providers/support/base.py
Elsie4ever: yeah but luckily all the calls have passed with param name!

azext_edge/edge/providers/support/base.py
```diff
@@ -494,37 +470,45 @@ def default_bundle_name(system_name: str) -> str:
...
+    capture_previous_log_runs = [False]
+
+    if capture_previous_logs:
+        capture_previous_log_runs.append(True)
+
+    for container in pod_containers:
+        for capture_previous in capture_previous_log_runs:
+            try:
+                logger_debug_previous = "previous run " if capture_previous else ""
+                logger.debug(f"Reading {logger_debug_previous} log from pod {pod_name} container {container.name}")
```
digimaun, line 493: There is going to be an extra space added to the debug message because a space was added between
"{logger_debug_previous}" and "log".

azext_edge/edge/providers/support/base.py
```diff
@@ -47,10 +47,10 @@ def process_crd(group: str, version: str, kind: str, plural: str, api_moniker: s
 
 def process_v1_pods(
     resource_api: EdgeResourceApi,
-    label_selector=None,
+    capture_previous_logs: bool = True,
+    include_metrics: bool = False,
     since_seconds: int = 60 * 60 * 24,
```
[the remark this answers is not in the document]
Elsie4ever: updated and removed the TODO

azext_edge/edge/providers/support/base.py
Elsie4ever: updated

digimaun (approved):
When CI passes

PR #172, opened 2024-03-21
feat: init supports deployment from file template

digimaun:
* Not intended for general usage, but advanced scenarios.
* Refactor util.read_file_content to be more modern + tests.
* Issues with pattern that will be resolved when TODO of decoupling template builder from `work`.
* Removed and replaced `generate_generic_id` with `generate_random_string`.
* Fixes an issue with support bundle test_mq_list_stateful_sets producing extraneous artifacts.


azext_edge/tests/generators.py
```diff
@@ -23,3 +25,8 @@ def generate_names(prefix: str = "", count: int = 1, max_length: int = 48) -> Un
 
 def get_zeroed_subscription() -> str:
     return "00000000-0000-0000-0000-000000000000"
+
+
+def generate_random_string(size: int):
```
vilit1, line 30: is the generate_id no longer enough?

azext_edge/tests/edge/init/custom_template.json
```diff
@@ -0,0 +1,1050 @@
+{
```
vilit1, line 1: is there a reason for saving a template rather than dumping one via az iot ops init and using that? (and deleting it at the end, you can move my delete all file created fixture wherever you need it)

azext_edge/edge/params.py
```diff
@@ -530,6 +530,11 @@ def load_iotops_arguments(self, _):
             arg_group="TLS",
             type=int,
         )
+        context.argument(
+            "template_path",
+            options_list=["--template-file"],
+            help="The path to a custom IoT Operations deployment template. Intended for advanced use cases.",
```
vilit1, line 536: no parameter file option?

digimaun: Didn't want to include a separate parameter from file option because CLI parameters are what should be used. Also, you can bake custom parameter values in the template rather than a separate param file.

vilit1: Wanted to make sure cause there may be demand (especially with how portal defaults to generating a template and param file when you export templates)
image

vilit1: Also please include a help example demonstrating at least the template param with the other CLI inputs (you can add a note that if the file already has the params, then you can omit them)

azext_edge/tests/edge/init/custom_template.json
digimaun: One use is a convenient non-standard template for smoke test/user test. But doesn't have to be.

azext_edge/edge/params.py
digimaun: My concern with adding an example, is we don't want to promote `--template-file` because in most cases a user should deploy with the built-in template. `--template-file` is intended for advanced use cases primarily internally.

azext_edge/tests/generators.py
digimaun: generate_id produces a uuid str to convey an Id like value with fixed size.  

generate_random_string is meant to produce a string of chosen length with wider range of characters (which can be further customized)

generate_id could be refactored to have variable length but then the logic would have to concatenate chunks and trim the extra + be more limited in character range.

c-ryan-k (approved):

vilit1 (approved):

azext_edge/tests/edge/init/custom_template.json
digimaun: Going to generate.

PR #182, opened 2024-04-09
feat: add general node checks for `az iot ops check` pre-requisite checks

vilit1:
**### make sure to credit ryan for POC**

image
Some minor refactoring:
- adding maps to enums
- moving functions out of base.py to a folder (base) with file names to categorize these functions
- renaming of some to make more sense (from a quick look perspective)
- COLOR_STR_FORMAT

Added general node checks (and removed outdated one):
- try to have one node (ok if more but warn)
-  at least 4 cpus per node
- 30 g of storage
- 16 g of memory

_make sure to credit ryan  for POC_

Future tasks (see added the TODO's)
- additional refactoring
   - padding
   - add unit tests for other base funcs

Visual examples:
Success:
image

Errors + Warnings:
image

`make sure to credit ryan  for POC`

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/check/common.py
```diff
@@ -327,4 +327,16 @@ class DataLakeConnectorTargetType(ListableEnum):
 
 MIN_K8S_VERSION = "1.20"
 
+# Node prerequisite constants
+
+MIN_NODE_MEMORY = "16G"
+MIN_NODE_STORAGE = "30G"
+MIN_NODE_VCPU = "4"
+AIO_SUPPORTED_ARCHITECTURES = ["amd64", "arm64"]
```
digimaun, line 335: arm64? Can you share the reference that arm64 is supported ?

vilit1: wat... ok will change as more comments come in

azext_edge/edge/providers/check/common.py
```diff
@@ -327,4 +327,16 @@ class DataLakeConnectorTargetType(ListableEnum):
 
 MIN_K8S_VERSION = "1.20"
 
+# Node prerequisite constants
+
+MIN_NODE_MEMORY = "16G"
```
Elsie4ever, line 332: out of curiosity, where did we find these requirements value?😋

vilit1: The wonderful work item (where we needed to make sure we do not need a machine to run COD)

azext_edge/edge/common.py
```diff
@@ -37,10 +61,28 @@ class ResourceState(Enum):
...
+    n_a = "N/A"
+
+    @classmethod
+    def map_to_color(cls, value) -> str:
+        return cls.map_to_status(value).color
+
+    @classmethod
+    def map_to_status(cls, value) -> CheckTaskStatus:
+        status_map = {
+            cls.starting.value: CheckTaskStatus.warning,
```
Elsie4ever, line 78: what about running?

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -0,0 +1,128 @@
...
+from rich.padding import Padding
+
+from .check_manager import CheckManager
+from .nodes import check_nodes
+from .resources import enumerate_ops_service_resources
+from ..common import CoreServiceResourceKinds, ResourceOutputDetailLevel
+from ....common import CheckTaskStatus, ListableEnum
+
+from ....providers.edge_api import EdgeResourceApi
+
```
Elsie4ever, line 22: wondering can we remove some lines in the import? or are these for grouping?

azext_edge/edge/common.py
vilit1: everything else gets caught into the get default aka success

azext_edge/edge/providers/check/base/deployment.py
vilit1: wdym - everything is used here so you need to import it (and it is better to import from the source)

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -0,0 +1,128 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
```
Elsie4ever, line 6: I'm thinking of the naming of the new files under this base folder... pods.py, nodes.py are plurals, therefore should we keep consistent to have plural on deployment.py and display.py as well?🤔

azext_edge/edge/providers/check/base/display.py
```diff
@@ -0,0 +1,145 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+from typing import Any, Dict, List, Optional, Tuple
+
```
Elsie4ever, line 8: I feel like the functions under this file is basically used for check manager, what do you think of combining this file with the check_manager.py?

vilit1: I want to keep the check manager class separate here from the display functionality

azext_edge/edge/providers/check/base/deployment.py
vilit1: I can rearrange it a bit

azext_edge/edge/providers/check/base/deployment.py
vilit1: well not sure if displays and deployments would make sense here since the idea was that there are multiple pods, (possibly) nodes, resources, etc while one (AIO) deployment, one display, one check manager

vilit1: I will just make everything singular

azext_edge/edge/providers/check/base/deployment.py
Elsie4ever: kk got it!

azext_edge/edge/providers/check/base/nodes.py
```diff
@@ -0,0 +1,162 @@
...
+        for condition, expected, actual in [
+            (
+                "info.architecture",
+                AIO_SUPPORTED_ARCHITECTURES,
+                node.status.node_info.architecture,
+            ),
+            (
+                "condition.cpu",
+                MIN_NODE_VCPU,
+                parse_quantity(node.status.capacity.get("cpu", 0)),
```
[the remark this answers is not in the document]
vilit1: it is part of the nodev1 class so unless something went horribly wrong, it should be there

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -0,0 +1,129 @@
...
+from ...base import client
+from ....common import CheckTaskStatus, ListableEnum
+from ....providers.edge_api import EdgeResourceApi
+
+
+logger = get_logger(__name__)
+# TODO: unit test
+
+
+def check_pre_deployment(
```
c-ryan-k, line 27: Just a general note - we should be able to be "smart" about when this is called in the check function, based on whether the target cluster already has an AIO deployment.

If the user doesn't have an AIO deployment, we should run this check regardless of whether they specified `--pre` (unless, maybe they set `--pre false`). If they do already have AIO deployed, we can skip this check unless they specify `--pre`

azext_edge/edge/providers/checks.py
```diff
@@ -71,7 +71,7 @@ def run_checks(
             )
```
digimaun, line 71: We discussed the following behavior changes:
  - If AIO is not deployed on the cluster, pre deployment checks run by default.
  - If AIO is deployed on the cluster, pre deployment checks do not run by default.
    - A user can still run pre-deployment checks at anytime using `--pre`.

azext_edge/edge/providers/check/base/deployment.py
vilit1: currently, it will run pre all the time. I agree with having pre run only if AIO is not installed or pre is explicitly called. 

Will change

azext_edge/edge/providers/check/base/node.py
```diff
@@ -0,0 +1,162 @@
...
+
+logger = get_logger(__name__)
+
+
+def check_nodes(as_list: bool = False) -> Dict[str, Any]:
+    from ...base import client
+    check_manager = CheckManager(check_name="evalClusterNodes", check_desc="Evaluate cluster nodes")
+    padding = (0, 0, 0, 8)
+    target = "cluster/nodes"
+    check_manager.add_target(target_name=target, conditions=["len(cluster/nodes)>=1"])
```
c-ryan-k, line 35: This might just be a semantic nit but is our target `>= 1`? If we warn for having more than one node, maybe our target should be `==1`

azext_edge/edge/providers/check/base/user_strings.py
```diff
@@ -0,0 +1,11 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+CONNECTIVITY_ERROR = "{0} Is there connectivity to the cluster?"
+MULTINODE_CLUSTER_MSG = "Currently, only single-node clusters are officially supported for AIO deployments"
+NO_NODES_MSG = "No nodes detected."
+UNABLE_TO_DETERMINE_VERSION_MSG = CONNECTIVITY_ERROR.format("Unable to determine.")
```
c-ryan-k, line 10: Does this need context like `Unable to determine version.` ?

azext_edge/edge/providers/checks.py
vilit1: currently, it will run pre all the time. I agree with having pre run only if AIO is not installed or pre is explicitly called.

Will change

azext_edge/edge/providers/check/base/user_strings.py
vilit1: yes

azext_edge/edge/providers/check/base/node.py
vilit1: I thought we should still continue with checks if there is more than one node... plus a warning isn't a "fail" so the condition is still successful? maybe?

vilit1:
Pre check change following this table:
image
Note that AIO pre means precheck only if AIO is deployed

### No AIO Deployments
nothing specified (run both)
image

run only pre
image

run only post
image


### AIO deployment present
nothing specified (run only post)
image

run only pre 
image

run both pre and post
image


### weird condition (why would anyone do this)
pre and post both false
image

c-ryan-k:
Some thoughts around pre/post:
- if `pre == false` and `post == false`, maybe we should return an error ("you cannot run checks by disabling all checks, etc")
- if `pre == true`, but AIO is deployed, should we not show post?

My thinking is more along the lines of:
- If AIO is deployed:
  - run post only unless `--post` is false
  - do not run pre unless `--pre` is true
- If AIO is not deployed
  - run pre unless `--pre` is false
  - do not run post unless `--post` is true
  
 Thoughts? I guess I can understand either way, but I figured we could use AIO's status as the default (post || pre) and then let the switches override those.

vilit1:
catering to the popular demand
image

c-ryan-k (approved):

c-ryan-k:
image
just needs one little lint fix

PR #215, opened 2024-05-13
feat: Delete IoT Operations command

digimaun:
* Greatly simplifies re-deployment of AIO. For example, going from `0.4.0` -> `0.5.0` against the same cluster. 
* Currently supports the most straight-forward path - deletion of all related AIO components and resources. No further customization supported.
  * Additional customization will be added in future iterations. 
* More functionality and organization in connected cluster.
* Introduces resource map.
* Initial UX shows a tree / hierarchy of related resources + min dynamic progress bar.
* There are gaps in the delete lifecycle unit tests which will be covered in a future iteration.


azext_edge/edge/_help.py
```diff
@@ -229,6 +230,26 @@ def load_iotops_help():
...
+          Resource Graph being eventually consistent does not guarantee a synchronized state at the time of execution.
+
+        examples:
+        - name: Minimum input for complete deletion.
+          text: >
+            az iot ops delete --cluster mycluster -g myresourcegroup
+        - name: Skip confirmation prompt and continue to deletion process. Useful for CI scenarios.
+          text: >
+            az iot ops delete --cluster mycluster -g myresourcegroup -y
+        - name: Force deletion regardless of warnings. Not recommended.
```
vilit1, line 248: would recommend not having this as a help if we do not recommend it

azext_edge/edge/params.py
```diff
@@ -47,6 +55,24 @@ def load_iotops_arguments(self, _):
             "If not defined, the fallback value `azure-iot-operations` will be used. ",
             validator=validate_namespace,
         )
+        context.argument(
+            "confirm_yes",
+            options_list=["--yes", "-y"],
+            arg_type=get_three_state_flag(),
+            help="Confirm [y]es without a prompt. Useful for CI and automation scenarios.",
```
vilit1, line 62: why not use the azure cli built in confirm? It should be something we can pass through to the delete function being called and add in a help prompt if we want to add flavor

azext_edge/edge/providers/orchestration/work.py
```diff
@@ -574,7 +574,7 @@ def render_display(self, category: Optional[WorkCategoryKey] = None, active_step
     def stop_display(self):
         if self._render_progress and self._live.is_started:
             if self._progress_shown:
-                self._progress_bar.update(self._task_id, description="DONE!")
+                self._progress_bar.update(self._task_id, description="Done.")
```
vilit1, line 577: i like DONE! more

azext_edge/edge/providers/orchestration/connected_cluster.py
```diff
@@ -38,6 +90,11 @@ def resource(self) -> dict:
     def location(self) -> str:
         return self.resource["location"]
 
+    @property
+    def connected(self) -> bool:
+        properties = self.resource.get("properties", {})
+        return "connectivityStatus" in properties and properties["connectivityStatus"].lower() == "connected"
```
vilit1, line 96: properties.get("connectivityStatus", "").lower() == "connected"

shorter and imo not more complicated?

azext_edge/edge/providers/orchestration/connected_cluster.py
```diff
@@ -48,14 +105,38 @@ def extensions(self) -> List[dict]:
...
+        result = self.resource_graph.query_resources(query=query)
+        return self._process_query_result(result)
+
+    def get_resource_sync_rules(self, custom_location_id: str) -> Optional[List[dict]]:
+        query = QUERIES["get_resource_sync_rules"].format(custom_location_id=custom_location_id)
+
+        result = self.resource_graph.query_resources(query=query)
+        return self._process_query_result(result)
+
+    def _process_query_result(self, result: dict, first: bool = False) -> Optional[Union[dict, List[dict]]]:
```
vilit1, line 138: why not just include first in the query?

azext_edge/edge/providers/orchestration/connected_cluster.py
```diff
@@ -48,14 +105,38 @@ def extensions(self) -> List[dict]:
...
-        resources
-        | where type =~ 'microsoft.extendedlocation/customlocations'
-        | where properties.hostResourceId =~ '{self.resource_id}'
-        | where properties.namespace =~ '{namespace}'
-        | project id, name, location, properties, apiVersion
-        """
+        query = QUERIES["get_custom_location_for_namespace"].format(resource_id=self.resource_id, namespace=namespace)
+
+        result = self.resource_graph.query_resources(query=query)
+        return self._process_query_result(result, first=True)
```
vilit1, line 111: can there be multiple custom locations? if yes, do we not care about getting all of them? if not, how do we determine the one to care about (assuming it is not always the first one)?

azext_edge/edge/providers/orchestration/deletion.py
```diff
@@ -0,0 +1,194 @@
...
+        self._progress_shown = False
+
+    def do_work(self, confirm_yes: Optional[bool] = None, force: Optional[bool] = None):
+        self._display_resource_tree()
+
+        should_delete = True
+        if not confirm_yes:
+            should_delete = Confirm.ask("Continue?")
+
+        if not should_delete:
```
vilit1, line 80: if (not confirm_yes) or Confirm.ask("Continue?"):

azext_edge/edge/providers/orchestration/connected_cluster.py
digimaun: Yet slower

azext_edge/edge/providers/orchestration/connected_cluster.py
digimaun: The ARG query? Like limit 1? In this case it would still return an array.

azext_edge/edge/providers/orchestration/connected_cluster.py
digimaun: There can only be 1 custom location associated to a namespace at a time. BTW `get_custom_location_for_namespace` is not used as part of the delete AIO operation, this was existing functionality support an `init`  pre-flight check, just refactoring here to conform with the updated pattern.

azext_edge/edge/providers/orchestration/deletion.py
digimaun: I think it is easier to read as it is now.

azext_edge/edge/providers/orchestration/work.py
digimaun: Currently DONE! is shown even in case of errors, which may convey a happy state even though it's not. We can make a different output for errors, but I don't want to do it in this PR.

azext_edge/edge/_help.py
digimaun: Changed the wording from Not Recommended to May lead to errors.

azext_edge/edge/params.py
```diff
@@ -47,6 +55,24 @@ def load_iotops_arguments(self, _):
...
+            "no_progress",
+            options_list=["--no-progress"],
+            arg_type=get_three_state_flag(),
+            help="Disable visual representation of work.",
+        )
+        context.argument(
+            "force",
+            options_list=["--force"],
+            arg_type=get_three_state_flag(),
+            help="Force the operation to execute.",
```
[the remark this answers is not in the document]
digimaun: Changed existing wording in help from Not Recommended to May lead to errors.

azext_edge/edge/params.py
digimaun: Because we already take a dependency on Rich which supports this functionality. The azure cli built in confirm is part of the command/knack configuration meaning we have less control over it and how we can present in our flows.

If we didn't have a dependency on Rich then the Knack version could be considered.

azext_edge/edge/providers/orchestration/connected_cluster.py
vilit1: what... how?

digimaun: "connectivityStatus in properties" is O(1) + properties["connectivityStatus"] is O(1) .... with properties.get("connectivityStatus", "") first the python runtime must lookup the attribute .get and execute, which will push and pop a frame to the stack, adding more compute expense compared to the former pattern.

azext_edge/edge/providers/orchestration/connected_cluster.py
vilit1: hmm we could just move the get one logic to where it is used (get_custom_location_for_namespace) to minimize params here

azext_edge/edge/providers/orchestration/deletion.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+        if not batched_work:
+            logger.warning("Nothing to delete :)")
+            return
+
+        if not force:
+            if not self.resource_map.connected_cluster.connected:
+                logger.warning(
+                    "Deletion cancelled. The cluster is not connected to Azure. "
+                    "Use --force to continue anyway. Not recommended."
```
vilit1, line 138: do you know what happens if a cluster is not deleted correctly? (aka will people get mad at us)

azext_edge/edge/providers/orchestration/connected_cluster.py
vilit1: also limit 1 would reduce the amount of data we receive (even tho we don't sift through it here)

azext_edge/edge/providers/orchestration/resource_map.py
```diff
@@ -0,0 +1,171 @@
...
+        self.cluster_name = cluster_name
+        self.resource_group_name = resource_group_name
+        self.subscription_id = get_subscription_id(cli_ctx=cmd.cli_ctx)
+        self.connected_cluster = ConnectedCluster(
+            cmd=cmd,
+            subscription_id=self.subscription_id,
+            cluster_name=self.cluster_name,
+            resource_group_name=self.resource_group_name,
+        )
+        self.base_id_prefix = f"/subscriptions/{self.subscription_id}/resourceGroups/{resource_group_name}/providers"
```
vilit1, line 38: not used? unless github is lying to me

azext_edge/edge/providers/orchestration/resource_map.py
```diff
@@ -0,0 +1,171 @@
...
+                    )
+                )
+
+            result = sorted(result, key=lambda r: (r.segments, r.display_name.lower()), reverse=True)
+
+        return result
+
+    def refresh_related_resource_ids(
+        self,
+    ) -> Dict[str, Union[List[Dict[str, str]], Dict[str, Dict[str, Union[List[Dict[str, str]], str]]]]]:
```
vilit1, line 111: nit: this is a very scary typing

azext_edge/edge/providers/orchestration/resource_map.py
```diff
@@ -0,0 +1,171 @@
...
+    def get_resource_sync_rules(self, custom_location_id: str) -> List[IoTOperationsResource]:
+        result = []
+        if (
+            "customLocations" in self._resource_map
+            and self._resource_map["customLocations"]
+            and custom_location_id in self._resource_map["customLocations"]
+            and self._resource_map["customLocations"][custom_location_id]
+            and "resourceSyncRules" in self._resource_map["customLocations"][custom_location_id]
+            and self._resource_map["customLocations"][custom_location_id]["resourceSyncRules"]
+        ):
```
vilit1, line 76: nit: could be a list in all

vilit1: same nit for get_resources if statement

azext_edge/edge/providers/orchestration/resource_map.py
vilit1: would be more helpful to have the typing of each resource (what type is the stored extension, etc)

azext_edge/edge/providers/orchestration/deletion.py
```diff
@@ -0,0 +1,194 @@
...
+        self,
+        resources: Optional[List[IoTOperationsResource]] = None,
+        resource_sync_rules: Optional[List[IoTOperationsResource]] = None,
+        custom_locations: Optional[List[IoTOperationsResource]] = None,
+        extensions: Optional[List[IoTOperationsResource]] = None,
+    ) -> Dict[str, List[List[IoTOperationsResource]]]:
+        batched_work: Dict[str, List[List[IoTOperationsResource]]] = {}
+
+        if resources:
+            resource_batches: List[List[IoTOperationsResource]] = []
```
vilit1, line 161: nit: prob add a note that resources is already sorted here

azext_edge/edge/providers/orchestration/deletion.py
digimaun: You can keep running the delete operation. However if a cluster is not online it will most likely fail.

azext_edge/edge/providers/orchestration/connected_cluster.py
digimaun: > hmm we could just move the get one logic to where it is used (get_custom_location_for_namespace) to minimize params here

The problem with this is having to duplicate logic in `get_custom_location_for_namespace`, to check if the returned value is truthy or not which already happens in `_process_query_result`.

Also when you say minimize parameters, there are only 2 parameters `_process_query_result`.

> also limit 1 would reduce the amount of data we receive (even tho we don't sift through it here)

You asked a related question here, where I am trying to convey there is at most 1 custom location associated to a namespace, so adding limit 1 makes a more verbose query for what value add?

image

azext_edge/edge/providers/orchestration/resource_map.py
digimaun: Not doing in this PR.

azext_edge/edge/providers/orchestration/connected_cluster.py
vilit1: regarding the first one:
yea - I know it is only two params,  wanted to keep it as clean as possible for now. If there will be dup code otherwise, then lets just keep the params

regarding the second one:
lets leave it be then (without the limit 1)

azext_edge/edge/providers/orchestration/resource_map.py
digimaun: I re-wrote this module that includes simplifying this typing.

c-ryan-k (approved):

PR #224, opened 2024-05-20
fix: Validate diagnostics pod is in running phase prior to connection

digimaun:
Identified by @vilit1

### DIFF
--- a/azext_edge/edge/providers/stats.py
+++ b/azext_edge/edge/providers/stats.py
@@ -15,7 +15,7 @@ from azure.cli.core.azclierror import ResourceNotFoundError
 from knack.log import get_logger
 from rich.console import Console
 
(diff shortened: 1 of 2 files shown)


azext_edge/edge/providers/stats.py
```diff
@@ -43,9 +43,14 @@ def _preprocess_stats(
         raise ResourceNotFoundError(
             f"Diagnostics service pod '{diag_service_pod_prefix}' does not exist in namespace '{namespace}'."
         )
-    diagnostic_pod = target_pods[0]
-
-    return namespace, diagnostic_pod
+    for pod in target_pods:
+        if pod.status.phase.lower() == PodState.running.name:
```
vilit1, line 47: can't we just leave it as the phase == PodState.running.value rather than lowercasing both?

or can pod phases be all lowercase?

vilit1 (approved):

azext_edge/edge/providers/stats.py
digimaun: Ok, let's try the other way

PR #251, opened 2024-06-28
feat: Intermediate instance support of show + list. Adds re-usable components. Increment min core CLI version.

digimaun:
* Plumbing and content for iot ops mgmt client.
* Queryable class as a higher-level abstraction for the micro ResourceGraph client.
* Application of new component to Instance show + list.
* Increment min core CLI version to 2.53.0.


.pylintrc
```diff
@@ -14,7 +14,7 @@ ignore=CVS
 ignore-patterns=
 
 # Match against paths supporting python regex.
-ignore-paths=^azext_edge/sdk/.*$
```
vilit1, line 17: are we going back to the sdk generation path? If so, then please document the versions (node, autorest, autorest-python) and commands used so we don't have to deal with guessing/having new sdks be wildly different.

azext_edge/edge/providers/orchestration/instances.py
```diff
@@ -0,0 +1,75 @@
...
+
+        return result
+
+    def list(self, resource_group_name: Optional[str] = None) -> List[dict]:
+        if resource_group_name:
+            return list(
+                self.iotops_mgmt_client.instance.list_by_resource_group(resource_group_name=resource_group_name)
+            )
+
+        return list(self.iotops_mgmt_client.instance.list_by_subscription())
```
vilit1, line 55: does the service not return a workable itterrable already?

azext_edge/edge/providers/orchestration/instances.py
```diff
@@ -0,0 +1,75 @@
...
+        | where type =~ 'microsoft.extendedlocation/customlocations'
+        | where id =~ '{resource_id}'
+        | project id, name, properties
+        """
+}
+
+INSTANCES_API_VERSION = "2021-10-01-privatepreview"
+# TODO temporary
+BASE_URL = "https://eastus2euap.management.azure.com"
+QUALIFIED_RESOURCE_TYPE = "Private.IoTOperations/instances"
```
vilit1, line 30: This can be baked into the ResourceTypeMapping/ResourceProviderMapping enums

.pylintrc
digimaun: autorest configuration is documented here https://github.com/Azure/azure-iot-ops-cli-extension/wiki/Autorest-config

azext_edge/edge/providers/orchestration/instances.py
digimaun: This is a temporary mapping for this intermediate PR to work with private rp. In the future will evaluate if an enum makes sense, which a datapoint will be which other components other than the mgmt class needs to know about a typing. If other operations don't then there isn't a reason to have an enum.

azext_edge/edge/providers/orchestration/instances.py
digimaun: You mean the client? It is an iterable, but we don't currently need to process them, therefore the whole iterable chain is collected into a list and returned.

vilit1: yes - it is rare to see the iterable be converted to a list before it is returned, especially since there is no transformations/extra code modifying it

azext_edge/edge/providers/orchestration/instances.py
```diff
@@ -0,0 +1,75 @@
...
+from knack.log import get_logger
+from rich import print
+
+from ...util.az_client import get_iotops_mgmt_client, parse_resource_id
+from ...util.queryable import Queryable
+
+logger = get_logger(__name__)
+
+
+QUERIES = {
```
Elsie4ever, line 18: are we making this a dict with one entry?

digimaun: It's a dict if more entries will be added later, similar to pattern in connected_cluster

azext_edge/edge/providers/orchestration/instances.py
Elsie4ever: looks like the function return type is Iterable[JSON]?

azext_edge/edge/providers/orchestration/instances.py
Elsie4ever: ok got it, thanks!

Elsie4ever (approved):

azext_edge/edge/providers/orchestration/instances.py
digimaun: We can refactor later if no new queries need to be made on behalf of instances.

azext_edge/edge/providers/orchestration/instances.py
digimaun: Changed the method to return an iterable directly rather than convert to list - now we rely on cli core to unpack. The instance list test changed to unpack the iterable.

vilit1 (approved):

PR #254, opened 2024-07-01
test: add integration testing for az iot ops init

vilit1:
--simulate-plc
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/9864712110
image
image

nothing
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/9864710383
image)
eyes

Note that failing tests are due to deployment/smoke tests/normal integration tests failing. Will update those in a future pr. maybe.
image

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/tests/edge/init/int/opcua_helper.py
```diff
@@ -0,0 +1,22 @@
...
+def assert_simulate_plc_args(
+    custom_location: str,
+    simulate_plc: Optional[bool] = None,
+    **_
+):
+    if not simulate_plc:
+        simulate_plc = False
+    # note that the simulator may take a bit
+    query_result = run(f"az iot ops asset query --cl {custom_location}")
+    assert bool(query_result) is simulate_plc
```
digimaun, line 20: This will depend on resources to be synced from edge to cloud. How is timing being handled? This does not look like a consistent assertion. Also the testing matrix does not currently include --simulate-plc.

azext_edge/tests/edge/init/int/test_init_int.py
```diff
@@ -0,0 +1,117 @@
...
+    settings.add_to_config(EnvironmentVariables.sp_app_id.value)
+    settings.add_to_config(EnvironmentVariables.sp_object_id.value)
+    settings.add_to_config(EnvironmentVariables.sp_secret.value)
+    settings.add_to_config(EnvironmentVariables.init_args.value)
+    settings.add_to_config(EnvironmentVariables.aio_cleanup.value)
+
+    if not all([settings.env.azext_edge_cluster, settings.env.azext_edge_rg, settings.env.azext_edge_kv]):
+        raise AssertionError(
+            "Cannot run init tests without a connected cluster, resource group, and precreated keyvault. "
+            f"Current settings: {settings}"
```
digimaun, line 33: See comment on why this can be dangerous.

.github/workflows/int_test.yml
```diff
@@ -7,8 +7,8 @@ on:
         type: string
         required: true
         default: ops-cli-int-test-rg
-      template-content:
-        description: Custom AIO deployment template file contents, used for all deployments in this workflow.
+      runtime-init-args:
+        description: Additional init arguments (beyond cluster name, resource group, key vault, and service principal arguments).
```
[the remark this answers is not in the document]
c-ryan-k: Something else we could do is add a conditional matrix step that `only` runs the `runtime-init-args`, and only runs if this value is provided - that way the existing scenarios still execute normally but we still have a way to do a custom init at runtime?

azext_edge/tests/edge/init/int/test_init_int.py
```diff
@@ -0,0 +1,117 @@
...
+        "additionalArgs": settings.env.azext_edge_init_args.strip('"')
+    }
+    if settings.env.azext_edge_aio_cleanup:
+        run(
+            f"az iot ops delete --cluster {settings.env.azext_edge_cluster} -g {settings.env.azext_edge_rg} "
+            "-y --no-progress --force"
+        )
+
+
+@pytest.mark.init_scenario_test
```
digimaun, line 51: This warning can be eliminated.

image

azext_edge/tests/settings.py
```diff
@@ -60,3 +64,6 @@ def _build_config(self, env_set: List[str], optional: Optional[bool] = False):
                     raise RuntimeError(
                         "{} environment variables required.".format(",".join(env_set))
                     )
+
+    def __repr__(self):
+        return " \n".join([f"{key}: {value}" for key, value in vars(self.env).items()])
```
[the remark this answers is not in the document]
vilit1: secret values are already masked by github. I will add in an option in this class to mask it at repr execution

.github/workflows/int_test.yml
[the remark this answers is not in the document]
vilit1: I will try ryan's suggestion. May take a bit of finageling

.github/workflows/int_test.yml
```diff
@@ -164,13 +168,16 @@ jobs:
...
           custom-locations-oid: ${{ env.CUSTOM_LOCATIONS_OID }}
-      - name: "Get Keyvault ID"
+      - name: "Tox test environment setup"
         run: |
-          KV_ID=$(az keyvault show -n ${{env.KV_NAME}} -g ${{ env.RESOURCE_GROUP }} -o tsv --query id)
-          echo "KV_ID=$KV_ID" >> $GITHUB_ENV
+          cd ${{ env.EXTENSION_SOURCE_DIRECTORY }}
+          python -m pip install tox
+          tox r -vv -e python-int --notest
+          tox r -vv -e python-init-int --notest
```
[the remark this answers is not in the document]
vilit1: will try without - I wasn't sure if each tox option will try to create multiple tox folders...

azext_edge/tests/settings.py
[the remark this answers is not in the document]
digimaun: To only rely on github runtime, and that is if you correctly configure secrets, is one slice of protection. We can write more secure code that is independent of action runners. You can find examples in Azure SDK where sensitive headers are masked in code.

.github/workflows/int_test.yml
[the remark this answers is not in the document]
c-ryan-k: A few things here - the `--notest` sets up the tox env by pre-installing dependencies and preparing the test env for a quicker run, and also helps break out test setup failures from actual test run failures.

The `python-int` setup only needs to run if we're not containerizing the tests (because tox will do its own setup inside the container), otherwise it's unused. That said, it doesn't technically hurt to do it (other than wasting time) - which we can kind of mitigate by utilizing the `-p` switch to run tox steps in parallel:

`tox r -vv -e "python-int, python-init-int" --notest`


That said, it's probably still better to break up the tox setup into:

- Tox setup for init (happens always in all scenarios we're running using the tests to init)
- Tox setup for full int tests (`if: ${{matrix.feature == 'default' && !inputs.use-container}}`)

.github/workflows/int_test.yml
```diff
@@ -189,42 +196,23 @@ jobs:
...
-          mq-insecure: ${{ matrix.mq-insecure }}
-          disable-rsync-rules: ${{ matrix.disable-rsync-rules }}
-          ca-valid-days: ${{ matrix.ca-valid-days || '' }}
-          ca-file: ${{ matrix.ca-file || '' }}
-          ca-key-file: ${{ matrix.ca-key-file || '' }}
-          kv-spc-secret-name: ${{ matrix.kv-spc-secret-name || '' }}
-          template-file: ${{ inputs.template-content && 'custom-template.json' || '' }}
-          csi-config: ${{ matrix.csi-config || ''}}
-          include-dp: ${{ matrix.include-dp }}
-      - name: "Allow cluster to finish provisioning"
```
c-ryan-k, line 219: Curious about the test success rate when this sleep is removed. I know there were some int and smoke tests (stats as an example) that required a little bit of time for the cluster to finish setting up after deployment, but that may no longer be an issue.

.github/workflows/int_test.yml
```diff
@@ -71,8 +76,10 @@ jobs:
         tenant-id: ${{ secrets.AZURE_TENANT_ID }}
         subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
     - name: "Create Key Vault for clusters"
+      id: create-keyvault
```
c-ryan-k, line 79: nit: probably don't need an id on this step if we're not referencing it anywhere later

.github/workflows/int_test.yml
```diff
@@ -275,7 +269,7 @@ jobs:
           set +e
           docker run \
             --rm \
-            "${envVars[@]}" \
+            "${ envVars[@]}" \
```
c-ryan-k, line 273: nit: spacing - the line above has a space before `\` the break character

azext_edge/tests/edge/init/int/test_init_int.py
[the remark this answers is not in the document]
vilit1: updated:
image

azext_edge/tests/edge/init/int/helper.py
```diff
@@ -0,0 +1,132 @@
...
+    assert result["csiDriver"]["keyVaultId"] == key_vault
+    assert result["csiDriver"]["kvSpcSecretName"] == arg_dict.get("kv_spc_secret_name", "azure-iot-operations")
+    assert result["csiDriver"]["version"] == arg_dict.get("csi_ver", KEYVAULT_ARC_EXTENSION_VERSION)
+
+    if sp_app_id:
+        assert result["csiDriver"]["spAppId"] == sp_app_id
+    if sp_object_id:
+        assert result["csiDriver"]["spObjectId"] == sp_object_id
+    csi_config = result["csiDriver"]["configurationSettings"]
+    enabled_rotation = 'false' if arg_dict.get("disable_rotation") else 'true'
```
c-ryan-k, line 49: What about `--disable-rotation false`?

vilit1: was going to leave that part alone (since this is a double negative and it is the same as just not having the arg)
but I guess it would help to have it

azext_edge/tests/edge/init/int/helper.py
```diff
@@ -0,0 +1,132 @@
...
+    assert csi_config["secrets-store-csi-driver.enableSecretRotation"] == enabled_rotation
+    assert csi_config["secrets-store-csi-driver.rotationPollInterval"] == arg_dict.get("rotation_int", "1h")
+    assert csi_config["secrets-store-csi-driver.syncSecret.enabled"]
+
+    # deployment state
+    assert result["deploymentState"]["correlationId"]
+    assert result["deploymentState"]["timestampUtc"]
+    _assert_aio_versions(result["deploymentState"]["opsVersion"], arg_dict.get("include_dp", False))
```
c-ryan-k, line 57: do we still need a test for `--include-dp`?

vilit1: the dp resources are still generated (since we have the include-dp option). If we remove that option then I will remove this part

azext_edge/tests/edge/init/int/opcua_helper.py
```diff
@@ -0,0 +1,33 @@
...
+    custom_location: str,
+    simulate_plc: Optional[bool] = None,
+    **_
+):
+    if not simulate_plc:
+        simulate_plc = False
+    # note that the simulator may take a bit
+    query_result = run(f"az iot ops asset query --cl {custom_location}")
+    tries = 0
+    while simulate_plc and not query_result and tries < MAX_TRIES:
```
digimaun, line 26: To test --simulate-plc, we can do cluster side check the opcplc-00000 pod was deployed and related asset & aep CRs are created. I don't see why we need to test resource hydration via rsync rules.

vilit1: i thought we wanted to avoid cluster side if there is a cloud side solution...

azext_edge/tests/edge/init/int/helper.py
vilit1: image

azext_edge/tests/edge/init/int/opcua_helper.py
digimaun: The goal is to assert cloud side when resources are created cloud first, then assert cluster side when resources are created edge first. Both the simulator and its asset & aep are edge side creation and rely on RSRs to hydrate cloud.

.github/workflows/int_test.yml
vilit1: aside from the simulate plc taking time, it seems like everything else works. A bit hard to check smoke tests with the current state of the cluster deployments

azext_edge/tests/edge/init/int/opcua_helper.py
vilit1: changed to just asset the opcplc pod (rather than wait for the simulator to be done and have the assets/etc be populated).

digimaun: Sounds good

.github/workflows/int_test.yml
```diff
@@ -7,8 +7,8 @@ on:
         type: string
         required: true
         default: ops-cli-int-test-rg
-      template-content:
-        description: Custom AIO deployment template file contents, used for all deployments in this workflow.
+      runtime-init-args:
+        description: Additional init arguments (beyond cluster name, resource group, key vault, and service principal arguments).
         type: string
         required: false
```
c-ryan-k, line 13: do we also need `default: ''` here? In case they don't provide it in `workflow_call`, will it get resolved as null or `''`?

vilit1: I don't want to take any chances and just say '' since null and '' are treated differently

.github/workflows/int_test.yml
```diff
@@ -57,7 +62,7 @@ permissions:
   id-token: 'write'
 
 env:
-  KV_NAME: "opskv${{ github.run_number }}"
+  KV_NAME: "opskv${{ github.run_number }}int"
```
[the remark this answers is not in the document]
vilit1: ok - will look into other identifiers too

.github/workflows/int_test.yml
c-ryan-k: Have you tested `workflow_call`? The default is set to `''` [here](https://github.com/Azure/azure-iot-ops-cli-extension/pull/254/files#diff-3034c5f172c4708ce5a8534f97d8430c2052be88efba38cb968b470b73a7349dR52) in `workflow_dispatch`, I figured this way it would run the same regardless of how the test was triggered.

.github/workflows/int_test.yml
```diff
@@ -84,36 +89,50 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        feature: [default, mq-insecure, no-syncrules, ca-certs]
-        include:
-          # default / limited options
+        feature: [custom-input, default, mq-insecure, no-syncrules, ca-certs]
+        runtime-args:
+          - ${{ inputs.runtime-init-args != '' }}
+        exclude:
```
c-ryan-k, line 100: Talked offline about how this works, but perhaps we should document this in the integration test readme somewhere since it's a bit more convoluted

.github/workflows/int_test.yml
vilit1: oh whoops - will update the workflow call - I do not want to take chances

.github/workflows/int_test.yml
```diff
@@ -290,44 +296,53 @@ jobs:
...
-          az iot ops mq get-password-hash -p test
+          # az iot ops support create-bundle --svc broker --broker-traces
+          # az iot ops check
+          # az iot ops check --pre
+          # az iot ops check --post
+          # az iot ops check --as-object
+          # az iot ops check --svc broker --resources broker brokerlistener diagnosticservice
+          # az iot ops mq stats
+          # az iot ops mq stats --raw
+          # az iot ops mq get-password-hash -p test
```
c-ryan-k, line 307: nit/note for future, this command is also removed (and `ops mq stats` will also be renamed under broker)

.github/workflows/int_test.yml
```diff
@@ -290,44 +296,53 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+    - name: "Az CLI login"
+      uses: azure/login@v2
+      with:
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+    - name: "Delete Key Vault for clusters"
+      run: az keyvault delete -n ${{ env.KV_NAME }} -g ${{ env.RESOURCE_GROUP }} --no-wait
```
c-ryan-k, line 348: Just a thought, is there a way to delete and purge at the same time? Deleting will remove the KV resource but we'll still end up with name conflicts if we don't also purge the KV.

It probably won't work with `--no-wait` though

tox.ini
```diff
@@ -65,6 +65,33 @@ commands =
     pytest -k _unit ./azext_edge/tests --cov=azext_edge/edge --cov-config .coveragerc --junitxml=junit/test-aziotops-ext-unit.xml {posargs}
     # You can pass additional positional args to pytest using `tox -e [env] -- -s -vv`
 
+# init integration tests
+[testenv:py{thon,38,39,310,311}-init-int]
+skip_install = True
+description =
+    {[base]description}
+    int: integration tests
```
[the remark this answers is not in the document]
c-ryan-k: or `init-only` etc, just so we know that these int tests are ONLY running `init` checks

.github/workflows/int_test.yml
vilit1: we will have to change it to not have the no-wait then... 

Let's just leave it as is for now (since it should get purged at some point either way)

.github/workflows/int_test.yml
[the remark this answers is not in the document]
vilit1: We decided to:
1. use `opsint${{ github.run_id }}` so we get out of clashing (and hopefully avoid clashing later since run_id should be unique per repo.
2. change the clean-up pipeline to purge all keyvaults + remove anything `az iot ops delete` did not take care of.

vilit1:
gonna close this since we wont stare at this branch.
next pr comming soon (TM)

PR #262, opened 2024-07-03
refactor!: Wave of mq -> broker renames.

digimaun:
* Formalizes broker command group.
* Also removes the get-password-hash command.


azext_edge/tests/edge/checks/int/test_pre_post_int.py
```diff
@@ -31,7 +31,7 @@ def test_check_pre_post(init_setup, post, pre):
     result = run(command)
 
     # default service title
-    expected_title = "Evaluation for {mq} service deployment"
```
Elsie4ever, line 34: I thought for display we would use mqttbroker?🤔

Elsie4ever (approved):

PR #294, opened 2024-08-02
feat: Init includes default dataflow profile

digimaun:


digimaun:
IT run: https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/10220455435

azext_edge/edge/commands_edge.py
```diff
@@ -224,6 +225,7 @@ def init(
         mq_broker_name=str(mq_broker_name),
         mq_authn_name=str(mq_authn_name),
         mq_insecure=mq_insecure,
+        dataflow_profile_instances=int(dataflow_profile_instances),
```
Elsie4ever, line 228: why do we need to cast dataflow_profile_instances into int if the typing is already int?

vilit1: if you somehow get a string, it means you need to fix your param (arg_type=int or soemthing)

azext_edge/edge/params.py
```diff
@@ -619,6 +619,13 @@ def load_iotops_arguments(self, _):
             help="The path to a custom IoT Operations deployment template. Intended for advanced use cases.",
             deprecate_info=context.deprecate(hide=True),
         )
+        context.argument(
+            "dataflow_profile_instances",
```
vilit1, line 623: the dataflow profile should have a name right (what is the default then)? a bit weird that only a number is exposed...

digimaun: The name is 'profile' and should not be changed as DOE depends on a static name.

vilit1: oh then I guess applying the dataflow crds isn't valid lol

azext_edge/edge/commands_edge.py
```diff
@@ -124,6 +124,7 @@ def init(
     mq_authn_name: str = "authn",
     mq_broker_config_file: Optional[str] = None,
     mq_insecure: Optional[bool] = None,
+    dataflow_profile_instances: int = 1,
```
vilit1, line 127: can we not get away with no dataflow profile instances?

digimaun: What is the concern with exposing this ?

vilit1: oh I meant as in make an aio deployment without the dataflow instance - sorry, I took a motion sickness pill and am pretty sleepy now

azext_edge/edge/commands_edge.py
digimaun: The reason for this is when default values are assigned to the command handler, it isn't the native Python type but a `knack` equivalent. This can cause issues down the line in processing, if say you depend on isinstance() and you assume the native type. The code is continuously being refactored, but because of AIO's reliance on `init` at this point in time, I kept the same patterns as before (with other default args). 

If after reading, and it is still a concern, happy to discuss further.

> if you somehow get a string, it means you need to fix your param (arg_type=int or soemthing)

^-- the new parameter introduced in this pr has arg_type=int.

azext_edge/edge/commands_edge.py
digimaun: You can deploy AIO without a dataflow profile, but every dataflow requires a profile. And dataflows are common scenarios in AIO. Also, DOE needs the default profile available.

azext_edge/edge/params.py
digimaun: One can create other profiles if they want, we are just ensuring the default one is deployed with init.

vilit1 (approved):

Elsie4ever (approved):

PR #301, opened 2024-08-08
refactor: remove cl graph lookup + add Instances get_resource_map()

digimaun:
* Eliminate usage of resource graph for fetching custom location associated with instance.
* Provide useful public method on `Instances`, `get_resource_map`.

    Example:
    
    ```python
    instances = Instances(cmd)
    myinstance = instances.show(name=instance_name, resource_group_name=resource_group_name)
    resource_map = instances.get_resource_map(myinstance)

    # Check if connected cluster is online
   assert resource_map.connected_cluster.connected
    
    ```


c-ryan-k (approved):

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -50,28 +48,25 @@ def list(self, resource_group_name: Optional[str] = None) -> Iterable[dict]:
...
-                cluster_name=resource_id_container.resource_name,
-                resource_group_name=resource_id_container.resource_group_name,
-            )
+            resource_map.refresh_resource_state()
         print(resource_map.build_tree(category_color="cyan"))
 
-    def get_associated_cl(self, instance: dict) -> dict:
-        return self.query(
-            QUERIES["get_cl_from_instance"].format(resource_id=instance["extendedLocation"]["name"]), first=True
+    def _get_associated_cl(self, instance: dict) -> dict:
```
vilit1, line 56: not sure if this should be abstracted since the assets also need the custom location...

vilit1 (approved):

azext_edge/edge/providers/orchestration/resources/instances.py
vilit1: unless you expect us to create an instance of this class for assets/other rpsaas services

digimaun: What do you mean by this ? The instance has reference to the custom location Id. What do you need from the custom location to create an asset ?

vilit1: I need the custom location id (for the extended location) + all the cluster + custom location checks for asset/endpoint creation...

PR #303, opened 2024-08-12
feat: use table view in pod check

Elsie4ever:
update pod check display to use table view in detail level 1 and 2
0 will keep the same as before to summarize pod check in one line

level 0:
image

level 1:
image
image

level 2:
image
image

for pods does not have conditions property, it will look like:
image

---

### General Guidelines

Intent for Production

Basic expectations


Elsie4ever:
IT runs
image

azext_edge/edge/providers/check/cloud_connectors.py
```diff
@@ -246,18 +245,18 @@ def _display_connector_runtime_health(
...
-        for pod in pod_name_prefixes:
-            evaluate_pod_health(
-                check_manager=check_manager,
-                target=target,
-                namespace=namespace,
-                pod=pod,
-                display_padding=padding,
-                detail_level=detail_level,
-            )
+        # pod_name_prefixes = [
```
vilit1, line 248: commented out but not deleted?

Elsie4ever: for this file Ryan mentioned the logic might be reusable for dataflow logic so he would want to keep this for longer, therefore I thought comment it out so he would still see the structure

c-ryan-k: I think we can probably safely remove the cloud_connector code. Dataflows are different enough that I don't think we'll get much use out of this.

Elsie4ever: got it, will update!

Elsie4ever:
Talked with @c-ryan-k for some refactoring, will re open the pr once refactor done

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -28,228 +28,186 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
+    pod_conditions: list = pod_dict.get("status", {}).get("conditions", [])
+    pod_phase_deco, status = decorate_pod_phase(pod_phase)
+
+    pod_eval_value = {}
+    pod_eval_status = status
+    pod_eval_value["status.phase"] = pod_phase
+
+    conditions_readiness = True
+    conditions_display_list: List[Tuple[str, str]] = []
+    unknown_conditions_display_list: List[Tuple[str, str]] = []
```
c-ryan-k, line 129: It looks like these are lists of tuples, like `("key: value", formatted)`, but then later on they are joined together into one string.

Could we just create them as one string when they're added to this list?

Something like:
```py
lines = [
    f"{condition_type}: {pod_condition_deco}",
    formatted_reason
]
conditions_display_list.append('/n'.join(lines))
```

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -28,228 +28,186 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
+                condition_type = type
+                condition_status = condition.get("status")
+
+            formatted_reason = ""
+            condition_reason = condition.get("reason", "")
+
+            if condition_reason:
+                formatted_reason = f"[red]Reason: {condition_reason}[/red]"
+
+            known_condition_values = [value.replace(" ", "").lower() for value in POD_CONDITION_TEXT_MAP.values()]
```
c-ryan-k, line 152: This should probably be stored outside of the loop, no need to regenerate this list every time

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -28,228 +28,186 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
+
+    pod_conditions = [
+        f"{target_service_pod}.status.phase",
+        f"{target_service_pod}.status.conditions",
+    ]
+
+    if check_manager.targets.get(target, {}).get(namespace, {}).get("conditions", None):
+        check_manager.add_target_conditions(target_name=target, namespace=namespace, conditions=pod_conditions)
+    else:
+        check_manager.set_target_conditions(target_name=target, namespace=namespace, conditions=pod_conditions)
```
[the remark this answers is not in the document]
Elsie4ever: it's because of the latter, sometimes a condition already exists before we evaluate the pod check, calling set_target_conditions will overwrite the existing conditions. meanwhile I encountered time when target/namespace not existed, and that will cause an exception

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -28,228 +28,186 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
+
+    pod_dict = pod.to_dict()
+    pod_name = pod_dict["metadata"]["name"]
+    pod_phase = pod_dict.get("status", {}).get("phase")
+    pod_conditions: list = pod_dict.get("status", {}).get("conditions", [])
```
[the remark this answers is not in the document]
Elsie4ever: omg these are 2 different meaning of "condition", thanks for catching it! line 107 means the check conditions and this one means the actual status.conditions value

azext_edge/edge/providers/check/base/pod.py
Elsie4ever: I separated them because we are not showing the formatted_reason in all scenarios, for detail level we will only show reason when one of the condition failed, but in verbose we will always show when there is one. Will think if there is a better way to do this

Elsie4ever: for this one, I also made a namedTuple called PodStatusConditionResult, hopefully that can be more cleared!

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -27,229 +33,197 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
-            display_padding=display_padding,
             detail_level=detail_level,
         )
+        pod_statuses.append(pod_status_result.eval_status)
+        table.add_row(*pod_status_result.display_strings)
 
-        check_manager.add_target_eval(
+    add_footer = not all([status == CheckTaskStatus.success.value for status in pod_statuses])
+
+    if pods:
```
c-ryan-k, line 76: this conditional likely isn't needed due to lines 44-45

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -27,229 +33,197 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
-            eval_status=CheckTaskStatus.warning.value,
-            eval_value=None,
-            resource_name=target_service_pod,
-            namespace=namespace,
-            padding=(0, 0, 0, display_padding),
-        )
-
-    for pod in pods:
-        target_service_pod = f"pod/{pod.metadata.name}"
+        return
```
c-ryan-k, line 45: I can't remember if this was always the case with this PR - but did this method display a "no pods" message? Or is that up to the individual CR check now that it's removed?

c-ryan-k (commented):
Overall looks good, just a final few questions / nits to make sure we've got everything covered :)

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -27,229 +33,197 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
+
+    conditions_readiness = True
+    conditions_display_list: List[PodStatusConditionResult] = []
+    unknown_conditions_display_list: List[PodStatusConditionResult] = []
+
+    # When pod in obnormal state, sometimes the conditions are not available
+    if pod_conditions:
+        known_condition_values = [value.replace(" ", "").lower() for value in POD_CONDITION_TEXT_MAP.values()]
+        for condition in pod_conditions:
+            type = condition["type"]
```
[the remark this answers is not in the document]
Elsie4ever: yeah whenever we have condition property, the type is a required field

azext_edge/edge/providers/check/base/pod.py
Elsie4ever: yeah now the no pod message will be under individual CR check, since we need the label to output the error message, meanwhile we don't want to pass the label down to evaluate_pod_health

azext_edge/edge/providers/check/mq.py
```diff
@@ -495,26 +497,53 @@ def evaluate_brokers(
...
+
+            for prefix, label in [
+                (AIO_BROKER_DIAGNOSTICS_PROBE_PREFIX, MQ_NAME_LABEL),
+                (AIO_BROKER_FRONTEND_PREFIX, MQ_NAME_LABEL),
+                (AIO_BROKER_BACKEND_PREFIX, MQ_NAME_LABEL),
+                (AIO_BROKER_AUTH_PREFIX, MQ_NAME_LABEL),
+                (AIO_BROKER_HEALTH_MANAGER, MQ_NAME_LABEL),
+                (AIO_BROKER_DIAGNOSTICS_SERVICE, MQ_NAME_LABEL),
+                (AIO_BROKER_OPERATOR, MQ_NAME_LABEL),
+                (AIO_BROKER_FLUENT_BIT, MQ_NAME_LABEL),
```
c-ryan-k, line 510: Just a note - and this might change in the future, but I think for now we can probably remove the label from these tuples since they're all using the same one

azext_edge/edge/providers/check/base/pod.py
```diff
@@ -27,229 +33,191 @@ def decorate_pod_phase(phase: str) -> Tuple[str, str]:
...
-            conditions_readiness=conditions_readiness,
-            conditions_display_list=conditions_display_list,
-            unknown_conditions_display_list=unknown_conditions_display_list,
-            display_padding=display_padding,
             detail_level=detail_level,
         )
+        pod_statuses.append(pod_status_result.eval_status)
+        table.add_row(*pod_status_result.display_strings)
+
+    add_footer = not all([status == CheckTaskStatus.success.value for status in pod_statuses])
```
c-ryan-k, line 74: This should fix the linter issue (please double-check) - it's asking us to use a generator instead of creating a new comprehension object to pass to `all`
```suggestion
    add_footer = not all(status == CheckTaskStatus.success.value for status in pod_statuses)
```

Elsie4ever: oh I see, thanks a lot!

azext_edge/edge/providers/check/mq.py
Elsie4ever: oh got catch, will update!

c-ryan-k (approved):

PR #359, opened 2024-09-16
feat: secret sync enablement and additional identity mgmt refactors and improvementss

digimaun:
- Includes `az iot ops dataflow identity *` -> `az iot ops identity *`.


azext_edge/edge/_help.py
```diff
@@ -1215,7 +1268,7 @@ def load_iotops_help():
         "iot ops schema"
     ] = """
         type: group
-        short-summary: Schema management.
+        short-summary: Schema and registry management.
```
Elsie4ever, line 1271: why is this "Schema and registry" instead of "Schema registry"?

PR #382, opened 2024-09-24
feat: add authentication and authorization to broker service check

Elsie4ever:
1. add authentication and authorization reference check in broker listener
2. add broker reference check for all resources type excluding broker itself, now it moved to ownerReference
3. Add authentication and authorization CR check(including the secret and configmap checks)

level 0
image

level 1
image

level 2
image
image

All green version
image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/base.py
```diff
@@ -366,6 +366,19 @@ def delete_namespaced_custom_object(
             raise RuntimeError(error_msg)
 
 
+def get_namespaced_configmap(namespace: str, cm_name: str) -> dict:
```
digimaun, line 369: Why not use https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/k8s/config_map.py#L16 ?

Also if you are going to add new k8s resource fetch functions, please put them in the k8s subpackage like above.

Elsie4ever:
IT runs
image

azext_edge/tests/edge/checks/conftest.py
```diff
@@ -241,3 +243,25 @@ def _handle_list_deployments(*args, **kwargs):
...
+def mocked_validate_ref(mocker):
+    def _handle_validate_ref(*args, **kwargs):
+        ref_text = ""
+        if kwargs["ref_type"] == ValidationResourceType.secret:
+            ref_text = (
+                f"[green]Valid[/green] {ValidationResourceType.secret.value} reference {{[green]mock_secret[/green]}}."
+            )
+        elif kwargs["ref_type"] == ValidationResourceType.configmap:
+            ref_text = f"[green]Valid[/green] {ValidationResourceType.configmap.value}\
+                reference {{[green]mock_configmap[/green]}}."
```
vilit1, line 258: end the string before you split it, otherwise the indentation is grabbed as extra spaces:
`[green]Valid[/green] configmap                reference {[green]mock_configmap[/green]}`

azext_edge/edge/providers/check/mq.py
```diff
@@ -121,13 +127,17 @@ def evaluate_broker_listeners(
...
         listeners = list(listeners)
         listeners_count = len(listeners)
-        listener_count_desc = "- Expecting [bright_blue]>=1[/bright_blue] broker listeners per namespace. {}"
+        listener_count_desc = f"- Expecting {colorize_string('>=1')} broker listeners per namespace. "
         listeners_eval_status = CheckTaskStatus.success.value
 
         if listeners_count >= 1:
-            listener_count_desc = listener_count_desc.format(f"[green]Detected {listeners_count}[/green].")
+            listener_count_desc = (
+                listener_count_desc + f"{colorize_string(color='green', value=f'Detected {listeners_count}')}."
```
vilit1, line 135: nit: could do +=

azext_edge/edge/providers/check/base/resource.py
```diff
@@ -468,3 +470,19 @@ def process_custom_resource_status(
                     namespace=namespace,
                     display=Padding(status_text, (0, 0, 0, padding + 4)),
                 )
+
+
+def validate_ref(name: str, namespace: str, ref_type: ValidationResourceType) -> Tuple[str, bool]:
```
vilit1, line 475: is this helper going to be specific to mq only?

azext_edge/edge/providers/check/mq.py
```diff
@@ -547,6 +674,328 @@ def evaluate_brokers(
...
+    ref_eval_status = CheckTaskStatus.success.value
+    ref_eval_value = {}
+
+    if broker_reference_name not in valid_broker_refs:
+        ref_display = f"{colorize_string(color='red', value='Invalid')} broker reference {{{colorize_string(color='red', value=broker_reference_name)}}}."
+        ref_eval_status = CheckTaskStatus.error.value
+        ref_eval_value["valid(spec.brokerRef)"] = False
+    else:
+        ref_display = f"{colorize_string(color='green', value='Valid')} broker reference {{{colorize_string(color='green', value=broker_reference_name)}}}."
+        ref_eval_value["valid(spec.brokerRef)"] = True
```
c-ryan-k, line 986: Similar nits as mentioned above re: breaking out the strings and values, but another consideration is there are some places we use this format:

`[resource_kind] reference {resource_name} is [valid/invalid]` - with the resource name standard cyan and only the valid/invalid colored as success / error.

And it looks like here it's:

`[Valid/Invalid] [resource_kind] reference {resource_name}` with both the valid/invalid and the resource_name colorized as green/red.

I'm not sure how many of these we have across, say, brokers and dataflows, but we should align them to be consistent

azext_edge/edge/providers/check/mq.py
```diff
@@ -746,10 +1199,354 @@ def _evaluate_broker_diagnostics_service(
...
+    if custom_objects:
+        objects: List[dict] = custom_objects.get("items", [])
+        for object in objects:
+            o: dict = object
+            metadata: dict = o.get("metadata", {})
+            name = metadata.get("name")
+            if name:
+                result[name] = True
+
+    return result
```
c-ryan-k, line 1223: This might be helpful if pulled into a base resource function that takes an API object, what do you think?

azext_edge/edge/providers/check/mq.py
```diff
@@ -746,10 +1199,354 @@ def _evaluate_broker_diagnostics_service(
...
+    errors_displays = []
+    for result in sub_check_results:
+        # summary level will only show header and non-success results
+        # detail level will show all results with evaluation status
+        # verbose level will show all results
+        if (
+            detail_level == ResourceOutputDetailLevel.summary.value
+            and result.eval_status == CheckTaskStatus.error.value
+        ):
+            text = result.display.renderable.replace("- ", "")
```
c-ryan-k, line 1243: This might remove hyphens from other parts of displays - if we only care about hyphens on the outside of strings, we could also use `.strip("- ")`

azext_edge/edge/providers/check/mq.py
```diff
@@ -746,10 +1199,354 @@ def _evaluate_broker_diagnostics_service(
...
+                namespace=namespace,
+                display=result.display,
+            )
+
+    # for errors displayed in summary level, align padding with parent
+    for error_display in errors_displays:
+        check_manager.add_display(
+            target_name=target_name,
+            namespace=namespace,
+            display=Padding(error_display, (0, 0, 0, parent_padding + 4)),
```
c-ryan-k, line 1259: just a nit - I think we have a constant for default padding size somewhere, not sure if this needs to be custom but just in case that value ever changes this may break some alignment.

For what it's worth we also have this problem in several other places where we hardcode `8` and `12`

c-ryan-k (commented):
Chunk of review - I left some comments that are nits that can just be resolved / addressed in a future PR regarding places where we create essentially the same string twice just with a different color / status / value property

azext_edge/edge/providers/check/mq.py
```diff
@@ -121,13 +127,17 @@ def evaluate_broker_listeners(
...
         if listeners_count >= 1:
-            listener_count_desc = listener_count_desc.format(f"[green]Detected {listeners_count}[/green].")
+            listener_count_desc = (
+                listener_count_desc + f"{colorize_string(color='green', value=f'Detected {listeners_count}')}."
+            )
         else:
-            listener_count_desc = listener_count_desc.format(f"[yellow]Detected {listeners_count}[/yellow].")
+            listener_count_desc = (
+                listener_count_desc + f"{colorize_string(color='yellow', value=f'Detected {listeners_count}')}."
+            )
```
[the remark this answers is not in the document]
vilit1: do think ryan's nit is better

azext_edge/edge/providers/check/base/resource.py
Elsie4ever: feel like this might be useful across service as we add more runtime resource checks, if I recall correctly dataflow also going to check the runtime resource ref in near terms

azext_edge/edge/providers/check/base/resource.py
```diff
@@ -468,3 +470,19 @@ def process_custom_resource_status(
...
+        ref_obj = get_namespaced_secret(secret_name=name, namespace=namespace)
+    elif ref_type == ValidationResourceType.configmap:
+        ref_obj = get_config_map(name=name, namespace=namespace)
+    is_valid = bool(ref_obj)
+
+    if is_valid:
+        text = f"[green]Valid[/green] {ref_type.value} reference {{[green]{name}[/green]}}."
+    else:
+        text = f"[red]Invalid[/red] {ref_type.value} reference {{[red]{name}[/red]}}."
+    return text, is_valid
```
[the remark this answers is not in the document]
Elsie4ever: got it, in that case tho should I just call the get_namespaced_{resource} from the individual service check file? I don't see an example of that in our code, I only see those called in base files, so that's another reason I added a wrapper here

azext_edge/edge/providers/check/mq.py
```diff
@@ -375,27 +493,36 @@ def evaluate_brokers(
...
-                    backend_replicas_colored = f"[red]Actual {backend_redundancy}[/red]."
+                    backend_replicas_colored = f"{colorize_string(color='red', value=f'Actual {backend_redundancy}')}."
                     broker_eval_status = CheckTaskStatus.error.value
 
                 if backend_workers and backend_workers >= 1:
-                    backend_workers_colored = f"[green]Actual {backend_workers}[/green]."
+                    backend_workers_colored = f"{colorize_string(color='green', value=f'Actual {backend_workers}')}."
                 else:
-                    backend_workers_colored = f"[red]Actual {backend_workers}[/red]."
+                    backend_workers_colored = f"{colorize_string(color='red', value=f'Actual {backend_workers}')}."
```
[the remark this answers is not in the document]
Elsie4ever: np! your suggestion make sense, yeah I'm trying to clean up all strings use the colorize_string you introduced earlier(feels like that's better), I can address your suggestion in current pr since it's not in a rush

azext_edge/edge/providers/check/mq.py
```diff
@@ -547,6 +674,328 @@ def evaluate_brokers(
...
+                (0, 0, 0, 8),
+            ),
+        )
+
+        authentications = list(authentications)
+        added_broker_ref_condition = False
+
+        for auth in authentications:
+            auth_name = auth["metadata"]["name"]
+            auth_metadata = auth["metadata"]
```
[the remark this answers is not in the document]
Elsie4ever: oops sorry will change!

azext_edge/edge/providers/check/mq.py
```diff
@@ -547,6 +674,328 @@ def evaluate_brokers(
...
+
+            auth_desc = f"\n- Broker Authentication {{{colorize_string(auth_name)}}}. {ref_display}"
+            check_manager.add_display(
+                target_name=target_authentications,
+                namespace=namespace,
+                display=Padding(auth_desc, (0, 0, 0, 8)),
+            )
+
+            # status
+            status = auth.get("status", {})
```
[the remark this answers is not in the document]
Elsie4ever: yeah good catch, I have always seen status for mq but will add a check here to not call process_custom_resource_status if it's non existed

azext_edge/edge/providers/check/mq.py
Elsie4ever: Updated to use DEFAULT_PADDING in mq.py, but for this function the parent_padding  is passed in and it's a common function being used a lot, + 4 is for sub property under the parent property

azext_edge/edge/providers/check/base/resource.py
```diff
@@ -468,3 +470,30 @@ def process_custom_resource_status(
...
+
+
+def validate_ref(name: str, namespace: str, ref_type: ValidationResourceType) -> Tuple[str, bool]:
+    ref_obj = None
+    if ref_type == ValidationResourceType.secret:
+        ref_obj = get_namespaced_secret(secret_name=name, namespace=namespace)
+    elif ref_type == ValidationResourceType.configmap:
+        ref_obj = get_config_map(name=name, namespace=namespace)
+
+    return bool(ref_obj)
```
c-ryan-k, line 482: If this is in a base module, we may want to raise an exception if `ref_type` is not provided or does not match any of the supported types. My worry is that another module uses this method with a different `ref_type` (like `DataFlowEndpoint`) which always silently returns `False`

azext_edge/edge/providers/check/base/resource.py
```diff
@@ -468,3 +470,30 @@ def process_custom_resource_status(
...
+
+
+def get_valid_references(
+    api: EdgeResourceApi, kind: Union[Enum, str], namespace: Optional[str] = None
+) -> Dict[str, Any]:
```
c-ryan-k, line 487: From the implementation it looks like this returns a `Dict[str, bool]` - if the `key` is what we're really looking for and the `value` is unused, perhaps we could just return a `list[str]` that contains the valid reference names?

something like:

```
if custom_objects:
    return [obj.get('metadata', {}).get('name') for obj in custom_objects]
```

I think we also have a helper class somewhere to just return the `metadata.name` of a resource

azext_edge/edge/providers/check/mq.py
```diff
@@ -184,11 +210,95 @@ def evaluate_broker_listeners(
...
+                    check_manager.add_target_conditions(
+                        target_name=target_listeners,
+                        namespace=namespace,
+                        conditions=[authn_condition],
+                    )
+
+                    authn_eval_value = {"spec.ports[*].authenticationRef": authn}
+
+                    authn_validity = "valid" if authn in valid_authns else "invalid"
+                    authn_color = "green" if authn in valid_authns else "red"
```
c-ryan-k, line 235: nit - we could store this value so we don't have to search the array twice

azext_edge/edge/providers/check/mq.py
```diff
@@ -184,11 +210,95 @@ def evaluate_broker_listeners(
...
+
+                    check_manager.add_target_conditions(
+                        target_name=target_listeners,
+                        namespace=namespace,
+                        conditions=[authz_condition],
+                    )
+
+                    authz_eval_value = {"spec.ports[*].authorizationRef": authz}
+                    authz_validity = "valid" if authz in valid_authzs else "invalid"
+                    authz_color = "green" if authz in valid_authzs else "red"
```
c-ryan-k, line 275: Same here - we only need to check `authz in valid_authzs` once

azext_edge/edge/providers/check/mq.py
```diff
@@ -539,14 +625,349 @@ def evaluate_brokers(
...
+                resource_name=auth_name,
+                added_condition=added_broker_ref_condition,
+            )
+
+            auth_desc = f"\n- Broker Authentication {{{colorize_string(auth_name)}}}. {ref_display}"
+            check_manager.add_display(
+                target_name=target_authentications,
+                namespace=namespace,
+                display=Padding(auth_desc, (0, 0, 0, DEFAULT_PADDING)),
+            )
```
c-ryan-k, line 712: Similar question here to the previous one about `evaluate_broker_ref`: https://github.com/Azure/azure-iot-ops-cli-extension/pull/382/files#r1868441670

azext_edge/edge/providers/check/mq.py
```diff
@@ -539,14 +625,349 @@ def evaluate_brokers(
...
+                resource_name=authz_name,
+                added_condition=added_broker_ref_condition,
+            )
+
+            authz_desc = f"\n- Broker Authorization {{{colorize_string(authz_name)}}}. {ref_display}"
+            check_manager.add_display(
+                target_name=target_authorizations,
+                namespace=namespace,
+                display=Padding(authz_desc, (0, 0, 0, DEFAULT_PADDING)),
+            )
```
c-ryan-k, line 854: Similar question here to the previous one about `evaluate_broker_ref`: https://github.com/Azure/azure-iot-ops-cli-extension/pull/382/files#r1868441670

azext_edge/edge/providers/check/base/resource.py
Elsie4ever: thanks for the catch, looks like I removed the usage of "value" but forgot to update the function😢

azext_edge/edge/providers/check/mq.py
```diff
@@ -250,6 +361,13 @@ def evaluate_broker_listeners(
                 resource_name=listener_name,
             )
 
+        # remove duplicate conditions
+        listener_conditions = check_manager.targets.get(target_listeners, {}).get(namespace, {}).get("conditions", [])
+        listener_conditions = list(set(listener_conditions))
+        check_manager.set_target_conditions(
+            target_name=target_listeners, namespace=namespace, conditions=listener_conditions
+        )
+
```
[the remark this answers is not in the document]
Elsie4ever: got it! I will write a TODO

azext_edge/edge/providers/check/mq.py
```diff
@@ -149,32 +162,45 @@ def evaluate_broker_listeners(
...
-                check_manager=check_manager,
-                status=listener_status_state,
+            listener_desc = f"\n- Broker Listener {{{colorize_string(listener_name)}}}. {ref_display}"
+            check_manager.add_display(
                 target_name=target_listeners,
                 namespace=namespace,
-                resource_name=listener_name,
-                padding=12,
-                detail_level=detail_level,
+                display=Padding(listener_desc, (0, 0, 0, DEFAULT_PADDING)),
```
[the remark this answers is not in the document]
Elsie4ever: good idea will change!

c-ryan-k (approved):

PR #392, opened 2024-09-28
test: targets test content

digimaun:
* Add safety around `packaging` import.


azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -97,8 +95,10 @@ def _check_k8s_version(as_list: bool = False) -> Dict[str, Any]:
     )
 
     try:
+        from packaging import version
+
         version_details: VersionInfo = version_client.get_code()
-    except ApiException as ae:
+    except Exception as ae:
```
vilit1, line 101: Wouldn't this still be (ApiException, ImportError)? If it was more than 3 or so, then I would say global exception is ok (but include comment about what exceptions can pop up) but for two...

vilit1 (approved):

PR #445, opened 2024-11-06
refactor: Change OPCUA AEP connector type value 

digimaun:
* From OpcUa -> Microsoft.OpcUa


vilit1 (approved):

c-ryan-k (approved):
Are we sure this is the final enum? :)

PR #459, opened 2024-12-02
feat: +deployment validations, test coverage, and restructure

digimaun:
* UX improvement by showing release train for deps in `init`, and the AIO version in `create`.


azext_edge/tests/edge/orchestration/test_work_unit.py
```diff
@@ -0,0 +1,775 @@
...
+    def _handle_common(self, request_kpis: RequestKPIs) -> Optional[tuple]:
+        # return (status_code, headers, body)
+        if request_kpis.method == responses.HEAD:
+            if request_kpis.url == ARM_ENDPOINT:
+                self.call_map[CallKey.CONNECT_RESOURCE_MANAGER].append(request_kpis)
+                return (200, {}, None)
+
+        if request_kpis.method == responses.GET:
+            if request_kpis.path_url == f"/subscriptions/{ZEROED_SUBSCRIPTION}/providers":
+                assert request_kpis.params["api-version"] == "2024-03-01"
```
c-ryan-k, line 133: nit - thoughts on moving some of these API-versions to at least a local constant value? Benefit would be easier updates when api-versions change, but cost is that we might lose failed-test specificity if we have them all mapped to the same const

c-ryan-k (approved):

azext_edge/tests/edge/orchestration/test_work_unit.py
digimaun: Good thought, went for a local `ExpectedAPIVersion` enum.

PR #466, opened 2024-12-19
test: Test performance updates

c-ryan-k:
Major changes:
- Workflow cleanup and refactoring for performance
- New workflow solely for containerized e2e tests
- pytest markers `rpsaas` and `e2e` to further define test groupings for pytest
- new tox envs for various integration test configs using pytest markers
- Quicker integration tests due to parallelization with `pytest-xdist`

High-level summary of workflow file changes:
- `int_test.yml`
  - instance delete / redeploy is now only done in workflow, not inside init integration tests
  - removed unused inputs and env vars (`AIO_SP_*` and `CA_*`)
  - removed OIDC token exchange step (no longer necessary on github workflows)
  - updated tox env to run with parallelization params
- `container_int_test.yml` added
  - This is mainly for quickly running e2e tests, no init matrix, default init and create, and runs the containerized `python-e2e-int` tox env
- `create-k3s-cluster/action.yml` action added - can be used in other workflows to quickly start a k3s cluster and setup kubeconfig 
- `publish_test_container_image.yml` modified to use `container_int_test.yml` instead of full `int_test.yml` 
- `Dockerfile` updated to run `python-e2e-int` as well as add parallelization to run tests more quickly.

Code changes:
- `tox.ini`
  - consolidated integration testing environments into one pytest runner to avoid duplicate configuration - `[testenv:python-{init,e2e,rpsaas,edge,all}-int]`
  - under `setenv` you can see the pytest marker inputs for each scenario:
    - `init: SCENARIO="init_scenario_test"`
    - `e2e: SCENARIO="e2e"`
    - `rpsaas: SCENARIO="rpsaas"`
    - `edge: SCENARIO="not rpsaas and not init_scenario_test"`
    - `all: SCENARIO="not init_scenario_test"`
  - added `pytest-randomly` dependency to integration tests to fix test distribution / randomization issue with `pytest-xdist`
- `azext_edge/tests/generators.py`
  - Changed `generate_names` to use insecure random strings (no need for cryptographically "strong" resource names), allowing `pytest-randomly` to assist in seeding randomization per test run (allowing us to parallelize pytest with `pytest-xdist`)
- Marked `check` and `support` int tests with `pytest.mark.e2e` to break them out to a separate tox env to run in our containerized test environment. 
- Marked asset lifecycle and schema registry lifecycle integration tests with `pytest.mark.rpsaas` to break them out to a separate tox env

---

### General Guidelines

Intent for Production

Basic expectations


.github/workflows/int_test.yml
```diff
@@ -284,14 +238,12 @@ jobs:
...
         run: |
-          cd ${{ env.EXTENSION_SOURCE_DIRECTORY }}
           python -m pip install tox
           tox r -vv -e python-init-int --notest
       - name: "Tox test environment setup for integration tests"
         if: ${{ matrix.feature == 'default' && !inputs.use-container }}
         run: |
-          cd ${{ env.EXTENSION_SOURCE_DIRECTORY }}
-          tox r -vv -e python-int --notest
+          tox r -vv -e "python-all-int" --notest
```
vilit1, line 246: porque no more cd extension directory?

.github/workflows/int_test.yml
```diff
@@ -301,9 +253,9 @@ jobs:
           azext_edge_instance: ${{ steps.env_out.outputs.INSTANCE_NAME }}
           azext_edge_init_args: ${{ steps.init.outputs.INIT_ARG }}
           azext_edge_create_args: ${{ steps.init.outputs.CREATE_ARG }}
-          azext_edge_init_redeployment: ${{ steps.init.outputs.REDEPLOY }}
+          # redeploy tested later in this workflow
+          azext_edge_init_redeployment: false
```
vilit1, line 257: tbh None/missing should be treated as false but I am fine with extra caution here

azext_edge/tests/edge/orchestration/resources/test_instance_int.py
```diff
@@ -13,6 +13,9 @@
 
 logger = get_logger(__name__)
 
+# pytest mark for rpsaas (cloud-side) tests
+pytestmark = pytest.mark.rpsaas
+
```
vilit1, line 18: if I add in a kubectl see that the assets are there after create - would this be considered a e2e or still rpsaas?

vilit1: pretend I commented in the asset int test file

vilit1: also instance does have cluster side checks

vilit1 (approved):
update https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/docs/integration-tests.md if needed and also post some example runs

otherwise looks good enough to me

digimaun (approved):

azext_edge/tests/edge/orchestration/resources/test_instance_int.py
c-ryan-k: > if I add in a kubectl see that the assets are there after create - would this be considered a e2e or still rpsaas?

That is an interesting situation, though I think it's probably fine to mark as `rpsaas`. An `rpsaas` test can technically also be marked as `e2e`, but the important distinctions are:

- Tests marked `e2e` are designated to be run in the AIO pipeline, and we want to be careful how we introduce tests with cloud side dependencies to these, since these may interfere / interact with the other AIO pipeline stuff. Namely (at least currently) - we don't want to block PRs for service flakiness, and we don't want to do anything instance/update/init related to the cluster since other tests components from the E2E tests rely on the cluster instance.

- Tests marked `rpsaas` are designated as having a cloud-side dependency. They can be run in other groupings like our normal integration tests, but they have this mark since they are typically longer running and highly dependent on an RP. This is *mostly* for local testing - so you can run the `edge` tox scenario to run quicker tests that only tests cluster-side stuff, or the `rpsaas` tests if you're mostly concerned about the cloud interactions.

A test can of course have both marks, but giving it the `rpsaas` mark will exclude it from the tox `edge` environment, and giving it the `e2e` mark will include it in external tests.

I will add this logic and explanation to the testing document with some other updates to reflect this PR's changes

PR #467, opened 2024-12-19
feat!: Upgrade rev.2

digimaun:
* Checkpoint PR to merge upgrade revision 2.
  * Replaces existing implementation. Module temporarily called upgrade2.
* Upgrade allows configuration of settings, version and release train for any AIO related extension.
* Unit tests adding 98% code coverage, with Test Scenario pattern established in prior work module unit test refresh.
* Adds `parse_kvp_nargs` util function. This should be the go forward for parsing any kvp nargs based args we support. It is superior to the carry over `assemble_nargs_to_dict` in every way. `assemble_nargs_to_dict` should eventually be deleted.
* Table display "The Upgrade Story" by default which shows a clear upgrade strategy.

Follow ups:
- System extension TBD.
- Some handling of non-success starting state of an extension.
- Look into extension status and determine if a status of error can occur with a success http response code. 
- Future of packaging use.
- QOL and nits.


azext_edge/edge/_help.py
```diff
@@ -574,24 +574,27 @@ def load_iotops_help():
...
-        - name: Upgrade the instance and specify the schema registry resource Id.
+        - name: Set extension config settings that apply should be during upgrade.
+           To remove a setting provide the key with no value.
+          text: >
+            az iot ops upgrade --name myinstance -g myresourcegroup --ops-config key1=value1 deletekey
+        - name: Provide an explicit IoT Operations version or release train to upgrade to.
+            Not recommended for typical use cases.
           text: >
-            az iot ops upgrade --name myinstance -g myresourcegroup --sr-resource-id $SCHEMA_REGISTRY_RESOURCE_ID
+            az iot ops upgrade --name myinstance -g myresourcegroup --ops-version x.y.z --ops-train preview
```
vilit1, line 597: do we expect external customers to use this?

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+    def analyze_cluster(self, **override_kwargs: dict) -> "ClusterUpgradeState":
+        with Progress(
+            SpinnerColumn("star"),
+            *Progress.get_default_columns(),
+            "Elapsed:",
+            TimeElapsedColumn(),
+            transient=True,
+            disable=bool(self.no_progress),
+        ) as progress:
+            _ = progress.add_task("Analyzing cluster...", total=None)
```
vilit1, line 102: can you just do `progress.add_task("Analyzing cluster...", total=None)` - the wildcard is extra

azext_edge/edge/_help.py
digimaun: Yes

azext_edge/edge/util/common.py
```diff
@@ -19,6 +19,20 @@
...
+    """
+    Parses kvp nargs into a dict handling values of null and empty string.
+    """
+    result = {}
+    if not kvp_nargs:
+        return result
+
+    for item in kvp_nargs:
+        key, sep, value = item.partition("=")
+        result[key] = value if sep else None
```
vilit1, line 32: This accepts inputs like "potato" -> {"potato": None}

I thought the og command (below) did not take these cases for a reason? Is that reason still valid

(PS: I do like the simplicity of this more - just double checking)

digimaun: Correct. This function would replace most usage of the prior function. The null case is useful and necessary, for example it is used to delete a PATCH configuration setting. It is better in every way. Look at the unit test for all the cases.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
```diff
@@ -0,0 +1,413 @@
...
+    get_base_endpoint,
+    get_mock_resource,
+)
+from .resources.test_instances_unit import (
+    get_instance_endpoint,
+    get_mock_cl_record,
+    get_mock_instance_record,
+)
+
+T = TypeVar("T", bound="UpgradeScenario")
```
vilit1, line 47: prob use something more descriptive than just T (you can do US or UpScen or something if you want to keep it short)

azext_edge/edge/providers/orchestration/common.py
```diff
@@ -40,6 +40,34 @@
...
+    "iotOperations": "ops",
+}
+
+EXTENSION_ALIAS_TO_TYPE_MAP = {
+    "plat": EXTENSION_TYPE_PLATFORM,
+    "osm": EXTENSION_TYPE_OSM,
+    "ssc": EXTENSION_TYPE_SSC,
+    "acs": EXTENSION_TYPE_ACS,
+    "ops": EXTENSION_TYPE_OPS,
+}
```
vilit1, line 65: I like how this becomes a connected single direction triangle

azext_edge/edge/commands_edge.py
```diff
@@ -220,24 +220,52 @@ def create_instance(
...
 
-def upgrade(
+def upgrade_instance(
     cmd,
     resource_group_name: str,
     instance_name: str,
-    schema_registry_resource_id: Optional[str] = None,
     no_progress: Optional[bool] = None,
     confirm_yes: Optional[bool] = None,
+    ops_config: Optional[List[str]] = None,
```
vilit1, line 229: do we have a list of key value pairs that are actually accepted by the service? or do we assume that the service teams are the main consumers for this (and they would know the kv pairs themselves)

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+        self.no_progress = no_progress
+        self.instances = Instances(self.cmd)
+        self.resource_map = self.instances.get_resource_map(
+            self.instances.show(name=self.instance_name, resource_group_name=self.resource_group_name)
+        )
+        self.targets = InitTargets(
+            cluster_name=self.resource_map.connected_cluster.cluster_name, resource_group_name=resource_group_name
+        )
+        self.init_version_map: Dict[str, dict] = {}
+        self.init_version_map.update(self.targets.get_extension_versions())
```
vilit1, line 90: nit: could combine this line with the one above

azext_edge/edge/commands_edge.py
digimaun: Some user kvp's are described in AIO learn documentation intended for end users. Kvp's can be provided and used both internally and externally (power uses, troubleshooting etc) for whatever reasons.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+        return bool(self.override.version) or (
+            self.desired_version[0] and version.parse(self.desired_version[0]) > version.parse(self.current_version[0])
+        )
+
+    def _has_delta_in_train(self) -> bool:
+        return bool(self.override.train) or (
+            self.desired_version[1] and self.desired_version[1].lower() != self.current_version[1].lower()
+        )
+
+    def _has_delta_in_config(self) -> bool:
```
vilit1, line 291: are we going to check the desired config states?

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+    ):
+        self.cmd = cmd
+        self.instance_name = instance_name
+        self.resource_group_name = resource_group_name
+        self.no_progress = no_progress
+        self.instances = Instances(self.cmd)
+        self.resource_map = self.instances.get_resource_map(
+            self.instances.show(name=self.instance_name, resource_group_name=self.resource_group_name)
+        )
+        self.targets = InitTargets(
```
vilit1, line 86: nit: this is neat over importing the template and then crying

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+        self.override_map = override_map
+        self.extension_upgrades = self.refresh_upgrade_state()
+
+    def has_upgrades(self) -> bool:
+        return any(ext_state.can_upgrade() for ext_state in self.extension_upgrades)
+
+    def refresh_upgrade_state(self) -> List["ExtensionUpgradeState"]:
+        ext_queue: List["ExtensionUpgradeState"] = []
+
+        # TODO @digimaun - deterine further pre-checks.
```
vilit1, line 214: are we handling system extensions or not yet (TODO then?)

digimaun: Yea, need input for those decisions. Will include in description.

azext_edge/edge/providers/orchestration/resources/clusters.py
```diff
@@ -58,6 +58,7 @@ def update_cluster_extension(
         cluster_name: str,
         extension_name: str,
         update_payload: dict,
+        **client_kwargs,
```
vilit1, line 61: nit: can prob remove line 54

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
+# coding=utf-8
```
vilit1, line 1: if you delete the og file - may as well just rename this to upgrade.py

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def get_cluster_extensions_endpoint(resource_group_name: str, cluster_name: str = "mycluster") -> dict:
+    resource_path = f"/connectedClusters/{cluster_name}/providers/Microsoft.KubernetesConfiguration/extensions"
+    return get_base_endpoint(
+        resource_group_name=resource_group_name,
+        resource_path=resource_path,
+        resource_provider="Microsoft.Kubernetes",
+        api_version=CLUSTER_EXTENSIONS_API_VERSION,
+    )
```
vilit1, line 84: resource group casing change only for base cluster?

digimaun: It is due to the rest api spec, which leads into the generated client, something I don't want to alter.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -0,0 +1,309 @@
...
+    def _has_delta_in_train(self) -> bool:
+        return bool(self.override.train) or (
+            self.desired_version[1] and self.desired_version[1].lower() != self.current_version[1].lower()
+        )
+
+    def _has_delta_in_config(self) -> bool:
+        return bool(self.override.config)
+
+
+def get_default_table() -> Table:
```
vilit1, line 295: nit: please add screenshot of examples

azext_edge/edge/commands_edge.py
vilit1: would be nice to include these in the long help but not necessary

azext_edge/edge/providers/orchestration/upgrade2.py
vilit1: if you are doing this for the git compare - you can do it as a last push before merge

digimaun: Yes this is to simplify the review.

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: You are talking about some delta between extension X 1.0 -> X 2.0 right? Lets cross that bridge when we get there.

c-ryan-k (approved):

azext_edge/edge/providers/orchestration/upgrade2.py
c-ryan-k: I can't remember if it was, or still is, a thing - but I think there was a linter rule regarding calling functions that return a value without assigning that value to something?

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: Going to do in subsequent PR.

PR #475, opened 2025-01-08
test: add integration testing for `az iot ops secretsync`

vilit1:
tests secret sync operations, including all inputs aside from `--self-hosted-issuer`.

also makes sure github action can test this

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/.github/workflows/int_test.yml
+++ b/.github/workflows/int_test.yml
@@ -88,25 +88,26 @@ jobs:
     env:
       CLUSTER_NAME: "opt${{ github.run_number }}${{ matrix.feature }}"
       INSTANCE_NAME: "inst${{ github.run_number }}${{ matrix.feature }}"
+      KEYVAULT_NAME: "kv${{ github.run_number }}"
     strategy:
       fail-fast: false
       matrix:
-        feature: [custom-input, default, insecure-listener, syncrules, redeploy, trustbundle]
-        runtime-args:
-          - ${{ inputs.runtime-init-args != '' || inputs.runtime-create-args != '' }}
-        exclude:
-          - feature: custom-input
-            runtime-args: false
-          - feature: default
-            runtime-args: true
-          - feature: insecure-listener
-            runtime-args: true
-          - feature: syncrules
-            runtime-args: true
-          - feature: redeploy
-            runtime-args: true
-          - feature: trustbundle
-            runtime-args: true
+        feature: [secretsync]  # [custom-input, default, insecure-listener, syncrules, redeploy, trustbundle]
+        # runtime-args:
+        #   - ${{ inputs.runtime-init-args != '' || inputs.runtime-create-args != '' }}
+        # exclude:
+        #   - feature: custom-input
+        #     runtime-args: false
+        #   - feature: default
+        #     runtime-args: true
+        #   - feature: insecure-listener
+        #     runtime-args: true
+        #   - feature: syncrules
+        #     runtime-args: true
+        #   - feature: redeploy
+        #     runtime-args: true
+        #   - feature: trustbundle
+        #     runtime-args: true
     name: "Run cluster tests"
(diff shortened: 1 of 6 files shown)


vilit1:
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/12719071171

note - the default failure is one ryan is going to fix in support bundle

.github/workflows/int_test.yml
```diff
@@ -236,14 +249,14 @@ jobs:
...
-        run: |
-          python -m pip install tox
-          tox r -vv -e python-init-int --notest
-      - name: "Tox test environment setup for integration tests"
-        if: ${{ matrix.feature == 'default' && !inputs.use-container }}
-        run: |
-          tox r -vv -e "python-all-int" --notest
+          enable-workload-identity: ${{ matrix.feature == 'trustbundle' }}
+      - name: "Az CLI login"
+        if: ${{ matrix.feature == 'trustbundle' && !inputs.use-container }}
```
c-ryan-k, line 254: Do we need this `if` here?

I think since ARC might still somehow take too long, I'm ok with another login step here regardless to ensure the token doesn't timeout before the tests start.

pytest.ini.example
```diff
@@ -23,3 +23,4 @@ markers =
     rpsaas: mark tests that are purely cloud-side
     e2e: mark tests that run in e2e pipeline
     upgrade_scenario_test: mark tests that will run az iot ops upgrade
+    secretsync_test: mark tests that will run az iot ops secretsync
```
c-ryan-k, line 26: FWIW all of these are 'test' marks so i feel like `_test` is a bit redundant

Are there other tests that will eventually run as part of this "scenario"? Wondering if there's a broader term we should use for this mark that signifies it needs WLIF etc rather than just secretsync - but it's probably fine for now.

azext_edge/tests/settings.py
```diff
@@ -16,6 +16,7 @@ class EnvironmentVariables(Enum):
     instance = "azext_edge_instance"
     context_name = "azext_edge_context_name"
     kv = "azext_edge_kv"
+    user_assigned_mi_id = "azext_edge_user_assigned_mi_id"
```
c-ryan-k, line 19: Is this variable passed into / set in a pipeline somewhere, or how is it configured?

tox.ini
```diff
@@ -92,8 +93,9 @@ setenv =
     init: SCENARIO="init_scenario_test"
     e2e: SCENARIO="e2e"
     rpsaas: SCENARIO="rpsaas"
+    secretsync: SCENARIO="secretsync_test"
     edge: SCENARIO="not rpsaas and not init_scenario_test"
```
c-ryan-k, line 97: If `secretsync_test` requires prereqs on the cluster, I'm thinking we might want to modify the `edge` tox scenario to also exclude this test. While it is edge-side, it requires setup that's technically optional for all other edge tests. Thoughts?

pytest.ini.example
vilit1: I can change it to wlif - its prob a better marker

.github/workflows/int_test.yml
vilit1: well originally, i had it as an if cause I def needed it for the trustbundle part but if we may as well have it for everything, will remove the if

tox.ini
vilit1: I was thinking the not rpsaas was going to handle that part already ...

but if we remove the rpsaas/consolidate the markers for said test, then I will need to change edge

azext_edge/tests/settings.py
vilit1: no... 
it currently makes a new mi... 
do I need to add in a pass through... 
can I not...

digimaun (approved):

PR #496, opened 2025-02-18
chore: sync dev with 2502_release_version_rebase

digimaun:


azext_edge/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "1.2.0a1"
+VERSION = "1.2.0b3"
```
vilit1, line 10: going to assume this is the correct version and/or will be fixed later

digimaun: yea will be fixed right after this merge

vilit1 (approved):
Have pointed out my concerns offline

PR #506, opened 2025-03-12
feat: cert management extract subject name and application URI from certificate content

Elsie4ever:
The opcua connector pods will go to crash loop if user accidentally provide different value for --subject-name and --application-uri in comparison to the value provided in certificate

therefore now to prevent this from happening, our CLI tool directly extract those 2 values from the certificate provided by --public-key-file(after decode)

Also deprecated the --subject-name and --application-uri params

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -827,6 +812,45 @@ def _begin_delete_secret(self, keyvault_name: str, secret_name: str):
...
+            # and validate it with the provided values
+            cert_subject_name = certificate.subject.get_attributes_for_oid(NameOID.COMMON_NAME)[0].value
+
+            cert_application_uri = ""
+            extension_value = certificate.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME).value
+
+            for general_name in extension_value:
+                if isinstance(general_name, x509.UniformResourceIdentifier):
+                    cert_application_uri = general_name.value
+
```
digimaun, line 840: Please remind me, what is the benefit around requiring subject name or application uri input? Is it a problem if they are always derived from the certificate?

digimaun:
image
Formatting issue that applies here and subject name if it is deemed both params are still necessary

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: originally it was required since we don't process the certificate content, now that we do we can actually save 2 params for user. Will update, thanks!

Elsie4ever: FYI: derived these 2 values from the certificate was actually the suggestion from connector team as well

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -221,9 +217,9 @@ def client_add(
...
         public_key_file: str,
         private_key_file: str,
-        subject_name: str,
-        application_uri: str,
         overwrite_secret: bool = False,
+        subject_name: Optional[str] = None,
+        application_uri: Optional[str] = None,
         public_key_secret_name: Optional[str] = None,
         private_key_secret_name: Optional[str] = None,
     ) -> dict:
```
digimaun, line 225: The warning as written on line 227 is not conveying (as well) the same point written in the comment on line 226, recommend some wording change.

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -370,8 +368,7 @@ def remove(
         for name in certificate_names:
             if name not in [mapping["targetKey"] for mapping in secret_mapping]:
                 logger.warning(
-                    f"Certificate {name} not found in secretsync resource {secretsync_name}. "
-                    "Skipping removal..."
+                    f"Certificate {name} not found in secretsync resource {secretsync_name}. " "Skipping removal..."
```
digimaun, line 371: Same split string irregularity is present here

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -235,6 +231,9 @@ def client_add(
         # process all the file validations before secret creations
         self._validate_key_files(public_key_file, private_key_file)
```
digimaun, line 233: Were you planning on adding a warning (**if** subject name or app uri are provided), that those are auto-derived from certificates? The tailored warning is more helpful than the generic deprecation msg when using the params. This _ensures_ the user is aware of the change.

azext_edge/edge/util/x509.py
```diff
@@ -73,3 +73,13 @@ def generate_self_signed_cert(valid_days: int = DEFAULT_VALID_DAYS) -> Tuple[byt
...
     return (cert.public_bytes(serialization.Encoding.PEM), key_bytes)
+
+
+def decode_der_certificate(der_data) -> x509.Certificate:
+    # Decodes a DER-encoded X.509 certificate.
+    try:
+        cert = x509.load_der_x509_certificate(der_data, default_backend())
+        return cert
+    except Exception as e:
+        print(f"Error decoding DER certificate: {e}")
```
digimaun, line 84: We shouldn't use print here

digimaun: An exception is immediately raised from the consumer when this function doesn't have a result

azext_edge/edge/util/x509.py
```diff
@@ -73,3 +73,13 @@ def generate_self_signed_cert(valid_days: int = DEFAULT_VALID_DAYS) -> Tuple[byt
     )
 
     return (cert.public_bytes(serialization.Encoding.PEM), key_bytes)
+
+
+def decode_der_certificate(der_data) -> x509.Certificate:
```
digimaun, line 78: This should be `Optional[x509.Certificate]`

azext_edge/edge/util/x509.py
```diff
@@ -73,3 +73,13 @@ def generate_self_signed_cert(valid_days: int = DEFAULT_VALID_DAYS) -> Tuple[byt
...
+    # Decodes a DER-encoded X.509 certificate.
+    try:
+        cert = x509.load_der_x509_certificate(der_data, default_backend())
+        return cert
+    except Exception as e:
+        print(f"Error decoding DER certificate: {e}")
+        return None
```
digimaun, line 85: Don't need an explicit `return None` here, but if you like the return (I understand) just use `return`.

azext_edge/tests/edge/orchestration/resources/connector/opcua/test_opcua_certs_client_unit.py
```diff
@@ -231,20 +237,55 @@ def test_client_add(
             {},
             "/fake/path/pubkey.der",
             "/fake/path/prikey.pem",
+            "subjectname",
+            "uri",
             "Public key file name pubkey and private key file name prikey must match.",
         ),
+        # subject name not found
+        (
```
vilit1, line 245: personally I like using dicts for params, especially when the tuples (like here) get very long.

a dict should help with avoiding incorrect order/missing params.

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: oh got it, let me make sure to  overwrite the auto derived warning, thanks!

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: 😢I don't get why it reformat to this on save, will update

Elsie4ever: even more strange, I do see this in 2 lines without modification😐
image

azext_edge/tests/edge/orchestration/resources/connector/opcua/test_opcua_certs_client_unit.py
Elsie4ever: yeah good idea, will add this to a TODO, since this would be a bigger change also include the other 2 test files

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
digimaun: Think it's because of string continuation on multiple lines, with formatter setting having a longer line length. And the formatter doesn't want to combine string content.

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -221,20 +217,33 @@ def client_add(
...
         overwrite_secret: bool = False,
+        subject_name: Optional[str] = None,
+        application_uri: Optional[str] = None,
         public_key_secret_name: Optional[str] = None,
         private_key_secret_name: Optional[str] = None,
     ) -> dict:
         # inform user if the provided cert was issued by a CA, the CA cert must be added to the issuers list.
-        logger.warning("Please ensure the certificate must be added to the issuers list if it was issued by a CA.")
+        logger.warning(
+            "If this certificate was issued by a CA, then please ensure that the certificate is "
```
digimaun, line 228: "...please ensure that the **CA** certificate is..."  <-- for specificity

azext_edge/edge/util/x509.py
digimaun: Re-opened this because I'm wondering about `der_data: bytes`, is adding bytes type hint causing an issue?

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -827,6 +824,38 @@ def _begin_delete_secret(self, keyvault_name: str, secret_name: str):
...
+
+        # Get the subject name and application uri from the certificate
+        # and validate it with the provided values
+        cert_subject_name = certificate.subject.get_attributes_for_oid(NameOID.COMMON_NAME)[0].value
+
+        cert_application_uri = ""
+        extension_value = certificate.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME).value
+
+        for general_name in extension_value:
+            if isinstance(general_name, x509.UniformResourceIdentifier):
```
digimaun, line 850: What if there are multiple URIs defined? How is that handled?

azext_edge/edge/util/x509.py
Elsie4ever: no, my eyes missed something. will update!😋

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: talked offline for the solution here

azext_edge/edge/_help.py
```diff
@@ -1544,13 +1544,11 @@ def load_iotops_help():
         - name: Add a client certificate.
```
digimaun, line 1544: In the long summary we should explain the dynamic of the command. Explain the auto derivation, and what happens when subject name and app uri are explicitly provided.

azext_edge/edge/_help.py
```diff
@@ -1560,8 +1558,6 @@ def load_iotops_help():
             --private-key-file "newopc.pem"
             --public-key-secret-name public-secret-name
             --private-key-secret-name private-secret-name
-            --subject-name "aio-opc-opcuabroker"
```
digimaun, line 1563: There should be an example with explanation for explicit subject name and app uri dynamic

azext_edge/edge/params.py
```diff
@@ -787,13 +787,13 @@ def load_iotops_arguments(self, _):
         context.argument(
             "subject_name",
             options_list=["--subject-name", "--sn"],
-            help="The subject name string embedded in the application instance certificate."
+            help="The subject name string embedded in the application instance certificate. "
```
digimaun, line 790: Explain the purpose e.g. "When provided the input will be validated against the value embedded in the certificate."

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -221,20 +217,27 @@ def client_add(
...
         overwrite_secret: bool = False,
+        subject_name: Optional[str] = None,
+        application_uri: Optional[str] = None,
         public_key_secret_name: Optional[str] = None,
         private_key_secret_name: Optional[str] = None,
     ) -> dict:
         # inform user if the provided cert was issued by a CA, the CA cert must be added to the issuers list.
-        logger.warning("Please ensure the certificate must be added to the issuers list if it was issued by a CA.")
+        logger.warning(
+            "If this certificate was issued by a CA, then please ensure that the CA certificate is "
```
digimaun, line 228: Maybe in a later iteration there can be logic on when this warning is raised. Something like determining if the cert was self-signed or not, if not raise the warning.

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -827,6 +818,71 @@ def _begin_delete_secret(self, keyvault_name: str, secret_name: str):
...
+    def _extract_cert_content(
+        self,
+        public_key_file: str,
+    ) -> Tuple[str, str]:
+        from cryptography import x509
+        from cryptography.x509.oid import NameOID, ExtensionOID
+
+        der_data = read_file_content(file_path=public_key_file, read_as_binary=True)
+        certificate = decode_der_certificate(der_data)
+
```
digimaun, line 832: What do you think about a later iteration validation for cert expiry?

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: good idea, will add a TODO

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
Elsie4ever: good idea, since we already decoded the cert content, will add TODO

digimaun (dismissed):

azext_edge/edge/_help.py
```diff
@@ -1540,17 +1540,18 @@ def load_iotops_help():
             and secretsync 'aio-opc-ua-broker-client-certificate' will be created
             if not found. The newly added certificate will replace the existing
             certificate if there is any.
+            Note: Subject name and application uri will be auto derived from the provided
```
digimaun, line 1543:
```suggestion
            Note: the subject name and application uri will be auto derived from the provided
            certificate. Optional parameters may be used to validate the respective values
            meet expectations before the operation proceeds.
```

digimaun (approved):

digimaun (approved):

PR #528, opened 2025-04-04
feat: broker listener port add & remove.

digimaun:


azext_edge/edge/providers/orchestration/resources/brokers.py
```diff
@@ -85,6 +94,201 @@ def create(
...
+            cm_config["san"]["dns"] = tls_auto_san_dns
+        if tls_auto_san_ip:
+            cm_config["san"]["ip"] = tls_auto_san_ip
+        if tls_auto_secret_name:
+            cm_config["secretName"] = tls_auto_secret_name
+
+        if tls_manual_secret_ref:
+            man_config["secretRef"] = tls_manual_secret_ref
+
+        if all([cm_config, man_config]):
```
YingXue, line 140: why not have this check happen earlier in line 111 so it fails early?

digimaun: The current placement is simpler (more readable) and ensures the check is performed after all potential configs have been gathered. It handles cases where different tls_auto_* keys are used without additional checks/conditions. The performance difference between the approaches is negligible as building the dictionaries in this scenario and comparing truthiness is negligible impact to performance.

YingXue: Oh I see! Coming from a UI lens, user don't even see the input because they choose either manual or auto before they can proceed. But because of the nature of CLI that all the input has been provided, you need to compare not a single value contrast

c-ryan-k (approved):

PR #546, opened 2025-04-30
feat: Adds `--check-cluster` option to init, pre-check updates

c-ryan-k:
- Added `--check-cluster` option to `ops init`
image

- Updated pre-check values to reflect new checks
  - Updated k8s min version to 1.29
  - Removed storage space requirement
  - Added min node kernel version (ACS / init checks only)
  - Added check for storage_classes (ACS / init checks only)
- Added error parsing for pre-checks to init output
- Added helper method to get merged ACS config to reduce duplicate template parsing
image
- Added tests for pre-check methods, validation and deployment checks
 ---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/params.py
```diff
@@ -535,6 +535,14 @@ def load_iotops_arguments(self, _):
                 arg_type=get_three_state_flag(),
                 help="Ensure the latest IoT Ops CLI is being used, raising an error if an upgrade is available.",
             )
+            # Cluster Precheck opt-in
+            context.argument(
+                "cluster_checks",
```
digimaun, line 540: I think we should put the option in preview for the initial release.

azext_edge/edge/params.py
```diff
@@ -535,6 +535,14 @@ def load_iotops_arguments(self, _):
...
+                options_list=["--cluster-checks"],
```
digimaun, line 541: --cluster-checks over --check-cluster?

azext_edge/edge/params.py
```diff
@@ -535,6 +535,14 @@ def load_iotops_arguments(self, _):
...
+                arg_type=get_three_state_flag(),
+                help="Enforce a check for minimum cluster requirements before bootstrapping.",
```
digimaun, line 543: Let's add an example to init help that utilizes new option.

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -24,22 +25,54 @@
...
+            for namespace in check["targets"][target]:  # for all prechecks, namespace is currently _all_
+                for idx, check_eval in enumerate(check["targets"][target][namespace]["evaluations"]):
+                    if check_eval["status"] not in NON_ERROR_STATUSES:
+                        # TODO - relies on same order and count of conditions / evaluations
+                        expected_condition = (
+                            check["targets"][target][namespace]["conditions"][idx]
+                            if idx < len(check["targets"][target][namespace]["conditions"])
+                            else None
+                        )
+                        if not errors.get(target):
```
digimaun, line 42: errors could be `errors = defaultdict(list)` and then this condition is not necessary.

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -24,22 +25,54 @@
...
+                            errors[target] = []
+                        errors[target].append(f"Expected: '{expected_condition}', Actual: '{check_eval['value']}'")
```
digimaun, line 44: Now (opposed to PR screenshot), curlies on both expected condition and actual value. Can we see an up-to-date screenshot?

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -24,22 +25,54 @@
...
-    as_list: bool = False,
-) -> None:
-    result["preDeployment"] = []
+def validate_cluster_prechecks(acs_config: Optional[dict] = None) -> None:
+    pre_checks = check_pre_deployment(acs_config=acs_config)
+    errors = {}
+    for check in pre_checks:
+        for target in check["targets"]:
+            for namespace in check["targets"][target]:  # for all prechecks, namespace is currently _all_
+                for idx, check_eval in enumerate(check["targets"][target][namespace]["evaluations"]):
```
digimaun, line 34: remind me what does idx reference?

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -134,3 +164,51 @@ def _check_k8s_version(as_list: bool = False) -> Dict[str, Any]:
         )
 
     return check_manager.as_dict(as_list)
+
+
+def _check_storage_classes(acs_config: dict, as_list: bool = False) -> Dict[str, Any]:
+    from kubernetes.client.models import V1StorageClassList
+
+    expected_classes = acs_config.get("feature.diskStorageClass", [])
```
digimaun, line 172: does default empty list make sense when the expected value is a comma delimited value str?

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -134,3 +164,51 @@ def _check_k8s_version(as_list: bool = False) -> Dict[str, Any]:
...
+        check_manager.add_target_eval(
+            target_name=target,
+            status=CheckTaskStatus.success.value,
+            value={"len(cluster/storage-classes)": len(storage_classes.items)},
+        )
+
+        expected_class_names = expected_classes.split(",")
+        storage_class_names = [sc.metadata.name for sc in storage_classes.items]
+        matches = [sc for sc in storage_class_names if sc in expected_class_names]
+        storage_status = CheckTaskStatus.success if len(matches) else CheckTaskStatus.error
```
digimaun, line 207: could omit `len` and just do `if matches`

azext_edge/edge/providers/check/base/node.py
```diff
@@ -52,42 +53,51 @@ def check_nodes(as_list: bool = False) -> Dict[str, Any]:
...
-        show_header=True, header_style="bold", show_lines=True, caption_justify="left"
-    )
+    table = Table(show_header=True, header_style="bold", show_lines=True, caption_justify="left")
     for column_name, justify in [
         ("Name", "left"),
         ("Architecture", "right"),
+        *([("Kernel version", "right")] if check_acsa_node_version else []),
         ("CPU (vCPU)", "right"),
         ("Memory (GB)", "right"),
-        ("Storage (GB)", "right"),
```
digimaun, line 81: lets chat about this one

azext_edge/edge/providers/check/base/deployment.py
c-ryan-k: No curlies, that's just the format string, might just be github highlighting that makes it look like text:
image

azext_edge/edge/providers/orchestration/work.py
```diff
@@ -368,6 +378,19 @@ def _do_work(self):
                     verify_write_permission_against_rg(
                         subscription_id=self.subscription_id, resource_group_name=self._targets.resource_group_name
                     )
+                if self._check_cluster:
+                    load_config_context(context_name=self._context_name)
+                    validate_cluster_prechecks(
```
digimaun, line 383: What if we did a dynamic kwargs strategy to pass into validate_cluster_prechecks(), removing inline if else conditionals, whose benefit grows as more parameters are used at the same time.

something like
```python
cluster_precheck_kwargs = {}
if self._targets.enable_fault_tolerance:
  cluster_precheck_kwargs["acs_config"] =  get_merged_acs_config(...)
validate_cluster_prechecks(**cluster_precheck_kwargs)
```

azext_edge/tests/edge/checks/base/test_deployment_checks_unit.py
```diff
@@ -0,0 +1,266 @@
...
+            ],
+            True,
+        ),
+    ],
+    ids=["all_success", "errors"],
+)
+@pytest.mark.parametrize(
+    "acs_config",
+    [
+        {"feature.diskStorageClass": "default,local-path"},
```
digimaun, line 189: What about a case for `{"feature.diskStorageClass": ""}` ?

digimaun: I see https://github.com/Azure/azure-iot-ops-cli-extension/blob/ea2f9a3da723fe8e6312876874982250ad90471a/azext_edge/tests/edge/checks/base/test_deployment_checks_unit.py#L47 has a case for singleclass

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -4,42 +4,82 @@
...
+    for check in pre_checks:
+        for target in check["targets"]:
+            for namespace in check["targets"][target]:  # for all prechecks, namespace is currently _all_
+                for idx, check_eval in enumerate(check["targets"][target][namespace]["evaluations"]):
+                    if check_eval["status"] not in NON_ERROR_STATUSES:
+                        # TODO - relies on same order and count of conditions / evaluations
+                        expected_condition = (
+                            check["targets"][target][namespace]["conditions"][idx]
+                            if idx < len(check["targets"][target][namespace]["conditions"])
+                            else None
```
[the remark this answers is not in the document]
c-ryan-k: I think the difference to me here is that we're only showing errors for evaluation objects (`actual`) with an error status - this would be the case where we know we have an error value (as an eval), but somehow we don't have a corresponding condition ('expected') for that eval.

To me this is more of a developer error that we can patch later by trueing up the conditions/evals - rather than an error object we'd want to skip because we don't know what the expected value is... what if we just change it to "Unknown condition"?

It boils down to whether we:
- don't show an error if we can't determine what the expected value is
- show an error that may be confusing / difficult to solve because there's no 'set' value to fix

At some point I want to consolidate 'conditions' with 'evals' to avoid this, but that'll likely come in a future PR.

azext_edge/edge/providers/orchestration/work.py
c-ryan-k: to kwarg or not to kwarg, that is the question :)

i like this strategy, will push an update soon with `check_kwargs` being built in a helper method to organize the logic

azext_edge/tests/edge/checks/base/test_deployment_checks_unit.py
c-ryan-k: This brings up an interesting question - What is the expected behavior if a customer overrides the storageclass property with an empty value?

`az iot ops init -g test --cluster test --check-cluster --acs-config feature.diskStorageClass="" `

Right now the result is an error, but the condition looks odd:

```
Cluster readiness pre-checks failed:
        Target 'cluster/storage-classes':
                Expected: 'contains(cluster/storage-classes, any())', Actual: 'local-path'
```

Should we throw a validation error if they're trying to `check` an init without a defined storageclass in the config? Or should we pass / continue since we don't know what to check for and therefore... anything is considered a success?

My guess is that this might break some config down the line and therefore we should error.

azext_edge/edge/providers/orchestration/work.py
```diff
@@ -368,6 +378,10 @@ def _do_work(self):
                     verify_write_permission_against_rg(
                         subscription_id=self.subscription_id, resource_group_name=self._targets.resource_group_name
                     )
+                if self._check_cluster:
+                    load_config_context(context_name=self._context_name)
```
digimaun, line 382: I think `load_config_context()` should not happen in the work module.

c-ryan-k: while we're on the subject of what goes where, regarding the [comment ](https://github.com/Azure/azure-iot-ops-cli-extension/pull/546#discussion_r2071889034) about storage class validation - does that make more sense to bring up in the `cluster_kwarg` building, or should we have that validation fire in the `_check_storage_classes` function?

azext_edge/tests/edge/checks/base/test_deployment_checks_unit.py
c-ryan-k: does this make more sense to raise as a `ValidationError` in the `cluster_kwarg` building, or should we have that validation fire in the `_check_storage_classes` function?

Or should we just output the standard pre-check error, rather than raising a `ValidationError`?

digimaun: Let's validate service behavior both when there is an empty or null value for feature.diskStorageClass. Want to make sure it doesn't default to something.

In the case it doesn't default (and --enable-fault-tolerance is not True), user theoretically shouldn't need to use --check-cluster for a validation error since we presume it will error without one.

digimaun (approved):

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -4,42 +4,86 @@
...
+    context_name = kwargs.get("context_name")
+    load_config_context(context_name=context_name)
+
+    acs_config = kwargs.get("acs_config")
+    storage_space_check = kwargs.get("storage_space_check")
+
+    pre_checks = check_pre_deployment(acs_config=acs_config, storage_space_check=storage_space_check)
+    errors = defaultdict(list)
+    for check in pre_checks:
+        for target in check["targets"]:
```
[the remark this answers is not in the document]
c-ryan-k: It should, yeah - it's defined as an empty object on [`CheckManager.__init__`](https://github.com/c-ryan-k/azure-iot-ops-cli-extension/tree/check/azext_edge/edge/providers/check/base/check_manager.py#L69)

digimaun (approved):

PR #558, opened 2025-05-08
test: dataflow profile integration tests

shahm5:


digimaun:
We should see this running in the IT workflow

azext_edge/tests/edge/orchestration/resources/test_dataflow_profiles_int.py
```diff
@@ -0,0 +1,109 @@
...
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+import pytest
+
+from ....generators import generate_random_string
+from ....helpers import run
+
+# pytest mark for rpsaas (cloud-side) tests
+pytestmark = pytest.mark.rpsaas
```
vilit1, line 13: use the edge marker for cluster connections

vilit1: nvm - keep as is to avoid conflicts with support bundle tests

PR #562, opened 2025-05-10
feat!: secretsync enable should raise error if role assignment fails + new functionality

digimaun:
* Added support for tags on default SPC via --tags.
* Added support for custom role via --custom-role-id.

### DIFF
--- a/azext_edge/edge/_help.py
+++ b/azext_edge/edge/_help.py
@@ -1050,15 +1050,22 @@ def load_iotops_help():
             and role assignments (Key Vault Reader, Key Vault Secrets User) of the managed identity
             against the target Key Vault.
 
+            The flow starts with ensuring Key Vault role assignments, applying them if they don't exist.
+            If necessary you can provide a custom role via --custom-role-id to use instead.
+
         examples:
         - name: Enable the target instance for Key Vault secret sync.
           text: >
             az iot ops secretsync enable --instance myinstance -g myresourcegroup
             --mi-user-assigned $UA_MI_RESOURCE_ID --kv-resource-id $KEYVAULT_RESOURCE_ID
-        - name: Same as prior example except flag to skip Key Vault role assignments.
+        - name: Usage of flag to skip Key Vault role assignments.
           text: >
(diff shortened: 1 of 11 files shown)


azext_edge/edge/_help.py
```diff
@@ -1050,12 +1050,21 @@ def load_iotops_help():
...
+            via --custom-role-id can be used in-place of the built-in roles. Or the --skip-ra flag can
+            be used to skip role assignments.
+
         examples:
         - name: Enable the target instance for Key Vault secret sync.
           text: >
             az iot ops secretsync enable --instance myinstance -g myresourcegroup
             --mi-user-assigned $UA_MI_RESOURCE_ID --kv-resource-id $KEYVAULT_RESOURCE_ID
-        - name: Same as prior example except flag to skip Key Vault role assignments.
+        - name: Enable secret sync and apply tags when creating the default secret provider class.
```
Elsie4ever, line 1063: is there an example of using `--custom-role-id`?

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
```diff
@@ -411,3 +449,230 @@ def test_parse_feature_kvp_nargs(feature_scenario: Optional[dict]):
...
     assert result == feature_scenario.get("expected", {}), f"Expectation failure for: {feature_scenario.get('case')}"
+
+
+@pytest.mark.parametrize(
+    "spc_name",
+    [None, generate_random_string()],
+)
+@pytest.mark.parametrize(
+    "skip_role_assignments",
+    [None],
```
Elsie4ever, line 460: if these parameters(and below) only have one value, should we assign them within the test function instead?

digimaun: They will have more, the reason it is like this is the test is a TODO.

Elsie4ever: ok that make sense

Elsie4ever (approved):

PR #579, opened 2025-06-06
feat: increment min core cli version

digimaun:


c-ryan-k (approved):
Before we release it's fine (and likely better) to leave as-is, but post-release we should update the readme to bump this there too :)

PR #581, opened 2025-06-10
fix: upgrade retries.

digimaun:
The behavior is:
- If status_code < 400: don't retry
- Else if Retry-After present: retry
- Else: retry based on the safe status code list ([408, 429, 500, 502, 503, 504])


vilit1 (approved):
lgtm

PR #594, opened 2025-06-28
feat: secretsync mgmt updates around default SPC

digimaun:
* Change ops secretsync enable & disable to handle the new instance attribute.
* Clean up tech debt by adding `connected_cluster.get_cl_resources_by_type` intends to replace `instance.find_existing_resources`. Less API calls will be necessary.
* Parallel deletion (vs serially) of resources by category (secretsync, spc) on disablement.
* secretsync list shows both spcs and secretsync resources.
* Increments AIO extension.

### DIFF
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -149,6 +149,7 @@ class Instances(Queryable):
         **kwargs: dict,
     ) -> dict:
         instance = kwargs.pop("instance", None) or self.show(name=name, resource_group_name=resource_group_name)
+        status_text = kwargs.pop("status_text", "Working...")
 
         if description:
             instance["properties"]["description"] = description
@@ -162,7 +163,7 @@ class Instances(Queryable):
         if tags or tags == {}:
             instance["tags"] = tags
 
-        with console.status("Working..."):
+        with console.status(status_text):
             poller = self.iotops_mgmt_client.instance.begin_create_or_update(
                 instance_name=name,
                 resource_group_name=resource_group_name,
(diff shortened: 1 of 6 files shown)


azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -375,50 +390,56 @@ def disable_secretsync(
...
-            resource_map = self.get_resource_map(instance)
-            cl_resources = resource_map.connected_cluster.get_aio_resources(
-                custom_location_id=instance["extendedLocation"]["name"]
-            )
-            secretsync_spcs = self.find_existing_resources(cl_resources=cl_resources, resource_type=SPC_RESOURCE_TYPE)
-            secretsyncs = self.find_existing_resources(
-                cl_resources=cl_resources, resource_type=SECRET_SYNC_RESOURCE_TYPE
+        instance: dict = self.show(name=name, resource_group_name=resource_group_name)
+        # remove the default secret provider class reference
+        default_spc_ref = instance["properties"].pop("defaultSecretProviderClassRef", None)
```
Elsie4ever, line 399: does it means in 2507 the default spc resource stored differently place then other spc resources? like below we are getting spc and ssc resources using `get_cl_resources_by_type`, will that included the default spc resource as well? thanks!

Elsie4ever: oh after looking again, is this `instance["properties"]["defaultSecretProviderClassRef"]` just a place to store the ref of the default spc, but to actually delete this default resource we would still need to call `ssc_mgmt_client.azure_key_vault_secret_provider_classes.begin_delete`?

Elsie4ever (approved):

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: Yes in 2507 the default SPC resource is stored differently than other spc resources, but AIO tooling will be concerned only with the default SPC. Right, `get_cl_resources_by_type` will include the default SPC resource if `microsoft.secretsynccontroller/azurekeyvaultsecretproviderclasses` is passed in as a desired type.

digimaun: > oh after looking again, is this instance["properties"]["defaultSecretProviderClassRef"] just a place to store the ref of the default spc, but to actually delete this default resource we would still need to call ssc_mgmt_client.azure_key_vault_secret_provider_classes.begin_delete?

Yeah, you got it. Also, just because the reference is set doesn't mean the resource exists so using the akv client get (or via get_cl_resources_by_type) works as a way to verify. I wouldn't delete the default SPC via opcua commands though -- that is handled with `ops secretsync disable`.

PR #603, opened 2025-07-08
fix(test): do not fail pod status check if pod phase is succeeded

c-ryan-k:
Int tests are currently failing for Akri checks because they have pod states that are `False` but the code is designed to[ mark those pods as successful if the overall pod phase is `succeeded`.](https://github.com/Azure/azure-iot-ops-cli-extension/blob/97189119afd9565d425071ee8fd106f51df65e56/azext_edge/edge/providers/check/base/pod.py#L129)

This change fixes the test to reflect the correct code result, but I'm curious if we want to take a look at this method in the future to make checks a bit more precise.

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/tests/edge/checks/int/helpers.py
+++ b/azext_edge/tests/edge/checks/int/helpers.py
@@ -6,6 +6,7 @@
 
 from typing import Any, Dict, List, Optional, Tuple
 from azure.cli.core.azclierror import CLIInternalError
+from azext_edge.edge.common import PodState
 from azext_edge.edge.providers.edge_api.base import EdgeResourceApi
 from ....helpers import (
     PLURAL_KEY,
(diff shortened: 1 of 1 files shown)


digimaun (approved):

digimaun:
> This change fixes the test to reflect the correct code result, but I'm curious if we want to take a look at this method in the future to make checks a bit more precise.

More precise checks are appealing, what is the downside

c-ryan-k:
> More precise checks are appealing, what is the downside

I don't think there's a downside, just time to investigate.

It may be situational for some pods which fields can / should be truthy/falsy etc. - I believe this was a patch due to some pods being marked as `error` status because of a readiness / schedule availability field being marked as `false` since they were single-run jobs? My memory is a bit fuzzy, @Elsie4ever do you remember the details behind this fix?

PR #611, opened 2025-07-11
feat: meso support bundle and dataflow v1beta1 updates

c-ryan-k:
- Adds new dataflow endpoint type (openTelemetry) and registry endpoint checks
- New Meso label for support bundle items
---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/check/dataflow.py
```diff
@@ -53,9 +61,13 @@
 dataflow_profile_check_name = "evalDataflowProfiles"
 dataflow_profile_check_desc = "Evaluate Dataflow Profiles"
 
+registry_endpoint_check_name = "evalRegistryEndpoints"
+registry_endpoint_check_desc = "Evaluate Registry Endpoints"
+
 dataflow_target = "dataflows.connectivity.iotoperations.azure.com"
 dataflow_endpoint_target = "dataflowendpoints.connectivity.iotoperations.azure.com"
 dataflow_profile_target = "dataflowprofiles.connectivity.iotoperations.azure.com"
+registry_endpoint_target = "registryendpoints.connectivity.iotoperations.azure.com"
```
digimaun, line 70: registry endpoints are also used for akri connector templates

azext_edge/edge/providers/check/dataflow.py
```diff
@@ -1565,3 +1724,119 @@ def evaluate_dataflow_profiles(
...
+                namespace=namespace,
+                status=status,
+                resource_name=registry_endpoint_name,
+                resource_kind=DataflowResourceKinds.REGISTRYENDPOINT.value,
+                detail_level=detail_level,
+                padding=INNER_PADDING,
+            )
+
+            # evaluate host condition - endsWith(spec.host, 'azurecr.io')
+            host_status = CheckTaskStatus.success if host.endswith("azurecr.io") else CheckTaskStatus.error
```
digimaun, line 1803: does casing matter with azurecr.io ?

azext_edge/edge/providers/support/dataflow.py
```diff
@@ -22,29 +22,19 @@
 
 logger = get_logger(__name__)
 
-DATAFLOW_NAME_LABEL = NAME_LABEL_FORMAT.format(label=DATAFLOW_API_V1.label)
-DATAFLOW_DIRECTORY_PATH = DATAFLOW_API_V1.moniker
+DATAFLOW_NAME_LABEL = NAME_LABEL_FORMAT.format(label=DATAFLOW_ACTIVE_API.label)
```
digimaun, line 25: Only active API / v1b1? I suppose we are about to talk about this.

azext_edge/edge/providers/support/meso.py
```diff
@@ -23,57 +23,87 @@
...
+        results.extend(process_cluster_roles(
+            directory_path=MESO_DIRECTORY_PATH,
+            label_selector=label_selector,
+        ))
+    # Fetch specific cluster role by name
+    results.extend(process_cluster_roles(
         directory_path=MESO_DIRECTORY_PATH,
-        label_selector=MESO_NAME_LABEL,
-    )
+        field_selector="metadata.name=aio-observability-operator-manager-role",
```
digimaun, line 94: should we capture this value as a const? (existing precedence in sb)

azext_edge/edge/providers/support_bundle.py
```diff
@@ -18,7 +18,6 @@
     AZUREMONITOR_API_V1,
     CERTMANAGER_API_V1,
     CLUSTER_CONFIG_API_V1,
-    CLUSTER_CONFIG_API_V1B1,
```
digimaun, line 21: Does this API need to be deleted in the client? No longer in use?

digimaun (approved):
any screnshots for the new displays?

azext_edge/edge/providers/check/dataflow.py
c-ryan-k: confirmed, it does - essentially case-sensitive `*.azurecr.io`

azext_edge/edge/providers/support/dataflow.py
c-ryan-k: In this instance V1 and V1B1 have the same label and moniker, but it's a good point that we may need to keep track of them separately.

In this instance/pattern, `moniker` lets us keep the folder in the support bundle tied to the most current ResourceAPI definition (so in this case still `dataflow`), and changing the moniker of the active API would allow us to change where the support bundle output goes.

Regarding the `label` - I don't know if we expect those to change much between API-versions (though Meso is one example), but in that case yes we'd have to loop through multiples.

As far as I know, we can also have multiple labels per-API-version, so maybe in the future we expand a resourceAPI to have multiple labels, or we just keep track of separate labels for the service (like the Meso implementation here) and assume that both labels may exist in any API version

azext_edge/edge/providers/support_bundle.py
c-ryan-k: This was actually just an attempted fix/ debugging stage for our failing support bundle integration tests - to pare down the "new" api-versions to see if we could get tests passing incrementally.

After our discussion yesterday, I'm going to raise another PR after this one that adds it back (with a few extras) and then we'll tackle the tests once we have the support bundle capturing everything new.

c-ryan-k:
> any screnshots for the new displays?

Registry endpoints look the same regardless of detail-level:
image

Telemetry endpoints summary:
image
Detail:
image
Verbose:
image

c-ryan-k:
Going to merge this and add the following in a new PR:
- const label for `aio-observability-operator-manager-role`
- Add back all v1beta1 APIs to support bundle

azext_edge/edge/providers/check/dataflow.py
vilit1: Are they shared or different?

PR #619, opened 2025-07-21
feat: ops identity assign makes opt-out role assignment for schema registry

digimaun:


azext_edge/edge/_help.py
```diff
@@ -1935,7 +1935,10 @@ def load_iotops_help():
         type: command
         short-summary: Assign a user-assigned managed identity with the instance.
         long-summary: |
-            This operation includes federation of the identity.
+            This operation includes federation of the identity for the applicable purpose.
+
+            When --usage 'schema' is input, by default, a role assignment of the identity against the
```
vilit1: when --usage 'schema' is used

azext_edge/edge/params.py
```diff
@@ -143,7 +143,7 @@ def load_iotops_arguments(self, _):
             "custom_role_id",
             options_list=["--custom-role-id"],
             help="Fully qualified role definition Id in the following format: "
-            "/subscriptions/{subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/{roleId}",
+            "/subscriptions/{subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/{roleId}.",
```
vilit1, line 146: nit: im iffy on whether or not to have the period - it helps with grammar but at the same time can be misinterpreted as having to have the dot or copy pasted wrong. I will say keep it for now so will just resolve this

azext_edge/edge/_help.py
digimaun: That's what it was in the last commit but double "use" was a nit to avoid.

azext_edge/edge/_help.py
```diff
@@ -1944,6 +1947,16 @@ def load_iotops_help():
...
             az iot ops identity assign --name myinstance -g myresourcegroup --mi-user-assigned $UA_MI_RESOURCE_ID --usage schema
+        - name: Assign and federate a desired user-assigned managed identity for use with schema registry with a
+            custom role to be used for the identity role assignment.
+          text: >
+            az iot ops identity assign --name myinstance -g myresourcegroup --mi-user-assigned $UA_MI_RESOURCE_ID --usage schema
+            --custom-sr-role-id $CUSTOM_ROLE_ID
+        - name: Assign and federate a desired user-assigned managed identity for use with schema registry but
+            skip the role assignment step of the operation.
+          text: >
+            az iot ops identity assign --name myinstance -g myresourcegroup --mi-user-assigned $UA_MI_RESOURCE_ID --usage schema
```
vilit1, line 1958: These two cancel each other out right? Is there a reason for having this?

vilit1: nvm stared at the code hard enough to understand

vilit1 (approved):
Everything looks good

PR #621, opened 2025-07-21
feat: opcua cert mgmt supports default SPC + test fixes.

digimaun:
* Goal for the PR is to make the minimum set of changes to support the default SPC while retaining backwards compatibility.


azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -665,10 +670,10 @@ def _add_secrets_to_spc(
         else:
             spc["properties"]["objects"] = spc_object
 
-        with console.status(f"Adding secret reference in Secret Provider Class resource {OPCUA_SPC_NAME}..."):
+        with console.status(f"Adding secret reference in Secret Provider Class resource {spc['name']}..."):
```
Elsie4ever, line 673: why sometimes we use `spc['name']` sometimes `spc["name"]`?

Elsie4ever:
code lgtm! thanks for taking care of this

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
digimaun: spc['name'] in this context is used because its included in an f-string surrounded with double quotes ""

Elsie4ever (approved):

PR #624, opened 2025-07-21
feat: update `az iot ops schema version` to allow for `--version`

vilit1:
There was a suggestion to change --name to --version. Adding in the option so a user can choose their poison.

image

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/params.py
```diff
@@ -1669,6 +1669,30 @@ def load_iotops_arguments(self, _):
             help="Storage account container name where schemas will be stored.",
         )
 
+    with self.argument_context("iot ops schema version") as context:
+        context.argument(
+            "version_name",
+            options_list=["--version", "--ver", "--name", "-n"],
```
c-ryan-k, line 1675: If we switch to `--version` do we need the `--name` and `-n`? Or is this one of those situations where the version IS the name - just want to make sure we don't block usage of `-n` in the future if we need it

c-ryan-k (approved):
approved with a suggestion / question - feel free to resolve and merge

azext_edge/edge/params.py
digimaun: version is the name
image

digimaun: `--name`/`-n` could be removed to reduce potential confusion. The PR stems from user confusion with `--name`. What's your take @c-ryan-k . 

Edit: If --name/-n is removed, I think the description for version name should just say "Schema version."

c-ryan-k: I'm very conflicted because `-n` has precedent as `identifier` of an object that's the context of the command ( `schema show -n` etc.), but in this case it more closely resembles a `--version` datatype, as a value of digits.

I'm slightly in the favor of `--version / -ver` over `--name / -n` I think, based on the backing value and the datatype.

vilit1: or we can just do both for now - I feel like having --version and --name will evoke the idea of version == name

PR #625, opened 2025-07-22
feat: ops create ensures adr namespace exists + de-dupe ops create errors

digimaun:


azext_edge/edge/providers/orchestration/work.py
```diff
@@ -430,12 +431,16 @@ def _do_work(self):
...
             if self._targets.instance_name:
                 self._headers["CommandName"] = "iot ops create"
-                # Ensure schema registry exists.
-                self.resource_client.resources.get_by_id(
-                    resource_id=self._targets.schema_registry_resource_id,
-                    # TODO: Is this preview version still necessary?
-                    api_version=DeviceRegistryMgmtApiVersion.V20240901_preview.value,
-                )
+                # Ensure schema registry and namespace resources exist.
+                for resource_id_pair in [
```
vilit1: teeny nit: you can unpack at this level

I prefer unpacking since the variable names tend to be more descriptive than pair[0], pair[1] but it is a teeny nit cause you aren't doing much in the for loop (so another person can easily figure out pair[0], pair[1])

vilit1 (approved):

vilit1 (approved):

PR #632, opened 2025-07-24
fix: support bundle connector resources

Elsie4ever:
1. Make sure support bundle capture pod and service resource that start with "aio-onvif-", "aio-media-"
2. Test updates for connectors and arc container storage edge case when both `azure-arc-containerstorage` and `azure-arc-cstor` exist
image
image

---

### General Guidelines

Intent for Production

Basic expectations


Elsie4ever:
FYI: IT is failing due to meso test failure

HangyiWang (approved):

azext_edge/tests/edge/support/create_bundle_int/test_auto_int.py
```diff
@@ -69,6 +70,9 @@ def test_create_bundle(cluster_connection, ops_service, bundle_dir, mq_traces, t
     arc_namespace = namespaces.get("arc")
     certmanager_namespace = namespaces.get("certmanager")
 
+    # cleanup paths in walk_result that are not in namespaces (user customized namespaces)
```
vilit1: Does this mean the support bundle is capturing information that it shouldn't be?

azext_edge/tests/edge/support/create_bundle_int/helpers.py
```diff
@@ -316,6 +316,25 @@ def get_all_kinds_from_manager(
...
 
+def cleanup_walk_result(
+    walk_result: Dict[str, Dict[str, List[str]]],
+    exclude_namespaces: List[str],
+) -> Dict[str, Dict[str, List[str]]]:
+    """
+    Cleans up the walk result by removing paths that contains namespaces to be excluded.
+
+    :param walk_result: The original walk result dictionary.
+    :param exclude_namespaces: List of namespaces to exclude from the walk result.
```
vilit1: This function description is confusing - 

so if I have exclude_namespaces = [a,b,c] and my walk_result is [b,c,d] -> does this mean the end result is [b,c] or [d]?

if it is [b,c] then fix this param description or change the exclude_namespaces to "kept_namespaces" or "namespaces_to_keep" or something

azext_edge/tests/edge/support/create_bundle_int/helpers.py
```diff
@@ -316,6 +316,25 @@ def get_all_kinds_from_manager(
     return result - set(exclude_kinds)
 
 
+def cleanup_walk_result(
```
vilit1: also can this functionality be shoved into clean_up_folders? (either called from there or the functionality moved)

azext_edge/tests/edge/support/create_bundle_int/test_auto_int.py
Elsie4ever: this is non related to production code. this is only to reduce the noise from testing, for example if user decide to deploy some connector simulator resource in new namespace, we would not want to check the files under that namespace folder, that will definitely cause test failure because we don't know what to expect.

azext_edge/tests/edge/support/create_bundle_int/helpers.py
Elsie4ever: it's [d] actually

azext_edge/edge/providers/support/connectors.py
```diff
@@ -21,28 +21,33 @@
...
-# TODO: more connectors will be added
 SIMULATOR_PREFIX = "opcplc-"
 OPC_PREFIX = "aio-opc-"
 OPC_APP_LABEL = "app in (aio-opc-supervisor, aio-opc-admission-controller)"
 OPC_NAME_LABEL = NAME_LABEL_FORMAT.format(label="aio-opc-opcua-connector, opcplc")
 OPC_NAME_VAR_LABEL = "name in (aio-opc-asset-discovery)"
 CONNECTORS_DIRECTORY_PATH = "connectors"
+# The prefix for connectors, will use common labels once they are available
+AIO_ONVIF_PREFIX = "aio-onvif"
+ONVIF_PREFIX = "onvif"
```
vilit1: I am still skeptical about using these non-aio generic prefixes - did you get confirmation about those resources being necessary to capture?

azext_edge/tests/edge/support/create_bundle_int/helpers.py
vilit1: Sorry - meant clean_up_folders not process_top_level - I need to add docstrings everywhere now

azext_edge/edge/providers/support/connectors.py
Elsie4ever: this is what originally from the bug (from connector service team)
image

asked follow up questions but no answer so far

azext_edge/tests/edge/support/create_bundle_int/helpers.py
```diff
@@ -316,6 +316,25 @@ def get_all_kinds_from_manager(
...
+    walk_result: Dict[str, Dict[str, List[str]]],
+    exclude_namespaces: List[str],
+) -> Dict[str, Dict[str, List[str]]]:
+    """
+    Cleans up the walk result by removing paths that contains namespaces to be excluded.
+
+    :param walk_result: The original walk result dictionary.
```
digimaun: Moving forward can we please use the docstring style shown here https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/orchestration/resources/brokers.py#L134, vs the legacy sphinx style written here.

azext_edge/tests/edge/support/create_bundle_int/helpers.py
Elsie4ever: ircc clean_up_folders is only used for aio namespace check, while cleanup_walk_result will be use for both within and out of aio namespace resource check

azext_edge/tests/edge/support/create_bundle_int/helpers.py
vilit1: oh let me add that comment into the other pr

azext_edge/tests/edge/support/create_bundle_int/helpers.py
vilit1: so then in test_auto_int.py it is used as:
`walk_result = cleanup_walk_result(walk_result, namespaces)`

this means that we are removing all of the namespaces? like "aio" and what not? What is this testing then?

azext_edge/tests/edge/support/create_bundle_int/helpers.py
vilit1: clean_up_folders is called in process_top_levels.

process_top_levels is called in get_file_map (which is called in every non auto bundle test) and directly in test_auto_int.py for walk_result

the clean_up_folders function seems to be called for every test already...

azext_edge/tests/edge/support/create_bundle_int/test_auto_int.py
vilit1: This circles back to the broad prefixes - we should not be in a situation where we are capturing things we should not be (like ~user data~ user customized namespaces). If we are, we can get into trouble later on.

Honestly, I would just tell the service team we are not going to capture that broad of a prefix/they need labels and leave it at that.

azext_edge/edge/providers/support/connectors.py
vilit1: Considering lack of time - I would just tell the service team we are not going to capture that broad of a prefix/they need labels and leave it at that.

azext_edge/tests/edge/support/create_bundle_int/helpers.py
vilit1: > it's [d] actually

The code is a bit confusing to read (keeping track of double negatives is hard)
I simplified it to focus on the `not in exclude_namespaces`:
image

azext_edge/tests/edge/support/create_bundle_int/test_auto_int.py
Elsie4ever: following up with service team

Elsie4ever:
image

c-ryan-k (approved):

PR #658, opened 2025-08-20
feat: root asset migration part 1

digimaun:
* Placeholder command name. Maybe it persists.


azext_edge/edge/_help.py
```diff
@@ -2670,3 +2670,38 @@ def load_iotops_help():
...
+          replacing the original root assets.
+
+          During the migration, namespace devices will be created in-place of the endpoint profiles
+          referenced by the assets. If multiple assets reference the same endpoint profile, a
+          single namespace device will be referenced by the migrated assets.
+
+          Post migration use the `az iot ops ns asset` and `az iot ops ns device` command groups to
+          manage namespace assets and devices.
+
+          It is highly recommended to take a snapshot of the target instance via `az iot ops clone`
```
vilit1, line 2690: nit: I still think having a --back-up assets + profiles is a good idea even though they could accomplish the same thing with clone. I believe running one command (even with an extra parameter) is less error prone than 2 commands.

feel free to say no since I do know this is a temporary command.

azext_edge/edge/providers/orchestration/migration.py
```diff
@@ -0,0 +1,106 @@
+# coding=utf-8
```
vilit1, line 1: nit: i think this may live happily in the adr folder (since it is adr operations) - assuming this is going to only contain the asset migration

azext_edge/edge/command_map.py
```diff
@@ -39,6 +39,7 @@ def load_iotops_commands(self, _):
         cmd_group.command("delete", "delete")
         cmd_group.command("clone", "clone_instance", is_preview=True)
         cmd_group.command("get-versions", "get_versions", is_experimental=True)
+        cmd_group.command("migrate-assets", "migrate_assets")
```
vilit1: Do you know how many versions we will have this? can we add in experimental or preview to give an easier time of deprecation?

azext_edge/edge/providers/orchestration/migration.py
```diff
@@ -0,0 +1,106 @@
...
+        self,
+        instance_name: str,
+        resource_group_name: str,
+        name_patterns: Optional[list[str]] = None,  # Accept exact names or glob patterns (e.g. 'pump*')
+        confirm_yes: Optional[bool] = None,
+        **kwargs,
+    ):
+        with console.status("Querying resources..."):
+            instance_record = self.instances.show(name=instance_name, resource_group_name=resource_group_name)
+            instance_ns_id = instance_record["properties"].get("adrNamespaceRef", {}).get("resourceId")
```
vilit1, line 51: I got a nifty helper that gets all you need and checks for connectivity here: https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/adr/helpers.py#L43

you will still need the not instance_ns_id check though

azext_edge/edge/providers/orchestration/migration.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if not confirm_yes:
+            console.print(f"The following {len(resource_ids)} asset resource Id(s) will be migrated:")
+            console.print_json(data=resource_ids)
+            console.print("Post migration - unreferenced endpoint profiles can be deleted.")
+        should_bail = not should_continue_prompt(confirm_yes=confirm_yes, context="Migration")
+        if should_bail:
+            return
+
+        if not resource_ids:
```
vilit1: move the no migration work to do before the confirmation

azext_edge/edge/providers/orchestration/migration.py
```diff
@@ -0,0 +1,106 @@
...
+            logger.warning("No migration work to do.")
+            return
+
+        payload = {"resourceIds": resource_ids, "scope": "Resources"}
+        with console.status("Working..."):
+            poller = self.ops.begin_migrate(
```
vilit1, line 103: since the api already handles removing the old assets but not the asset endpoint profiles, there may be demand for a command to remove all the old asset endpoint profiles.

digimaun: agreed, that scope is not part of this PR, and probably not the upcoming milestone

azext_edge/edge/_help.py
digimaun: One advantage with clone is the output is intended to be deployable, since it produces ARM templates (more specifically deployments). With the linked mode that already supports assets + profiles, a user can just deploy those specific templates. Just saving off assets and profiles as-is necessitates user having to do extra work to transform to ARM template format.

Either way that's more scope than what _this_ PR is intended (baseline operation).

azext_edge/edge/command_map.py
digimaun: This would be applicable for as long as root assets are supported. Right now, this branch of CLI the package itself is preview (not available in existing GA CLI branch), once this branch GA's the API backing this also GA's. I don't think experimental is appropriate here, but I can explicitly add preview.

azext_edge/edge/providers/orchestration/migration.py
digimaun: An advantage of the way it is - the user sees an array of applicable resource Ids with the confirm prompt, in this case they see an empty array, so there is a bit more information before the No migration work to do warning.

What do you think?

azext_edge/edge/providers/orchestration/migration.py
digimaun: The reason it's not in adr sub package is because I wanted to put multiple migration operations in the migration module and they operate on the instance as a whole.

azext_edge/edge/_help.py
vilit1: I was thinking of something like saving just arm templates for assets + endpoint profiles (similar to `az group export`). Since the asset can be deployed before the asset endpoint profile, the linking should not matter.

we can table this and I can let you sit on this for longer. I do understand the temporary nature, which is why I am pretty hesitant on how much effort should be put in.

azext_edge/edge/providers/orchestration/migration.py
digimaun: Projection is not going to work when the cluster is not connected, that is why it's a hard ValidationError in the migration flow.

Something else is I still want a reference to ConnectedCluster for get_cl_resources_by_type capability.

azext_edge/edge/providers/orchestration/migration.py
vilit1: I still think it is better to have the no migration work first. You can change the message to include "no assets to migrate found" or "no matched assets found" so it is clear it is from the lack of assets

in the current way, you are given a prompt where you know whatever you do, nothing will happen. You are just prompted to do an extra y/n before you can rerun the command with a better pattern match or continue with the rest of your day

azext_edge/edge/providers/orchestration/migration.py
vilit1: if you are fine with waiting for customer demand, then I am also fine

azext_edge/edge/_help.py
digimaun: We want users to be familiar with and use clone, and the use case of backing up resources is exactly what it was intended for. What I meant by linking, is in the context of ARM templates, in this case clone with --mode linked will produce separate independently deployable ARM templates for asset and endpoint profile resources, and it handles scale better than `az group export`.

vilit1 (approved):

PR #664, opened 2025-08-27
feat: migrate assets permission handling.

digimaun:
* +refactor: rewrite asset migrate tests.

### DIFF
--- a/azext_edge/edge/commands_edge.py
+++ b/azext_edge/edge/commands_edge.py
@@ -470,14 +470,20 @@ def migrate_assets(
     resource_group_name: str,
     name_patterns: Optional[list[str]] = None,
     confirm_yes: Optional[bool] = None,
+    adr_sp_oid: Optional[str] = None,
+    skip_role_assignments: Optional[bool] = None,
     **kwargs,
 ) -> dict:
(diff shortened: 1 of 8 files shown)


azext_edge/edge/commands_edge.py
```diff
@@ -470,14 +470,20 @@ def migrate_assets(
     resource_group_name: str,
     name_patterns: Optional[list[str]] = None,
```
Elsie4ever, line 471: saw the related flag `--name-pattern` changed to `--pattern`, should we update this variable name as well?

Elsie4ever (approved):

azext_edge/edge/commands_edge.py
digimaun: An advantage of the cli param is it comes with help that describes it's a pattern for asset names. Cut the "name" part there because I didn't want to conflict with instance name. The handler side here kept the "name_ " to be a bit more descriptive of what the pattern is for (combined with it being used in assert related method & class).

What do you think?

Btw the handler side being plural makes sense since --pattern supports multiple usage which get aggregated.

digimaun: Talked offline.

PR #674, opened 2025-09-09
feat: add create and update for Open telemetry dataflow endpoint

Elsie4ever:
---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/orchestration/common.py
```diff
@@ -265,6 +266,11 @@ class MqttRetainType(Enum):
         DataflowEndpointAuthenticationType.SASL.value,
         DataflowEndpointAuthenticationType.ANONYMOUS.value,
     },
+    DataflowEndpointType.OPENTELEMETRY.value: {
+        DataflowEndpointAuthenticationType.SERVICEACCESSTOKEN.value,
+        DataflowEndpointAuthenticationType.X509.value,
+        DataflowEndpointAuthenticationType.ANONYMOUS.value,
```
vilit1, line 272: anonymous is an option even though we do not expose it to users?

azext_edge/tests/edge/orchestration/resources/dataflow_endpoint/create_update/test_dataflow_endpoint_create_update_otel.py
```diff
@@ -0,0 +1,344 @@
...
+        # unsupported authentication type
+        (
+            {
+                "hostname": "https://otel-collector.monitoring.svc.cluster.local",
+                "port": 4317,
+                "latency": 1,
+                "message_count": 1,
+                "audience": "audience",
+                "tls_disabled": True,
+                "authentication_type": "UnsupportedType",
```
vilit1, line 190: nit: this should be caught by knack already so this case is unnecessary

you can leave it in if you want, an extra unit test doesn't harm anyone

azext_edge/edge/params.py
```diff
@@ -869,6 +869,42 @@ def load_iotops_arguments(self, _):
...
+            context.argument(
+                "port",
+                options_list=["--port"],
+                help="The port number of the open telemetry setting.",
+                type=int,
+            )
+            context.argument(
+                "secret_name",
+                options_list=["--secret-name", "-s"],
+                help="The name for the kubernetes secret that contains the X509 client certificate, private key "
```
digimaun: Don't we have multiple copies of this param defined? Do we need them all - can they be defined at a higher level if they are all the same?

digimaun: Makes me think if there are any other duplicate param removal opportunities for dataflow

azext_edge/edge/providers/orchestration/common.py
Elsie4ever: internally we all see it as one of the DataflowEndpointAuthenticationType, externally it's specified by user using specific flag --no-auth

azext_edge/edge/params.py
Elsie4ever: The reason why I had it not defined on the top level is: even tho flag are the same, but some of them doesn't have same help text, so defined directly under the command help to overwrite help text easily, is there a way to define the flag at the top level but help text can be overwritten on the command level? thanks for any insight in advanced!

digimaun: Alright, I pointed out this particular param because its verbatim defined the same multiple times...what is the difference between:
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/89a0a4aa223c3f3c07f8b67fe9ac49244df73c1e/azext_edge/edge/params.py#L841
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/89a0a4aa223c3f3c07f8b67fe9ac49244df73c1e/azext_edge/edge/params.py#L891
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/89a0a4aa223c3f3c07f8b67fe9ac49244df73c1e/azext_edge/edge/params.py#L766
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/89a0a4aa223c3f3c07f8b67fe9ac49244df73c1e/azext_edge/edge/params.py#L804

>  is there a way to define the flag at the top level but help text can be overwritten on the command level

There is, we can talk about it when you have time.

Elsie4ever: Yeah for the first line it's the same, difference for example:
image
image
image

Elsie4ever: do you mean for the example you made, the ones shared exactly same help text can be defined at top level, and the difference ones can remain under command level? I will make some updates, thanks

digimaun: Yeah, something like that. Your examples show usage of --secret-name for different purposes, the links I shared are multiple definitions of --secret-name with the same purpose and description.

edit:

> can be defined at top level

Where top level to me would be at `iot ops dataflow endpoint` level, unless the same param is needed at higher scope.

digimaun: `CLIArgumentType`'s can also be defined as necessary.

Elsie4ever: updated~

azext_edge/edge/providers/orchestration/resources/dataflows.py
```diff
@@ -654,7 +655,11 @@ def _process_authentication_type(
         })
 
         if authentication_method == DataflowEndpointAuthenticationType.ANONYMOUS.value:
-            return
+            if endpoint_type == DataflowEndpointType.OPENTELEMETRY.value:
+                # Only otel enpoint has anymous settings property
```
digimaun: nit spelling

digimaun (approved):

PR #678, opened 2025-09-12
feat: workaround for py 3.13 cert issue with arc proxy

digimaun:
This is necessary because connectedk8s proxy does not include an issuer for their cert which conflicts with [rfc5280 requirements](https://datatracker.ietf.org/doc/html/rfc5280.html).

### DIFF
--- a/azext_edge/edge/providers/base.py
+++ b/azext_edge/edge/providers/base.py
@@ -35,14 +35,81 @@ def load_config_context(context_name: Optional[str] = None):
     Load default config using a specific context or 'current-context' if not specified.
     """
     from ..util import set_log_level
+    import sys
 
     # This will ensure --debug works with http(s) k8s interactions
     set_log_level("urllib3.connectionpool")
 
     config.load_kube_config(context=context_name)
+
+    # TODO: Remove once connectedk8s proxy certificates are fixed
+    # Python 3.13+ enforces VERIFY_X509_STRICT which rejects certificates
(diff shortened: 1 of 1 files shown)


azext_edge/edge/providers/base.py
```diff
@@ -35,16 +35,83 @@ def load_config_context(context_name: Optional[str] = None):
...
+                        ssl_context.load_verify_locations(cafile=configuration.ssl_ca_cert)
+
+                    # Load client cert if configured
+                    if hasattr(configuration, "cert_file") and configuration.cert_file:
+                        keyfile = getattr(configuration, "key_file", None)
+                        ssl_context.load_cert_chain(configuration.cert_file, keyfile)
+
+                    self.pool_manager.connection_pool_kw["ssl_context"] = ssl_context
+            except Exception:
+                pass  # Silently continue if patching fails
```
vilit1, line 107: don't we want to warn the user in the case they are using py 3.13 + az connected k8s proxy? (this patch already only runs in py3.13 so would recommend warning here or in the other except)

digimaun: What is the user action? The connectdk8s proxy cert needs to conforms to rfc 5280.

vilit1: use a different way to connect?

vilit1 (approved):

azext_edge/edge/providers/base.py
digimaun: Will follow up on this

PR #689, opened 2025-09-26
feat!: wave of deploy updates for 2510.

digimaun:
Highlights:
- Dev deployment.
- Replaces platform extension with certmanager (cm).
- Config plumbing for cm in upgrade (todo: init options).
- Some upgrade validation changes and test-case re-write. More to come on the upgrade side.
- Update connected cluster ARG queries, remove deprecated segments.
- Param simplification.
- Breaking changes implied from platform to cm replacement.


c-ryan-k (approved):
nit - I think title should have a breaking change indicator `!` due to param updates?

digimaun:
> nit - I think title should have a breaking change indicator `!` due to param updates?

Done.

PR #700, opened 2025-10-07
fix: update schema delete operations to accept 200 status codes per GA API spec

ketkimnaik:
Issue: 
Schema delete operations were failing with "Operation returned an invalid status 'OK'" error when using the GA API v2025-10-01.

Root Cause: 
The Azure SDK client was expecting asynchronous response codes [202, 204] but the GA API specification returns synchronous codes [200, 204] for successful delete operations.

Solution:
- Updated SchemasOperations._delete_initial() to accept [200, 204] status codes
- Updated SchemaVersionsOperations._delete_initial() to accept [200, 204] status codes
- Updated corresponding unit tests to validate correct status codes
- Fixed application code to properly handle LRO responses using poller.result() pattern

Impact: 
Both az iot ops schema delete and az iot ops schema version remove commands now work correctly with the GA API.

Reference:
https://github.com/Azure/azure-rest-api-specs/commit/da641063f45bf0784b0e7be1680d7398130a623b#diff-8a80aa7ace699df85d9ad53b603ea8c9c113e3210e45b9a03b0f97001cc1e66a

### DIFF
--- a/azext_edge/edge/providers/orchestration/resources/schema_registries.py
+++ b/azext_edge/edge/providers/orchestration/resources/schema_registries.py
@@ -257,9 +257,10 @@ class Schemas(Queryable):
             return
 
         with console.status("Working..."):
-            return self.ops.begin_delete(
+            poller = self.ops.begin_delete(
                 resource_group_name=resource_group_name, schema_registry_name=schema_registry_name, schema_name=name
             )
+            return poller.result()
 
     def add_version(
         self,
@@ -333,12 +334,13 @@ class Schemas(Queryable):
         resource_group_name: str,
     ):
         with console.status("Working..."):
-            return self.version_ops.begin_delete(
+            poller = self.version_ops.begin_delete(
                 resource_group_name=resource_group_name,
                 schema_registry_name=schema_registry_name,
                 schema_name=schema_name,
                 schema_version_name=name,
             )
+            return poller.result()
 
     def list_dataflow_friendly_versions(
         self,
--- a/azext_edge/edge/providers/orchestration/resources/schema_registries.py
+++ b/azext_edge/edge/providers/orchestration/resources/schema_registries.py
@@ -252,6 +252,7 @@ class Schemas(Queryable):
         schema_registry_name: str,
         resource_group_name: str,
         confirm_yes: Optional[bool] = None,
+        **kwargs
     ):
         if not should_continue_prompt(confirm_yes=confirm_yes):
             return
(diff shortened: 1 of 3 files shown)


azext_edge/edge/providers/orchestration/resources/schema_registries.py
```diff
@@ -257,9 +257,10 @@ def delete(
             return
 
         with console.status("Working..."):
-            return self.ops.begin_delete(
+            poller = self.ops.begin_delete(
                 resource_group_name=resource_group_name, schema_registry_name=schema_registry_name, schema_name=name
             )
+            return poller.result()
```
digimaun: We should use `wait_for_terminal_state` to process the poller result. Look for existing examples.

azext_edge/edge/vendor/clients/deviceregistrymgmt/v20251001/operations/_operations.py
```diff
@@ -24825,7 +24825,7 @@ def _delete_initial(  # pylint: disable=inconsistent-return-statements
 
         response = pipeline_response.http_response
 
-        if response.status_code not in [202, 204]:
+        if response.status_code not in [200, 204]:
```
digimaun, line 24828: If the open api spec is correct, we should regenerate the client rather than manually change. This can be fine for now, but we have to make a note that it is fixed in the regenerated client.

azext_edge/edge/providers/orchestration/resources/schema_registries.py
ketkimnaik: Done. Tested by locally running integration test.

azext_edge/edge/vendor/clients/deviceregistrymgmt/v20251001/operations/_operations.py
ketkimnaik: I checked that this is auto generated file and tried to find the documentation to find the way to regenerate this file but didn't find a way. Hence, I updated it manually for the testing purpose.

digimaun (approved):

azext_edge/edge/vendor/clients/deviceregistrymgmt/v20251001/operations/_operations.py
digimaun: The doc is here https://github.com/Azure/azure-iot-ops-cli-extension/wiki/Autorest-config, also I can do this for you against the `dev` branch.

ketkimnaik: Sure, thank you :) I will complete my PR once all the build are completed successfully.

PR #716, opened 2025-10-15
feat: upgrade handles opcua spc migration

digimaun:


azext_edge/edge/providers/orchestration/migration.py
```diff
@@ -176,3 +179,83 @@ def migrate_to_namespace(
...
+
+        self.spc_opcua: Optional[dict] = None
+        self.spc_default: Optional[dict] = None
+
+    def has_v1_spc(self) -> bool:
+        if not self.secretsync_resources:
+            return False
+
+        for spc in self.secretsync_resources.get(SPC_RESOURCE_TYPE, []):
+            if spc["name"].lower() == "opc-ua-connector":
```
c-ryan-k, line 206: nit - could be a constant - but if it's not referenced elsewhere it's probably fine :)

c-ryan-k (approved):

azext_edge/edge/providers/orchestration/migration.py
digimaun: ack, will handle in next one.

PR #724, opened 2025-10-21
feat: wave of clone updates for 2510

digimaun:
* Handles an arrangement of extension configurations that diverge between v1 and v2 (platform vs cert-manager, acsa required vs optional).
* Adds registry endpoint to resource capture set.


azext_edge/tests/edge/orchestration/resources/registry_endpoint/__init__.py
c-ryan-k, line 1: oops, ty :)

c-ryan-k (approved):

PR #726, opened 2025-10-22
test(adr): fix integration test assertions and coverage gaps

ketkimnaik:
ADR Integration Test Fixes

- Device Endpoint Integration Test Fixes Fixed device endpoint removal assertions in integration tests that were causing test failures and corrected test validation logic for endpoint operations to properly handle API responses.

- Asset Management Test Corrections Resolved integration test issues for asset, asset management, and asset event operations by updating test assertions to align with current ADR API behavior and response formats.

- Feature Branch Synchronization Merged latest changes from upstream feature/adr_2025_ga branch to ensure compatibility and incorporated GA API updates while maintaining test coverage for the 2025 GA release.


azext_edge/tests/edge/adr/test_namespace_devices_int.py
```diff
@@ -330,12 +330,19 @@ def test_namespace_device_lifecycle_operations(require_init, tracked_resources:
...
-    assert endpoint_name_media not in result
-    assert endpoint_name_rest not in result
-    assert endpoint_name_opcua in result
-    assert endpoint_name_custom in result
+    # Filter out None values (removed endpoints) to get only active endpoints
+    active_endpoints = {k: v for k, v in result.items() if v is not None}
+    assert len(active_endpoints) == 2
+    # Removed endpoints should be present as keys but with None values
+    assert endpoint_name_onvif in result
+    assert result[endpoint_name_onvif] is None
```
digimaun, line 338: Why is it when the endpoints are deleted the key still exists in the result dictionary?

ketkimnaik: I tested it with the actual command implementation:
image

and we can see when we remove the endpoints there keys remain and only value becomes null.

digimaun: But is that the desired behavior? If I delete an inbound endpoint, I'm expecting it to be totally removed, not leaving the key. Imagine over time you setup and remove 20, 50, 100 endpoints ... why should a remnant of the deleted key remain?

digimaun: As a data point, can we check the behavior of DOE and the resource post inbound endpoint removal?

ketkimnaik: Sure, let me verify the DOE behavior.

digimaun (approved):

PR #752, opened 2025-11-10
fix: prevent health check warnings for legitimately absent MQ resources

ketkimnaik:
## Issue Description:

The az iot ops check command returns error status when optional MQTT Broker resources (BrokerListener, BrokerAuthentication) are not found in a cluster, causing pipeline validation failures. This occurs even when these resources are legitimately absent in certain deployment configurations.

## Current Behavior:

Health check API returns error status when BrokerListener resources are not found
Health check API returns error status when BrokerAuthentication resources are not found
Pipelines fail validation with repeated warnings: "Service broker contains some non successful statuses"
This affects SSE team's AIO Component Validation pipelines for v1.0.3 and v1.3.0 releases

## Expected Behavior:

Health checks should return skipped status for optional resources that are not present
Only core/required resources (like Broker) should return error when missing
Pipelines should pass validation when optional resources are legitimately absent

## Root Cause:

In 2510 AIO deployments, v1beta1 CRDs were removed and certain optional resources (BrokerListener, BrokerAuthentication) may not exist. The health check logic in mq.py was treating absence of these optional resources as errors instead of skipping them.

## Fix applied:

Modified evaluate_broker_listeners() to return skipped status when no listeners
Modified evaluate_broker_authentications() to return skipped status when no authentications found
Kept evaluate_brokers() returning error for general checks (conservative approach for core component)
Aligned with existing pattern used by evaluate_broker_authorizations() and DeviceRegistry checks

### DIFF
--- a/azext_edge/edge/providers/check/mq.py
+++ b/azext_edge/edge/providers/check/mq.py
@@ -98,7 +98,7 @@ def evaluate_broker_listeners(
         resource_name=resource_name,
     )
     if not all_listeners:
-        status = CheckTaskStatus.skipped.value if resource_name else CheckTaskStatus.error.value
+        status = CheckTaskStatus.skipped.value
         fetch_listeners_error_text = f"Unable to fetch {MqResourceKinds.BROKER_LISTENER.value}s in any namespace."
         check_manager.add_target(target_name=target_listeners)
         check_manager.add_target_eval(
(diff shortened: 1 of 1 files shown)


HangyiWang (approved):
So we’re not throwing errors because a missing resource is considered a valid scenario?

ketkimnaik:
> So we’re not throwing errors because a missing resource is considered a valid scenario?

We do have check and throw error for missing resource. We are avoiding throwing error when we don't have listeners, brokers, etc.

PR #758, opened 2025-11-24
feat: Support cross-resource group ADR namespaces

ketkimnaik:
- Removed client-side resource group validation to enable ADR namespaces to exist in different resource groups than IoT Operations instances
- Fixed clone command to properly generate cross-RG namespace references in ARM templates using full 3-parameter resourceId format

**Testing**
- Completed unit tests
- Successfully ran integration tets locally. Attached the screenshots
- Ran liveendpoints for clone and instance creation commands along with device and asset creation commands.

image

image

image


HangyiWang (approved):
Tests are failing.

azext_edge/edge/providers/orchestration/clone.py
```diff
@@ -1027,7 +1027,9 @@ def _analyze_instance(self):
             adr_namespace_ref = instance_copy["properties"].get("adrNamespaceRef", {})
             if adr_namespace_ref:
                 adr_namespace_ref["resourceId"] = (
-                    "[resourceId('Microsoft.DeviceRegistry/namespaces', "
```
digimaun, line 1030: Please ensure this has been validated with a live test.

azext_edge/tests/edge/adr/test_helpers_unit.py
```diff
@@ -183,11 +183,17 @@ def test_get_namespace_for_instance(
...
+    This test validates cross-RG namespace support by using different resource groups
+    for the instance and namespace, ensuring the function correctly extracts both
+    the namespace name and its resource group from the adrNamespaceRef.resourceId.
+    """
     from azext_edge.edge.providers.adr.helpers import get_namespace_for_instance
 
     instance_name = generate_random_string()
     instance_resource_group = generate_random_string()
-    namespace_resource_group = generate_random_string()
+    namespace_resource_group = generate_random_string()  # Different RG for cross-RG testing
```
digimaun, line 196: Change is just doc string and comment ?

azext_edge/edge/providers/orchestration/clone.py
digimaun: When the namespace exists in a different RG than the scope for deployment, just this change is not going to work.

azext_edge/tests/edge/adr/test_namespace_assets_int.py
```diff
@@ -666,3 +666,66 @@ def assert_asset_properties(result, **expected):
...
+    """
+    from azext_edge.edge.util.id_tools import parse_resource_id
+
+    instance_name = require_init["instanceName"]
+    instance_rg = require_init["resourceGroup"]
+
+    # Get the namespace ID from the instance
+    instance_result = run(f"az iot ops show -n {instance_name} -g {instance_rg}")
+    adr_namespace_ref = instance_result.get("properties", {}).get("adrNamespaceRef")
+    if not adr_namespace_ref or not adr_namespace_ref.get("resourceId"):
```
digimaun: When is the instance getting created without a namespace ref?

azext_edge/edge/providers/orchestration/clone.py
ketkimnaik: When I tested the scenario was like this:
1. My instance was created with namespace in different rg and 
2. Deleted the namespace in different rg 
3. I ran clone command and it created the instance
4. I created the namespace in the different rg and for the cloned instance I created the device which got created successfully.

azext_edge/tests/edge/adr/test_helpers_unit.py
ketkimnaik: yeah:
Added a docstring explaining what the test validates
Added a comment clarifying that the namespace RG is different for cross-RG testing

azext_edge/tests/edge/adr/test_namespace_assets_int.py
ketkimnaik: It is not required because of cross rg I added but removed it now.

azext_edge/edge/providers/orchestration/clone.py
digimaun: Clone captures devices and assets. When the clone definition is deployed those resource should be deployed against a namespace in the same or different rg.

azext_edge/tests/edge/orchestration/test_targets_unit.py
```diff
@@ -756,6 +745,40 @@ def test_custom_location_name_limits(target_scenario: dict, expected_error: Expe
             assert len(targets.custom_location_name) == len(target_scenario["custom_location_name"])
 
 
+def test_cross_rg_namespace_support():
```
digimaun: Instead of new tests, test cases in the existing tests such as test_init_targets can be used. The result being same coverage with less code. Additional assertions as necessary can be added to test_init_targets.

azext_edge/tests/edge/adr/test_namespace_devices_int.py
```diff
@@ -588,3 +588,44 @@ def assert_namespace_device_opcua_props(
         assert result_config["security"]["securityPolicy"] == expected_policy
     if "security_mode" in expected:
         assert result_config["security"]["securityMode"] == expected["security_mode"]
+
+
+def test_device_creation_with_cross_rg_instance(require_init, tracked_resources: List[str]):
```
digimaun: Why is this test needed when test_asset_creation_with_cross_rg_instance already creates a device?

azext_edge/tests/edge/adr/test_namespace_assets_int.py
```diff
@@ -666,3 +666,62 @@ def assert_asset_properties(result, **expected):
         assert result_props["serialNumber"] == expected["serial_number"]
     if "software_revision" in expected:
         assert result_props["softwareRevision"] == expected["software_revision"]
+
+
+def test_asset_creation_with_cross_rg_instance(require_init, tracked_resources: List[str]):
```
digimaun: What is the value add of this test when device and asset operations should be looking up the namespace and its resource group that is associated with the instance?

azext_edge/tests/edge/init/int/test_init_int.py
```diff
@@ -411,3 +411,100 @@ def _process_broker_config_file_arg(create_arg_dict: dict, tracked_files: List[s
...
+    try:
+        run(f"az iot ops delete --cluster {cluster_name} -g {instance_rg} -y")
+        logger.info("Cleaned up existing IoT Operations instance before cross-RG test")
+    except Exception:
+        logger.info("No existing instance to clean up")
+
+    # Create a separate resource group and namespace for cross-RG test
+    namespace_rg = f"{instance_rg}-ns"
+    try:
+        rg_result = run(f"az group create -n {namespace_rg} -l westus --only-show-errors")
```
digimaun: Ideally tests should be compatible with our IT workflow, this group create won't work currently due to perms (also I'm curious what was the reason for --only-show-errors here?). On one hand if it's a static rg name, it's possible to give the principal permissions to the specific rg, but id also to validate an IT is necessary value add over unit tests.

@c-ryan-k do you have any input here? What is the least impact way to support cross-rg testing in the IT workflow?

azext_edge/tests/edge/adr/test_namespace_assets_int.py
digimaun: There is also duplication in the test_init_int added content

azext_edge/tests/edge/init/int/test_init_int.py
c-ryan-k: I think for now we can create a second static RG - piped in through an env var like we do with the default RG, `azext_edge_secondary_rg` or something like that.

Since this is the first int test that requires it, I don't think we need to do anything dynamic or more involved yet.

Some future considerations:
- Should this RG be in a different region than our default for testing, or in the same region? This test's constraints can determine this initially, but we may eventually want same-region / cross-rg tests *and* cross-region / cross-rg tests
- We'll need to ensure the SP also gets permissions to this new RG, and that we extend our cleanup script to purge this one as well.

azext_edge/tests/edge/orchestration/test_targets_unit.py
ketkimnaik: Merged it.

azext_edge/tests/edge/adr/test_namespace_devices_int.py
ketkimnaik: Removed added for extra check but seems not required.

azext_edge/tests/edge/adr/test_namespace_assets_int.py
ketkimnaik: Remove for all device , asset and int tests.

azext_edge/tests/edge/init/int/test_init_int.py
ketkimnaik: This is not required since I remvoed this test.

azext_edge/edge/providers/orchestration/clone.py
ketkimnaik: Performed all the test as discussed:
For same cluster using deployment group
For different target cluster with adrnamspace in different rg
Deleted the cluster and deployed using the json template with newly created cluster and checked if everything is present.

digimaun (approved):

PR #764, opened 2025-12-02
feat(adr): namespace asset connector metadata validation

HangyiWang:
This PR introduces `ConnectorMetadataValidator` to validate Asset configurations against their associated Connector's requirements before submission. It also externalizes hardcoded schemas into standalone JSON files and standardizes validation using the `jsonschema` library.

### Connector Metadata Validation
- Added `ConnectorMetadataValidator` class to fetch OCI artifacts from container registries and validate configurations using `jsonschema`
- Integrated validation into `asset dataset add`, `asset dataset update`, and `asset dataset datapoint add` commands
- Added `jsonschema` dependency to `pyproject.toml`

### Schema Externalization
- Extracted 9 schemas (OPCUA, MQTT, REST, ONVIF, etc.) from `specs.py` into individual JSON files under `schemas/`
- Updated `ensure_schema_structure` in `helpers.py` to use `jsonschema`, replacing custom recursive validation
- Improved error messages for `oneOf` validation failures

### Asset Endpoint Profiles Refactor
- Refactored `_build_opcua_config` to construct and validate the full configuration object against the schema
- Removed manual parameter checks and redundant helper functions

### Testing
- Added unit and integration tests for validator logic and schema fetching

image

image


azext_edge/tests/helpers.py
```diff
@@ -227,6 +227,10 @@ def run(command: str, shell_mode: bool = True, expect_failure: bool = False) ->
         # logger since pytest can cut off long commands
         logger.error(f"Command `{command}` failed.")
         raise CLIInternalError(result.stderr)
+    elif expect_failure and result.returncode != 0:
```
digimaun: This condition looks odd. expect_failure typically indicates if a non-zero result code is expected (which aligns with behavior in the param description). This added condition combined with the previous branch means `result.returncode != 0` always raises `CLIInternalError`.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "opentelemetry-proto~=1.20.0",
     "packaging>=23.2",
     "semver>=3.0.4,<4",
+    "jsonschema>=4.0.0",
```
digimaun: Please verify there is no dep conflict with `azure-iot` extension, this has been an issue in the past.

azext_edge/edge/providers/adr/schemas/opcua_connector_metadata.json
```diff
@@ -0,0 +1,429 @@
...
+                "title": "AIO OPC UA Device Inbound Endpoint Additional Configuration Schema",
+                "description": "Schema for the additional configuration of OPC UA inbound endpoint of an device in Azure Industrial Operations",
+                "type": "object",
+                "properties": {
+                    "applicationName": {
+                        "type": "string",
+                        "default": "OPC UA Broker",
+                        "description": "Name that is used if AIO OPC UA Connector needs to create own (self-signed) certificate"
+                    },
+                    "keepAliveMilliseconds": {
```
digimaun, line 45: There are json schemas for asset types here https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/adr/specs.py , stored in this module which are python dicts over json files. Also there is some basic schema validation happening here sans jsonschema
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/adr/helpers.py#L349
- https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/adr/asset_endpoint_profiles.py#L233

Ideally we should be consistent in storing schemas and validation.

azext_edge/tests/helpers.py
HangyiWang: yeah it appears the original intention was to allow tests to verify why a command failed. let me remove it

pyproject.toml
HangyiWang: Good call - `azure-iot` does need `jsonschema 3.2.0`. let me fix it here.

azext_edge/edge/providers/adr/schemas/opcua_connector_metadata.json
HangyiWang: Thanks, I totally agree. I went ahead and moved all the schemas into JSON files in `schema` folder and standardized the validation using `jsonschema` across the board. It’s much cleaner now

HangyiWang: tests passed also 
image

azext_edge/edge/vendor/clients/deviceregistrymgmt/v20250701_preview/__init__.py
```diff
@@ -0,0 +1,5 @@
+# coding=utf-8
```
digimaun, line 1: v20250701_preview directory should not exist anymore

digimaun: https://github.com/Azure/azure-iot-ops-cli-extension/tree/dev/azext_edge/edge/vendor/clients/deviceregistrymgmt

HangyiWang: cool let me remove it

peolivei2:
I'll take a look in the next couple days.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+
+        Args:
+            cmd: The Azure CLI command context
+            asset: The asset resource dictionary
+            instance_name: The IoT Operations instance name
+
+        Returns:
+            ConnectorMetadataValidator instance
+        """
+        # Extract resource group and instance from asset's extended location or ID
```
peolivei2: The comment says extended location or Id, but the comment is only checking the asset Id (as it should).

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+        # Parse namespace from asset ID path
+        # Asset ID format:
+        # /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.DeviceRegistry/namespaces/{namespace}/assets/{asset}
+        # After parsing:
+        # - namespace: "Microsoft.DeviceRegistry"
+        # - type: "namespaces", name: "{namespace}"
+        # - child_type_1: "assets", child_name_1: "{asset}"
+        namespace_name = None
+        if (
+            asset_id.get("namespace") == "Microsoft.DeviceRegistry"
```
peolivei2: I guess it's more of a general CLI question: are this Id fragment lookups case insensitive?
In ARM, Ids are case insensitive (and case varies depending on the RP providing the data), so this should match both `Microsoft.DeviceRegistry` and `microsoft.deviceregistry`

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+        Returns:
+            Dict containing the connector metadata JSON
+        """
+        # Check cache first
+        cache_key = f"{self.endpoint_type}:{self.endpoint_version or 'none'}"
+        if cache_key in self._METADATA_CACHE:
+            return self._METADATA_CACHE[cache_key]
+
+        # Use local bundled schema for OPC UA
+        if self.endpoint_type.lower() in ["microsoft.opcua", "microsoft.deviceregistry.opcua", "opcua"]:
```
peolivei2: Any reason you are checking all these endpoint type strings?
For OPC, the endpoint type should be "Microsoft.OpcUa" and version empty. Check for that and if it matches, load the file (explicitly throwing if the file fails to load). If the endpoint type or version doesn't match, fall through, so other future versions of opc ua can be implemented through OCI.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+
+                for endpoint_type_info in device_endpoint_types:
+                    et = endpoint_type_info.get("endpointType")
+                    ev = endpoint_type_info.get("version")
+
+                    # Match endpoint type
+                    if not et or et.lower() != self.endpoint_type.lower():
+                        continue
+
+                    # Match version (if both specified, they must match; if either is None, match)
```
peolivei2: If the device specifies an endpoint version, it must match the template.
Only if the incoming endpoint version is empty, you can match with any template of that type.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+                        f"Matched connector template '{template_name}' for endpoint type "
+                        f"'{self.endpoint_type}' version '{self.endpoint_version or ev}'"
+                    )
+                    matched_template = template
+                    break
+
+                if matched_template:
+                    break
+
+            if not matched_template:
```
peolivei2, line 239: In the UI, we don't allow users to create devices or assets of a template that they don't have. Is this validation being used only during create/update? If so, check w/ PM if we want to have the same restriction in the CLI.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+                    f"No connector template found for endpoint type '{self.endpoint_type}' "
+                    f"version '{self.endpoint_version}'. Validation will be skipped."
+                )
+                return {}
+
+            # Step 4: Extract connectorMetadataRef
+            connector_metadata_ref = matched_template.get("properties", {}).get("connectorMetadataRef")
+
+            if not connector_metadata_ref:
+                logger.warning(
```
peolivei2: General comment to change the behavior to throw if we decide that a matching connector must be present. All connectors are required to have a metadata ref.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+
+        Args:
+            image_ref: The OCI image reference (e.g., mcr.microsoft.com/repo:tag)
+
+        Returns:
+            The parsed JSON content of the artifact.
+        """
+        logger.info(f"Fetching OCI artifact: {image_ref}")
+        # 1. Parse the reference
+        # Format: registry/repo:tag
```
peolivei2: This is Python. Why not a regex? 🥲

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+        if not target_digest and len(layers) > 0:
+            target_digest = layers[0]["digest"]
+            logger.warning(f"Using first layer as fallback: {target_digest}")
+
+        if not target_digest:
+            raise ValidationError(
+                f"Could not find connector-metadata.json layer in {image_ref}. "
+                f"Manifest has {len(layers)} layers but none match expected structure."
+            )
+
```
peolivei2, line 319: We can't rely on file name and there's a specific annotation we need to check for connectors. Ping me separately for the details.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+        headers = {
+            "Accept": "application/vnd.oci.image.manifest.v1+json, application/vnd.docker.distribution.manifest.v2+json"
+        }
+
+        # 2. Get Manifest
+        manifest_url = f"{base_url}/manifests/{tag}"
+
+        # Note: For private registries, we'd need authentication (Bearer token).
+        # MCR public images usually allow anonymous pull but might require a token handshake.
+        # For simplicity in this draft, we'll assume public access or implement a basic token flow.
```
peolivei2: So this implementation does not support ACR with env creds?
If this is the case, then we need to keep the validation fully optional (follow up from my comment above on whether we wanted to make it mandatory).

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+                # Create tar file object from bytes
+                tar_bytes = io.BytesIO(blob_response.content)
+                with tarfile.open(fileobj=tar_bytes, mode="r:*") as tar:
+                    # List all files in tar
+                    member_names = tar.getnames()
+
+                    # Find connector-metadata.json file (may be in root or subdirectory)
+                    metadata_file = None
+                    for member in member_names:
+                        if member.endswith("connector-metadata.json"):
```
peolivei2: The layer should contain a single JSON file. We can't rely on the name, as we don't specify it for 3p connectors.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+
+        endpoint_type = endpoint.get("endpointType")
+        endpoint_version = endpoint.get("version")  # May be None
+
+        if not endpoint_type:
+            raise ValidationError(f"Endpoint '{endpoint_name}' does not have endpointType specified.")
+
+        return cls(
+            cmd=cmd,
+            resource_group_name=resource_group_name,
```
peolivei2: The asset resource group doesn't necessarily match the AIO instance resource group. The ADR namespace passed to the AIO instance might be in a different resource group.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,636 @@
...
+    def _get_schema(self, schema_key: str) -> Optional[Dict[str, Any]]:
+        """
+        Extracts the specific schema from the metadata based on the endpoint type and version.
+        """
+        inbound_endpoints = self.metadata.get("inboundEndpoints", [])
+
+        for endpoint in inbound_endpoints:
+            endpoint_type = endpoint.get("endpointType")
+
+            if endpoint_type and endpoint_type.lower() == self.endpoint_type.lower():
```
peolivei2: You need to use the same logic that was used for template matching, including the endpoint version.

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -346,70 +347,42 @@ def process_authentication(
...
+            else:
+                msg = error.message
+
+            if path:
+                final_messages.append(f"Property '{path}' is invalid: {msg}")
+            else:
+                final_messages.append(f"Invalid configuration: {msg}")
+
+        error_msg = "\n".join(final_messages)
+        prefix = f"The following {name} connector arguments are invalid:\n" if name else "Invalid input data:\n"
```
peolivei2: Why connector argument? This looks like a generic validation function.

peolivei2 (changes requested):

azext_edge/edge/providers/adr/validator.py
HangyiWang: :) comment stales. will update

azext_edge/edge/providers/adr/validator.py
HangyiWang: will match both. `parse_resource_id` accepts both but it returns the captured case not any specific case.

azext_edge/edge/providers/adr/validator.py
HangyiWang: I was just trying to match all possibilities. okay I will update following the guidance here.

azext_edge/edge/providers/adr/validator.py
HangyiWang: gotcha

azext_edge/edge/providers/adr/validator.py
HangyiWang: The validation only runs during create/update of sub-resources like datapoint/datasets, not devices or assets

azext_edge/edge/providers/adr/validator.py
HangyiWang: ok will throw

azext_edge/edge/providers/adr/validator.py
HangyiWang: regex is too ugly 😆 , for a format like `registry/repo:tag`, string split is sufficient for now. Here's one copilot generated: 
```python
_OCI_REF_RE = re.compile(
    r"""
    ^(?P<registry>[^/]+)                 # registry host (no slash)
    /(?P<repository>[^:@]+(?:/[^:@]+)*)  # repository path: segments without ':' or '@'
    (?:                                  # optional tag
        :(?P<tag>[^@]+)                  # tag is anything up to an '@' (if present)
    )?$
    """,
    re.VERBOSE,
)
```

azext_edge/edge/providers/adr/validator.py
HangyiWang: ok

azext_edge/edge/providers/adr/helpers.py
HangyiWang: emm seems like it was in the connector validation file and i moved this out to helper. fixed

azext_edge/edge/providers/adr/validator.py
HangyiWang: ok

azext_edge/edge/providers/adr/validator.py
HangyiWang: fixed by fetching JSON from first layer

azext_edge/edge/providers/adr/validator.py
HangyiWang: implemented, testing now

azext_edge/edge/providers/adr/validator.py
HangyiWang: Now I'm mirroring with DOE in my PR, let's check with PM to see any adjustments later

azext_edge/edge/providers/adr/validator.py
HangyiWang: Now retrieve JSON from first layer

azext_edge/edge/providers/adr/validator.py
HangyiWang: Checked with PM, we're doing same thing in CLI.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "opentelemetry-proto~=1.20.0",
     "packaging>=23.2",
     "semver>=3.0.4,<4",
+    "jsonschema>=3.2.0",
```
digimaun: I think this is too broad of a range for jsonschema. Assuming the package abides by semver,  `jsonschema>=4.25,<5` constrains our version to the latest 4 major version (the 25 minor version, I heuristically selected looking at a newer build that support python 3.9). 3.x is from 2019 we don't need to support it. It would be good to double-check there is no conflict in jsonschema having azure-iot and azure-iot-ops installed at the same time.

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -346,70 +347,42 @@ def process_authentication(
...
+                # if there are other errors available.
+                context_errors = error.context
+                # Filter for errors that indicate a value mismatch rather than a structure mismatch
+                value_errors = [
+                    e for e in context_errors
+                    if e.validator not in ["additionalProperties", "required", "type", "const", "enum"]
+                ]
+
+                if value_errors:
+                    best = jsonschema.exceptions.best_match(value_errors)
```
digimaun: nit, could do `best = jsonschema.exceptions.best_match(value_errors or context_errors)` to avoid nesting

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -346,70 +347,42 @@ def process_authentication(
...
+                msg = error.message
+
+            if path:
+                final_messages.append(f"Property '{path}' is invalid: {msg}")
+            else:
+                final_messages.append(f"Invalid configuration: {msg}")
+
+        error_msg = "\n".join(final_messages)
+        prefix = f"The following {name} values are invalid:\n" if name else "Invalid input data:\n"
+        raise InvalidArgumentValueError(f"{prefix}{error_msg}")
```
digimaun, line 385: Mind sharing some example error msgs after all this formatting?

azext_edge/edge/providers/adr/specs.py
```diff
@@ -187,376 +55,29 @@ class SecurityMode(Enum):
...
-            "type": "integer",
-            "minimum" : 0,
-            "default": 0
-        }
-    }
-}
+NAMESPACE_ASSET_OPCUA_DATASET_CONFIGURATION_SCHEMA_V1 = _load_schema("opcua_dataset_config_v1.json")
 
 
 # not used yet
```
digimaun: Are these "not used yet" comments still applicable?

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -419,16 +419,31 @@ def add_dataset(
...
+            "dataSource": data_source,
+            "datasetConfiguration": processed_configs.get("datasetsConfiguration"),
+            "destinations": processed_configs.get("datasetsDestinations", []),
+            "dataPoints": [],  # TODO: future pr, add datapoints
+            "typeRef": type_ref
+        }
+
+        # Validate the dataset configuration against connector metadata
+        try:
+            from .validator import ConnectorMetadataValidator
```
digimaun: These imports are needed and across multiple methods, we should import them at the top of the module.

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -419,16 +419,31 @@ def add_dataset(
...
+
+            validator = ConnectorMetadataValidator.from_asset(
+                self.cmd, asset, instance_name, instance_resource_group
+            )
+            validator.validate_dataset(new_dataset)
+            logger.info(f"Dataset '{dataset_name}' configuration validated successfully.")
+        except ValidationError:
+            raise
+        except Exception as e:
+            logger.warning(f"Dataset validation skipped due to error: {e}")
```
digimaun: As these are warnings going to the user, what would be the typical user actions to resolve?

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,867 @@
...
+                config["destination"] = default_dest
+                return
+            if supported:
+                if len(supported) == 1:
+                    config["destination"] = supported[0]
+                    return
+                if self._DEFAULT_DESTINATION_MQTT in supported:
+                    config["destination"] = self._DEFAULT_DESTINATION_MQTT
+                    return
+                config["destination"] = supported[0]
```
digimaun, line 513: there is a condition `if len(supported) == 1:` for this assignment `config["destination"] = supported[0]` on line 845, but here the assignment happens anyway

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,867 @@
...
+            repository, tag = remainder.split(":", 1)
+        else:
+            repository = remainder
+            tag = "latest"
+
+        return registry, repository, tag
+
+    @classmethod
+    def _get_oci_auth_headers(cls, registry: str, repository: str, cmd=None) -> Dict[str, str]:
+        """Build HTTP headers with authentication for OCI registry requests.
```
digimaun: We should make sure the AIO CLI user agent is plumbed in.

digimaun: When the user appends --debug key request detail should be available, and security details like oauth tokens should be masked.

We typically generate necessary azure clients and have utilities to [instantiate](https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/util/az_client.py#L192) the clients with proper logging and user agent policies.

Also, the clients have useful default behavior like retry configs. If not using generated clients for reasons, then custom should cover the same key capabilities.

pyproject.toml
HangyiWang: yeah but azure-iot-cli-extension uses `jsonschema version ~=3.2.0`. we cannot use `jsonschema>=4.25,<5` though. probably `jsonschema>=3.2.0,<5` is fine.

azext_edge/edge/providers/adr/helpers.py
HangyiWang: sure

azext_edge/edge/providers/adr/helpers.py
HangyiWang: Sure, here're from integration tests: 
image

azext_edge/edge/providers/adr/specs.py
HangyiWang: good catch, they're not - also clean up other comments here

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: yep they're duplicated imports, should be moved

pyproject.toml
digimaun: Extending major version to 3 means we have to support both major versions, where there are likely breaking changes (why else increment the major version?). Is our code compatible with both major versions of jsonschema?

Also we should be up to date with certain loading behaviors of Azure CLI, this was an issue a while ago but depoendency loading behavior may have been patched since then

>  It would be good to double-check there is no conflict in jsonschema having azure-iot and azure-iot-ops installed at the same time.

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: yeah might be multiple reasons like cluster connection, connector not deployed, metadata unavailable, network issue, acr auth issue and etc. let me refine the message to user.

azext_edge/edge/providers/adr/validator.py
HangyiWang: 👻 fixed

azext_edge/edge/providers/adr/validator.py
HangyiWang: sure

HangyiWang: Good call, let me check the existing code patterns

azext_edge/edge/providers/adr/namespace_assets.py
digimaun: Is it accurate for me to say, even if the dataset validation fails, the input may still be valid and still work. Right?

Also to clarify the cluster connection you mentioned, this operation is not intended to directly interaction with the cluster (via arc proxy or otherwise) correct?

azext_edge/edge/providers/adr/validator.py
HangyiWang: add one `oci_client.py`

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: 1. yes.
2. yes, validation shouldn't

pyproject.toml
digimaun: A key aspect of when conflict can occur is if the dep import chain happens on extension load, so if jsonschema gets loaded when extension commands get enumerated by azure cli, that is when problems can happen. But if the dep only gets loaded when a user invokes a command, that should insulate us.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,772 @@
...
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+import hashlib
+import io
+import json
+import jsonschema
+import os
+import tarfile
+from jsonschema import validate
```
digimaun: I think we can simplify imports here, either import jsonschema or the attributes we need from jsonschema....so either `import jsonschema` and reference attributes using jsonschema pkg like `jsonschema.validate`, or `from jsonschema import x, y, z` (but in this path we'd need to rename some attributes to not conflict)

HangyiWang: yeah it makes sense, we can combine them

pyproject.toml
HangyiWang: Use deferred import with 4.25 version.

azext_edge/tests/edge/adr/test_validator_unit.py
```diff
@@ -0,0 +1,1161 @@
...
+            endpoint_type="Microsoft.Onvif",
+            endpoint_version="1.0",
+        )
+
+        config = {"filter": "Topic = 'motion'"}
+        validator.validate_event(config)
+        self.assertIsNone(config.get("destination"))
+
+
+class TestGetEndpointMetadata(unittest.TestCase):
```
digimaun, line 912: In a next iteration, we should consider using the mocked_response pattern for unit tests that call external resources via `requests`, most of our resource mgmt commands use the pattern which allows deeper and more thorough testing of the stack.

azext_edge/edge/providers/adr/validator.py
peolivei2: You're still checking for the name :(
We do not enforce the file name, so just get the first json file, regardless of the name. The file could also be inside folders.

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,774 @@
...
+        if et_lower == self._ENDPOINT_TYPE_OPCUA and version_empty:
+            metadata = self._load_local_opcua_metadata()
+            self._METADATA_CACHE[cache_key] = metadata
+            return metadata
+
+        try:
+            from ...vendor.clients.iotopsmgmt import MicrosoftIoTOperationsManagementService
+
+            iotops_client: MicrosoftIoTOperationsManagementService = get_iotops_mgmt_client(
+                subscription_id=self.cmd.cli_ctx.data.get("subscription_id"),
```
digimaun: Better to use `get_subscription_id()` from  `azure.cli.core.commands.client_factory` in case `cmd.cli_ctx.data.get("subscription_id")` is `None` , reference to the querayble base class which many of our resource mgmt classes inherit from https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/util/queryable.py#L23

azext_edge/edge/providers/adr/validator.py
```diff
@@ -0,0 +1,774 @@
...
+            token = cls._get_auth_token(registry, repository)
+        if token:
+            headers["Authorization"] = f"Bearer {token}"
+
+        return headers
+
+    @classmethod
+    def _fetch_oci_manifest(cls, base_url: str, tag: str, headers: Dict[str, str], image_ref: str) -> Dict[str, Any]:
+        """Fetch and validate the OCI manifest."""
+        manifest_url = f"{base_url}/manifests/{tag}"
```
digimaun: Better direction with the oci client carved out in a module but there is still too much knowledge about the API and its endpoints in this Validator class. Ideally API endpoint specifics are de-coupled from `ConnectorMetadataValidator`, which would use the API tailored client to interact with the API. Compare with our other resource mgmt interactions, API details are encompassed in the respective http client.

azext_edge/edge/util/oci_client.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+    def send_request(
+        self,
+        request: HttpRequest,
+        timeout: int = DEFAULT_TIMEOUT,
+        **kwargs: Any,
+    ) -> HttpResponse:
+        """Send an HTTP request through the pipeline."""
+        return self._pipeline.run(request, connection_timeout=timeout, **kwargs).http_response
```
digimaun: We can do a `deepcopy()` of the passed in request and use the request copy for the pipeline.run so the original request object is not modified.

azext_edge/edge/util/oci_client.py
```diff
@@ -0,0 +1,139 @@
...
+
+class OciRegistryClient:
+    """HTTP client for OCI registries with user-agent, debug logging, and retry."""
+
+    DEFAULT_TIMEOUT = 30
+
+    def __init__(self, **kwargs: Any) -> None:
+        self._pipeline = Pipeline(
+            transport=RequestsTransport(),
+            policies=[
```
digimaun, line 94: There are also authentication policies that handle usage of credential types like AZURE_CLI_CREDENTIAL, but if its not typical usage might have to override with custom logic (looking at validator._get_acr_access_token)

azext_edge/tests/edge/adr/test_validator_unit.py
HangyiWang: Sounds good that's approximately hundreds of lines changes, let's update in the future PR

azext_edge/edge/providers/adr/validator.py
HangyiWang: cool fixed

azext_edge/edge/providers/adr/validator.py
HangyiWang: Good point let me do a refactor, moving unnecessary functions to oci client file.

azext_edge/edge/util/oci_client.py
HangyiWang: sure

azext_edge/edge/util/oci_client.py
HangyiWang: Yeah good call on auth policies. I checked, and the standard `BearerTokenCredentialPolicy` won’t work because ACR uses a two-step, non-standard OAuth2 flow:
1. Get an ARM token, exchange it for an ACR refresh token (/oauth2/exchange)
2. Use the refresh token, exchange it for a scoped ACR access token (/oauth2/token)

The Azure SDK auth policies assume a single credential.get_token() call, so they don’t support this chained exchange.
We could build a custom policy to handle it, but that would add extra complexity.

azext_edge/edge/util/oci_client.py
```diff
@@ -0,0 +1,456 @@
...
+                            f"Successfully obtained anonymous auth token (length: {len(token) if token else 0})"
+                        )
+                        return token
+                    else:
+                        logger.warning(f"Token request failed: {token_resp.status_code} {token_resp.text()}")
+        except HttpResponseError as e:
+            logger.warning(f"Failed to obtain auth token: {e}")
+        return None
+
+    def _get_acr_access_token(self, cmd, registry: str, repository: str) -> Optional[str]:
```
digimaun, line 355: Double checking if the auth tokens are properly masked in logging

peolivei2 (approved):

azext_edge/edge/util/oci_client.py
HangyiWang: Good call. Redacted.

digimaun (approved):

digimaun (approved):

PR #765, opened 2025-12-03
feat: add SSE to endpoint type keyword mapping

ketkimnaik:
Add SSE to endpoint type keyword mapping.


c-ryan-k (approved):
LGTM so long as unit tests pass.

Are there any unit or int tests that need updated for this?

If not, we should consider adding one to ensure this doesn't get missed in the future if we add more connector types :)

PR #609, opened 2025-07-10
feat: min feature flag pattern and application + build version increment

digimaun:

PR #766, opened 2025-12-03
test: Integration test scenario / mark / matrix updates

c-ryan-k:
Refactors our integration tests to support a more flexible, scenario-based test matrix driven by a YAML configuration file.

It introduces custom actions for building the test matrix and also extracts existing steps for setting up cluster resources and trust bundles into new reusable actions to simplify the workflow file.

Also marks `test_namespace_device_lifecycle_operations` as an expected failure for the service error regarding `400216 Invalid Address during endpoint remove`

### Integration test matrix and scenario management

* Added `.github/test-scenarios.yml` to define integration test scenarios
* Introduced `.github/actions/build-int-test-matrix` with a Python script (`build_matrix.py`) to parse the scenario config and user input, outputting a JSON matrix for use in the workflow
* Updated `int_test.yml` workflow to dynamically build and consume the test matrix, replacing the previous hardcoded matrix and custom input logic
* Added `test-scenarios` input to int test workflow to specify which scenarios to run (comma-separated)
* Removed `custom` test scenario in favor of this input; all init/create args are now appended to each cluster run rather than a single custom scenario

### Test marker / tox env updates

* Reworked test marking scenario with the following improvements:
  * Any test without an explicit mark has a mark of `edge` auto-applied to ensure all tests fall under a specific marked scenario
  * All test marks are loaded in `conftest.py` to remove pipeline warnings about marks not existing in `pytest.ini`
  * Test mark logic has been simplified to help ensure tests run in their appropriate scenarios, and that the mark → scenario path is easier to follow
* Upgrade test moved into its own `upgrade` marker and scenario due to possible conflicts with other tests when running in parallel

### Final resulting test scenarios
| Scenario | Description | Tox Environment | Configuration |
|----------|-------------|-----------------|---------------|
| **edge** | Default edge/cluster tests (all tests marked as `edge` or without explicit marks) | `python-edge-int` | Standard deployment |
| **insecure-listener** | Long-running tests with insecure listener deployment | `python-long-int` | `--add-insecure-listener --persist-max-size 10Gi`, disables init preflight checks |
| **rpsaas** | Cloud-side (RPSaaS) tests marked with `@pytest.mark.rpsaas` | `python-rpsaas-int` | Standard deployment |
| **upgrade** | Azure IoT Operations upgrade tests marked with `@pytest.mark.upgrade` | `python-upgrade-int` | Runs serially (non-parallel) to avoid conflicts |
| **redeploy** | Tests cluster redeployment functionality | Runs init tests | Deletes and recreates AIO instance to validate redeployment |
| **trustbundle** | Workload identity federation tests marked with `@pytest.mark.require_wlif_setup` | `python-wlif-int` | Runs serially, sets up custom cluster issuer, `--user-trust` init args with trust settings |

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -1,11 +1,7 @@
 # Integration test scenarios.
-# Fields: name (required), description, tox_env (optional), needs_trust (bool), test_redeploy (bool), init_args, create_args, env (array of single-key maps).
+# Fields: name (required), description, tox_env (optional), needs_trust (bool), test_redeploy (bool), parallel (bool, default: true), init_args, create_args, env (array of single-key maps).
 
 scenarios:
-  - name: custom
-    description: "Custom init/create arguments"
-    tox_env: python-all-int
-
   - name: edge
     description: "Edge tests"
     tox_env: python-edge-int
@@ -30,3 +26,4 @@ scenarios:
     needs_trust: true
     init_args: "--user-trust"
     create_args: "--trust-settings configMapName=example-bundle configMapKey=trust-bundle.pem issuerKind=Issuer issuerName=selfsigned-issuer"
+    parallel: false
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -20,6 +20,7 @@ scenarios:
   - name: upgrade
     description: "Upgrade tests"
     tox_env: python-upgrade-int
+    parallel: false
 
   - name: redeploy
     description: "Redeployment tests"
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -7,7 +7,8 @@ scenarios:
     tox_env: python-edge-int
 
   - name: insecure-listener
-    description: "Insecure listener deployment tests"
+    description: "Insecure listener deploy, long running tests"
+    tox_env: python-long-int
     create_args: "--add-insecure-listener --persist-max-size 10Gi"
     env:
       - AIO_CLI_INIT_PREFLIGHT_DISABLED: "true"
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -17,6 +17,10 @@ scenarios:
     description: "Cloud tests"
     tox_env: python-rpsaas-int
 
+  - name: upgrade
+    description: "Upgrade tests"
+    tox_env: python-upgrade-int
+
   - name: redeploy
     description: "Redeployment tests"
     test_redeploy: true
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -1,5 +1,15 @@
-# Integration test scenarios.
-# Fields: name (required), description, tox_env (optional), needs_trust (bool), test_redeploy (bool), parallel (bool, default: true), init_args, create_args, env (array of single-key maps).
+# Integration test scenario definitions for building test matrix
+
+# Scenario configuration format:
+# - name: Required string identifier for the scenario
+# - description: Optional string description (defaults to name if not provided)
+# - tox_env: Optional string specifying the tox environment to run
+# - needs_trust: Optional boolean for workload identity trust setup (default: false)
+# - init_args: Optional string with extra arguments for 'ops init' command
+# - create_args: Optional string with extra arguments for 'ops create' command
+# - test_redeploy: Optional boolean to test redeployment in the scenario (default: false)
+# - env: Optional list of environment variable dictionaries with key-value pairs
+# - parallel: Optional boolean controlling parallel execution in pytest-xdist (default: true)
 
 scenarios:
   - name: edge
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -17,6 +17,11 @@ scenarios:
     description: "Cloud tests"
     tox_env: python-rpsaas-int
 
+  - name: upgrade
+    description: "Upgrade tests"
+    tox_env: python-upgrade-int
+    parallel: false
+
   - name: redeploy
     description: "Redeployment tests"
     test_redeploy: true
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -1,0 +1,32 @@
+# Integration test scenarios.
+# Fields: name (required), description, tox_env (optional), needs_trust (bool), test_redeploy (bool), init_args, create_args, env (array of single-key maps).
+
+scenarios:
+  - name: custom
+    description: "Custom init/create arguments"
+    tox_env: python-all-int
+
+  - name: edge
+    description: "Edge tests"
+    tox_env: python-edge-int
+
+  - name: insecure-listener
+    description: "Insecure listener deployment tests"
+    create_args: "--add-insecure-listener --persist-max-size 10Gi"
+    env:
+      - AIO_CLI_INIT_PREFLIGHT_DISABLED: "true"
+
+  - name: rpsaas
+    description: "Cloud tests"
+    tox_env: python-rpsaas-int
+
+  - name: redeploy
+    description: "Redeployment tests"
+    test_redeploy: true
+
+  - name: trustbundle
+    description: "Workload identity tests"
+    tox_env: python-wlif-int
+    needs_trust: true
+    init_args: "--user-trust"
+    create_args: "--trust-settings configMapName=example-bundle configMapKey=trust-bundle.pem issuerKind=Issuer issuerName=selfsigned-issuer"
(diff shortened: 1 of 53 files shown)


c-ryan-k:
image

Full run here: https://github.com/c-ryan-k/azure-iot-ops-cli-extension/actions/runs/19941489244

.github/test-scenarios.yml
```diff
@@ -0,0 +1,45 @@
...
+# - test_redeploy: Optional boolean to test redeployment in the scenario (default: false)
+# - env: Optional list of environment variable dictionaries with key-value pairs
+# - parallel: Optional boolean controlling parallel execution in pytest-xdist (default: true)
+
+scenarios:
+  - name: edge
+    description: "Edge tests"
+    tox_env: python-edge-int
+
+  - name: insecure-listener
```
digimaun, line 19: Wonder if this should have a "deployment variations" like name (though user defined trust is also in that umbrella), --persist-max-size enables broker persistence, at least the description should include enabling broker persistence. In the future if we add ITs for broker persistence (post deploy customization), at least --persist-max-size has to be set on deployment. I

.github/test-scenarios.yml
```diff
@@ -0,0 +1,45 @@
...
+    description: "Insecure listener deploy, long running tests"
```
digimaun, line 20: Looking at your run posted in the PR long running tests are not much longer than cloud tests or wlif tests 😄

.github/test-scenarios.yml
```diff
@@ -0,0 +1,45 @@
...
+      - AIO_CLI_INIT_PREFLIGHT_DISABLED: "true"
+
+  - name: rpsaas
+    description: "Cloud tests"
+    tox_env: python-rpsaas-int
+
+  - name: upgrade
+    description: "Upgrade tests"
+    tox_env: python-upgrade-int
+    parallel: false
```
digimaun, line 33: Cool config option

.github/workflows/ci_workflow.yml
```diff
@@ -2,6 +2,11 @@ name: CI Build and Test
 run-name: "CI Build and Test"
 permissions:
   contents: read
+# Cancels previous runs of this workflow on the same PR or ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
digimaun, line 7: This is an interesting add/idea - q is this workflow referenced (or intended to be) multiple times? One thought if it was, is if only one parent workflow (that references this wf) could run at a time.

azext_edge/tests/edge/adr/test_namespace_devices_int.py
```diff
@@ -374,12 +374,18 @@ def test_namespace_device_lifecycle_operations(require_init, tracked_resources:
...
-    )
+    try:
+        result = run(
+            f"az iot ops ns device endpoint inbound remove --device {device_name_2} "
+            f"--instance {instance_name} -g {resource_group} "
+            f"--endpoint {endpoint_name_onvif} {endpoint_name_media} {endpoint_name_rest} "
+            f"{endpoint_name_sse} {endpoint_name_mqtt} -y"
+        )
+    except CLIInternalError as e:
+        # TODO - disable once bug is fixed / test is passing
```
digimaun: maybe in a next iteration we can think of a means to configure pytest.xfail in a common way that is less local to the test impl

.github/test-scenarios.yml
c-ryan-k: They're not, but they are much longer per-test, so based on our struggles with agent parallelization - this does help us reduce overall runtime if we split tests into chunks, regardless of technically why :)

.github/workflows/ci_workflow.yml
c-ryan-k: This file is the parent workflow that calls `ci_build` - so no issues there for now. Not sure how concurrency gets determined in a calling workflow setting, but I believe we could change the grouping to account for whether the workflow itself was called or dispatched to make it work.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency

.github/test-scenarios.yml
c-ryan-k: This one was a junk drawer scenario - we had one matrix feature that deployed the insecure listener / persistence but didn't run any tests, and we had a test scenario that wasn't being run (long-running) so I just decided to toss them together.

It also may not hurt to have a few extra scenarios that just run the "edge" tests to compare consistency - I'll think about the groupings a bit

azext_edge/tests/edge/adr/test_namespace_devices_int.py
c-ryan-k: I honestly hesitated to use xfail at all 🍉 But you bring up a good point, would be a lot safer if we had one defined location for known flaky tests rather than sprinkling stuff like this around the codebase and hoping we remove them all one day 🙃

.github/test-scenarios.yml
c-ryan-k: Thanks :) Something else I considered was that we could also let scenarios define their own `pytest` positional arguments (injected after the base command in `tox.ini`) - so each scenario could have custom parallelization or test runtime config

azext_edge/tests/conftest.py
```diff
@@ -6,14 +6,51 @@
...
+# Marker definitions with metadata
+MARKERS = [
+    MarkerDefinition("edge", "mark tests that run against edge / cluster (auto-applied to unmarked tests)", True),
+    MarkerDefinition("e2e", "mark end-to-end containerized tests (support bundles, checks)", False),
+    MarkerDefinition("rpsaas", "mark tests that are cloud-side", True),
+    MarkerDefinition("upgrade", "mark tests that will run az iot ops upgrade", True),
+    MarkerDefinition("init_scenario_test", "mark tests that will run az iot ops init", True),
+    MarkerDefinition("require_wlif_setup", "mark tests that require workload identity trust setup", True),
+    MarkerDefinition("long_running", "mark tests that take a long time to run", False),
+    MarkerDefinition("no_global_setup", "mark tests that will not use global setup", False),
```
digimaun: What was the use case for no_global_setup & what is using it now?

.github/workflows/int_test.yml
```diff
@@ -439,21 +328,22 @@ jobs:
...
           azext_edge_cluster: ${{ steps.env_out.outputs.CLUSTER_NAME }}
           azext_edge_instance: ${{ steps.env_out.outputs.INSTANCE_NAME }}
-          azext_edge_init_args: ${{ steps.init.outputs.INIT_ARG }}
-          azext_edge_create_args: ${{ steps.init.outputs.CREATE_ARG }}
+          azext_edge_init_args: ${{ matrix.scenario.init_args }} ${{ needs.setup.outputs.runtime_init_args }}
+          azext_edge_create_args: ${{ matrix.scenario.create_args }} ${{ needs.setup.outputs.runtime_create_args }}
           azext_edge_schema_registry_id: ${{ steps.init_reqs.outputs.SCHEMA_REGISTRY_ID }}
           azext_edge_adr_namespace_id: ${{ steps.init_reqs.outputs.ADR_NAMESPACE_ID }}
         run: |
-          tox r -e python-init-int --skip-pkg-install -- --durations=0
```
digimaun, line 456: I think the reason we had --duration=0 in the past was to show the timing of each test

azext_edge/tests/conftest.py
c-ryan-k: I don't know what actually uses it (I don't think anything does) but it was introduced here:

https://github.com/Azure/azure-iot-ops-cli-extension/pull/609/files

If we don't plan to use it, we can remove it, I just figured it may have been important

.github/workflows/int_test.yml
c-ryan-k: Yeah it's actually defined in `tox.ini` already so that it's added no matter what, I think there might be somewhere else in this file I forgot to remove it, but it gets appended as a duplicate anyway:
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/19923347973/job/57116969577#step:18:189

`--durations 0 --junitxml=junit/JUnit.xml --durations=0`

digimaun (approved):
LGTM. A few items to follow up on in the next iteration. Let's talk more on groupings as well.

azext_edge/tests/conftest.py
digimaun: Ok, I backed out of that pattern but forgot to remove the marker, we can remove it.

digimaun (approved):

PR #775, opened 2026-01-08
fix: correct latency unit in dataflow endpoint otel command from milliseconds to seconds

HangyiWang:
For batching latency in dataflow endpoints:

- Kafka endpoints: Use milliseconds
- Other endpoints (DataExplorer, OpenTelemetry and etc.): Use seconds

---

Before the fix: There was a duplicate latency definition in the parent context that said "milliseconds" and was overwriting the "seconds" definition. This affected ALL child commands including otel.

After the fix:
 
- The parent context only has the "seconds" definition
- Kafka endpoints explicitly override with "milliseconds"
- Otel inherits "seconds" from the parent.

### DIFF
--- a/azext_edge/edge/params.py
+++ b/azext_edge/edge/params.py
@@ -460,13 +460,6 @@ def load_iotops_arguments(self, _):
             help="ID of consumer group that the data flow uses to read messages " "from the Kafka topic.",
         )
         context.argument(
-            "latency",
-            options_list=["--latency", "-l"],
-            help="The batching latency in milliseconds. Min value: 0, max value: 65535.",
-            type=int,
-            arg_group="Batching Configuration",
-        )
-        context.argument(
             "partition_strategy",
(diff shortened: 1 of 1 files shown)


azext_edge/edge/params.py
```diff
@@ -634,6 +627,13 @@ def load_iotops_arguments(self, _):
         "iot ops dataflow endpoint update eventhub",
     ]:
         with self.argument_context(cmd_space) as context:
+            context.argument(
+                "latency",
+                options_list=["--latency", "-l"],
+                help="The batching latency in milliseconds. Min value: 0, max value: 65535.",
```
digimaun, line 633: Is the max value still accurate for milliseconds ?

HangyiWang: Yes it is in DOE. Both second and millisecond have same max 65535 limit for latency.

digimaun (approved):

PR #778, opened 2026-01-15
fix: apply default version parameters for device inbound endpoints

ketkimnaik:
Hardcode version for device inbound endpoint command.
Bug: https://msazure.visualstudio.com/One/_workitems/edit/36417862

### DIFF
--- a/azext_edge/edge/providers/adr/namespace_devices.py
+++ b/azext_edge/edge/providers/adr/namespace_devices.py
@@ -50,7 +50,7 @@ class DeviceEndpointType(ListableEnum):
     def get_default_endpoint_version(cls, endpoint_type: str) -> Optional[str]:
         """
         Returns the default version for a given endpoint type.
-        
+
         Returns None if no default version is defined (ONVIF, Media, OPC-UA, custom types).
         """
         version_map = {
@@ -337,13 +337,13 @@ class NamespaceDevices(Queryable):
         **kwargs
     ):
         from .helpers import process_additional_configuration, process_authentication
-        
+
         # Set default version if not provided by user
         if endpoint_version is None:
             default_version = DeviceEndpointType.get_default_endpoint_version(endpoint_type)
             if default_version is not None:
                 endpoint_version = default_version
-        
+
         # get the original inbound endpoints
         device = self.show(
             device_name=device_name,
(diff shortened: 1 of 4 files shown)


azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -320,6 +337,13 @@ def add_inbound_endpoint(
         **kwargs
     ):
         from .helpers import process_additional_configuration, process_authentication
+
+        # Set default version if not provided by user
+        if endpoint_version is None:
+            default_version = DeviceEndpointType.get_default_endpoint_version(endpoint_type)
+            if default_version is not None:
```
digimaun: This condition is not exactly necessary when we expect some versions to be None.

ketkimnaik: I have set them None in the default function. This condition is checking if in the command version is not provided then get the default versions which could be none.

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -46,6 +46,23 @@ class DeviceEndpointType(ListableEnum):
     SSE = "Microsoft.SSEHttp"
     MQTT = "Microsoft.Mqtt"
 
+    @classmethod
+    def get_default_endpoint_version(cls, endpoint_type: str) -> Optional[str]:
```
digimaun: Adding this kind of functionality to a base Enum class is not my favorite, it obfuscates the primary purpose of Enum. We should refactor out of it in the future.

azext_edge/edge/providers/adr/namespace_devices.py
digimaun: Gotcha. Just for fun, what changes if this was the code

```python
        # Set default version if not provided by user
        if endpoint_version is None:
            endpoint_version = DeviceEndpointType.get_default_endpoint_version(endpoint_type)
```

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Removed it.

digimaun (approved):

azext_edge/tests/edge/adr/test_namespace_devices_unit.py
```diff
@@ -2235,10 +2239,12 @@ def test_add_inbound_mqtt_device_endpoint(
         )
```
digimaun, line 2240: Do the other inbound endpoint types need coverage for asserting None when endpoint_version is not provided?

PR #782, opened 2026-01-23
feat: connector metadata validation

HangyiWang:
Merge https://github.com/Azure/azure-iot-ops-cli-extension/pull/764 into dev branch.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -13,3 +13,7 @@ pytest-xdist
 pytest-rerunfailures
 tox
 build
+rich
+kubernetes
+jsonschema
+responses
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -13,7 +13,3 @@ pytest-xdist
 pytest-rerunfailures
 tox
 build
-rich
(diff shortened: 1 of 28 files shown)


dev_requirements.txt
```diff
@@ -13,3 +13,7 @@ pytest-xdist
 pytest-rerunfailures
 tox
 build
+rich
```
digimaun: rich, kubernetes and jsonschema should get installed with the extension, was there an issue with this?

dev_requirements.txt
```diff
@@ -13,3 +13,7 @@ pytest-xdist
...
+kubernetes
+jsonschema
+responses
```
digimaun: responses is already on line 5

dev_requirements.txt
HangyiWang: Every time I set up a virtual env, I have to manually run pip install for them so I guess we should add them here?

HangyiWang: okay removed

digimaun (approved):

PR #786, opened 2026-01-27
fix: gracefully stop loading connector metadata if no ref found in template

HangyiWang:


azext_edge/edge/providers/adr/validator.py
```diff
@@ -245,9 +245,11 @@ def _get_metadata(self) -> Optional[Dict[str, Any]]:
 
             connector_metadata_ref = matched_template.get("properties", {}).get("connectorMetadataRef")
             if not connector_metadata_ref:
-                raise ValidationError(
-                    f"Connector template '{matched_template.get('name')}' is missing connectorMetadataRef."
+                logger.warning(
```
digimaun: Does this need to be a warning, or can we get away with logger.debug() / logger.info() + help content clearly indicating validation is skipped without connectorMetadataRef set?

digimaun (approved):

azext_edge/edge/providers/adr/validator.py
HangyiWang: Good point, warning maybe too noisy. Fix it.

digimaun (approved):

PR #790, opened 2026-01-31
feat: Add connector template CRUD commands for Azure IoT Operations

ketkimnaik:
Connector Template Commands
Changes
New Commands Added:
	- az iot ops connector template create - Create connector templates from registry metadata
	- az iot ops connector template show - Show connector template details
	- az iot ops connector template list - List connector templates for an instance
	- az iot ops connector template update - Update connector template configuration
	- az iot ops connector template delete - Delete connector templates

Key Features:
	- Support for 1P connectors (MCR) and 3P connectors (ACR)
	- Automatic metadata fetching from container registries
	- Bucketized allocation policy with bucket-size validation
	- Secret sync integration with pre-check for enabled state
	- Version upgrade validation (blocks major version and downgrade)
	- Clear signal support ('') forimage-pull-secrets, secrets, storage-volumes, connector-config, trust-settings-secret-ref

Testing
Unit Tests:
	- Added comprehensive unit tests for all CRUD operations
	- Mocked registry metadata fetching for 1P/3P connectors
	- Version upgrade validation tests (major blocked, minor/patch allowed, downgrade blocked)
	- Secret sync state validation tests
        - Clear signal tests for optional fields

Integration tests:
image

Live Endpoint Tests:
image


azext_edge/edge/providers/adr/connector_templates.py
```diff
@@ -0,0 +1,1481 @@
...
+
+        try:
+            # Extract ACR name from hostname
+            acr_name = registry.split('.')[0]
+
+            logger.info("Getting ACR access token for: %s", acr_name)
+
+            # Get Azure AD token for ACR resource
+            logger.debug("Getting Azure AD token for ACR")
+            result = subprocess.run(
```
digimaun: This looks like it won't work on windows, has it been tested in a windows env?

azext_edge/edge/providers/adr/connector_templates.py
```diff
@@ -0,0 +1,1481 @@
...
+        except (ValidationError, InvalidArgumentValueError, CLIInternalError):
+            # Re-raise known errors
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected error fetching metadata: {str(e)}")
+            raise CLIInternalError(
+                f"Failed to fetch connector metadata from {metadata_ref}: {str(e)}"
+            )
+
+    def _fetch_metadata_from_registry(  # noqa: C901
```
digimaun: For container registry API interaction, a prior recent PR introduced `...util.oci_client.OciRegistryClient` which is intended to be a re-usable client. Please use that instead rather than re-implementing. If functionality for the client needs to be extended, it should be done in a generally re-usable way.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -0,0 +1,1481 @@
...
+                # No registry specified (shouldn't happen but handle it)
+                registry = ""
+                image_name = without_tag
+
+            return (registry, image_name)
+        except Exception:
+            # Fallback: return empty registry and full ref as image name
+            return ("", metadata_ref)
+
+    def _is_valid_version_upgrade(
```
digimaun, line 809: You can use `...util.machinery import scoped_semver_import`, then use comparison operators against the parsed semvers, for example

semver.parse(version2) > semver.parse(version1)

look at the upgrade2 module to see how its plumbed.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -0,0 +1,1481 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
```
digimaun, line 6: We should move this module to providers/orchestration/resources/

azext_edge/edge/providers/adr/connector_templates.py
```diff
@@ -0,0 +1,1481 @@
...
+
+        Args:
+            instance_name: Name of the IoT Operations instance
+            resource_group_name: Resource group name
+
+        Raises:
+            ValidationError: If secret sync is not enabled on the instance
+        """
+        try:
+            instance = self.iotops_mgmt_client.instance.get(
```
digimaun: I don't see why we need a direct reference to iot_ops_mgmt_client in this class thus far. We can take a reference to an instance of  orchestration.resources.instances.Instances (which is the higher level AIO Instance client), then use methods on it.

In this case

```python
self.instances.get_default_spc(
    instance_name=self.instance_name,
    resource_group_name=self.resource_group_name,
 )
```

can be used to validate if secret sync has been enabled or not, which will raise a ValidationError if not enabled.

digimaun (commented):
Initial feedback

azext_edge/edge/_help.py
```diff
@@ -393,7 +393,7 @@ def load_iotops_help():
                   {
                       "method": "Custom",
                       "customSettings": {
-                          "endpoint": "https://auth-server-template",
+                          "endpoint": "`https://auth-server-template`",
```
digimaun: This json block is wrapped in ``` , if the backtick ` renders in the example, its no longer a valid example.

HangyiWang:
btw where do we run those 'live endpoint tests'?

azext_edge/edge/_help.py
ketkimnaik: Done. I was fixing all the linting error in that I added these quotes but yes they are not required.

azext_edge/edge/providers/adr/connector_templates.py
ketkimnaik: Yes. I changed it.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Done.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Done.

azext_edge/edge/providers/adr/connector_templates.py
ketkimnaik: Refactored the code to reuse the functions.

azext_edge/edge/providers/adr/connector_templates.py
ketkimnaik: I had to add one more param now it is working in Windows. Next time I will perform testing in Windows as well.

ketkimnaik:
> btw where do we run those 'live endpoint tests'?

I activated the virtual env in the iot ops directory where my changes are present and then run the commands which I have created and check the result.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -0,0 +1,1275 @@
...
+VALID_IMAGE_PULL_POLICIES = ["Always", "IfNotPresent", "Never"]
+VALID_ALLOCATION_POLICIES = ["Bucketized"]
+
+
+class ConnectorTemplates(Queryable):
+    """Provider for connector template operations."""
+
+    def __init__(self, cmd):
+        super().__init__(cmd=cmd)
+        self.iotops_mgmt_client = get_iotops_mgmt_client(
```
digimaun: You can use the iotops_mgmt_client from self.instances so we don't need to initialize multiple of the same client. Look at the pattern in [brokers](https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/azext_edge/edge/providers/orchestration/resources/brokers.py#L39).

azext_edge/edge/util/oci_client.py
```diff
@@ -365,6 +365,15 @@ def _get_acr_access_token(self, cmd, registry: str, repository: str) -> Optional
...
 
         tenant_id = (cmd.cli_ctx.data or {}).get("tenant_id")
+        if not tenant_id:
+            # Fallback to getting tenant from the subscription profile
+            from azext_edge.edge.util.az_client import get_tenant_id
+
+            try:
+                tenant_id = get_tenant_id()
+            except Exception as ex:  # pragma: no cover - profile access failures
+                logger.warning(f"Failed to get tenant ID from profile: {ex}")
```
digimaun: We don't need to show a warning twice for the same thing (there is warning on line 378), the tenant Id fetching can be attempted with two paths here then if both fail show the warning.

However, is this flow going to work without a tenant Id in the exchange payload? CC: @HangyiWang

azext_edge/edge/params.py
```diff
@@ -1661,6 +1661,124 @@ def load_iotops_arguments(self, _):
...
+            nargs="+",
+            action="extend",
+            help="Space-separated connector-specific key-value configurations. "
+            "Format: key=value. Can provide multiple values in one call or use multiple times. "
+            "Examples: --cc brokerAddress=mqtt://broker:1883 qos=1 keepAlive=60 OR "
+            "--cc brokerAddress=mqtt://broker:1883 --cc qos=1 --cc keepAlive=60. "
+            "Use '' to clear existing configurations.",
+        )
+        context.argument(
+            "confirm_yes",
```
digimaun: What is the reason this needs to be redefined?

ketkimnaik: You're right, removed - the global definition already covers this.

azext_edge/edge/util/oci_client.py
ketkimnaik: I merged it and showing one warning in this case.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Following the referred pattern.

azext_edge/edge/util/oci_client.py
HangyiWang: Yes, the ACR OAuth2 exchange flow will work without the tenant parameter in the payload. @digimaun We can run this without the tenant parameter to get access token.

azext_edge/edge/params.py
```diff
@@ -1661,6 +1661,118 @@ def load_iotops_arguments(self, _):
...
+            "                To list available versions for 1st-party connectors:\n"
+            "                `curl https://mcr.microsoft.com/v2/azureiotoperations/"
+            "akri-connectors/TYPE-metadata/tags/list`",
+        )
+        context.argument(
+            "replicas",
+            options_list=["--replicas", "-r"],
+            type=int,
+            help="Number of connector pod replicas to deploy. "
+            "Default is taken from recommendedReplicas in metadata (default: 1).",
```
digimaun: What does it indicate (default: 1). if the default is taken from recommendedReplicas in metadata?

azext_edge/tests/edge/orchestration/resources/conftest.py
```diff
@@ -193,3 +194,16 @@ def append_role_assignment_endpoint(
 
 def echo_callback(request: requests.PreparedRequest) -> tuple[int, dict, Optional[dict]]:
     return (200, {"Content-Type": "application/json"}, request.body)
+
+
+@pytest.fixture()
+def require_init(init_setup):
```
digimaun: We should be able to define this in a higher scope conftest so it doesn't need to be redefined here and in adr.

azext_edge/edge/commands_connector.py
```diff
@@ -186,3 +187,118 @@ def show_connector_opcua_client(
...
+    image_pull_policy: Optional[str] = None,
+    image_pull_secrets: Optional[List[str]] = None,
+    allocation_policy: Optional[str] = None,
+    bucket_size: Optional[int] = None,
+    secrets: Optional[List[str]] = None,
+    storage_volumes: Optional[List[str]] = None,
+    connector_config: Optional[List[str]] = None,
+    trust_settings_secret_ref: Optional[str] = None,
+) -> dict:
+    """Create a new connector template."""
```
digimaun: nit doc strings that repeat the function name have limited value

azext_edge/edge/params.py
```diff
@@ -1661,6 +1661,118 @@ def load_iotops_arguments(self, _):
...
+            "                1st-party connectors (MCR):\n"
+            "                - `mcr.microsoft.com/azureiotoperations/akri-connectors/rest-metadata:VERSION`\n"
+            "                - `mcr.microsoft.com/azureiotoperations/akri-connectors/media-metadata:VERSION`\n"
+            "                - `mcr.microsoft.com/azureiotoperations/akri-connectors/mqtt-metadata:VERSION`\n"
+            "                - `mcr.microsoft.com/azureiotoperations/akri-connectors/sse-metadata:VERSION`\n"
+            "                - `mcr.microsoft.com/azureiotoperations/akri-connectors/onvif-metadata:VERSION`\n\n"
+            "                3rd-party connectors:\n"
+            "                - `REGISTRY.azurecr.io/PATH-metadata:VERSION`\n\n"
+            "                To list available versions for 1st-party connectors:\n"
+            "                `curl https://mcr.microsoft.com/v2/azureiotoperations/"
```
digimaun, line 1693: Seems appropriate to add a CLI command in the future to get this list for users, its better devEx for users not having to leave the context of the AIO CLI.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -0,0 +1,1270 @@
...
+
+            # Handle bucket size for Bucketized policy
+            current_policy = allocation_dict.get("policy", "")
+            if current_policy == "Bucketized" or (allocation_policy is not None and allocation_policy == "Bucketized"):
+                # Use provided bucket size or prompt for it
+                final_bucket_size = bucket_size
+
+                if final_bucket_size is None and allocation_policy is not None:
+                    # Only prompt if we're setting/changing to Bucketized
+                    from knack.prompting import prompt
```
digimaun: Is there enough value add to prompt for this? If it is required after processing the connector template, the user gets the  RequiredArgumentMissingError, then they can provide it in the next run. We haven't really prompted for missing inputs in this project and some contexts like running in CI might have issues with the prompt.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -0,0 +1,1270 @@
...
+            return self._parse_metadata_blob(artifact_info.content, metadata_ref)
+
+        except ValidationError:
+            # Re-raise validation errors from OciRegistryClient
+            raise
+        except Exception as e:
+            if isinstance(e, (ValidationError, InvalidArgumentValueError, CLIInternalError)):
+                raise
+            logger.error(f"Unexpected error fetching metadata: {str(e)}")
+            raise CLIInternalError(
```
digimaun, line 693: If the client fails to fetch connector metadata is this blocking the creation call?

digimaun:
As this is pointing to `dev` can we see an IT run from the IT workflow?

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Yes, fetching metadata is required for creation. If it fails, the command raises an error with details about what went wrong. The metadata provides essential fields (connectorType, version, image references, defaults) needed to construct the template.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: I felt prompt approach would help users to cover in same command but if we don't follow prompt approach which could cause issue then better to avoid it. I have remove it and instead providing the error.

azext_edge/edge/params.py
ketkimnaik: Yeah, I will add command for this in next PR :)

azext_edge/edge/commands_connector.py
ketkimnaik: Removed it.

azext_edge/tests/edge/orchestration/resources/conftest.py
ketkimnaik: That's a great idea. Done.

azext_edge/edge/params.py
ketkimnaik: This text was bit confusing. If user don't provide any value then by default it is taken as 1. I have corrected it to be simple to understand.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
digimaun: Since the raised exception has the error `logger.error` is redundant here, but we can still send it to logger.debug

ketkimnaik: Done.

ketkimnaik:
> As this is pointing to `dev` can we see an IT run from the IT workflow?

I have triggered integration test in IT workflow on my branch. Integration test run link: https://github.com/ketkimnaik/azure-iot-ops-cli-extension/actions/runs/21733449503

ketkimnaik:
> As this is pointing to `dev` can we see an IT run from the IT workflow?

Integration tests completed successfully in IT workflow: https://github.com/ketkimnaik/azure-iot-ops-cli-extension/actions/runs/21733449503/job/62693134455

digimaun (approved):

PR #800, opened 2026-02-19
feat(mgmt-actions): add enable command with EG namespace validation and setup

digimaun:


azext_edge/edge/commands_mgmt_actions.py
```diff
@@ -0,0 +1,55 @@
...
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+from typing import Dict, List, Optional
+
+from knack.log import get_logger
+
+from .providers.orchestration.mgmt_actions import MgmtActions
+
+logger = get_logger(__name__)
```
ketkimnaik: Nit: We have defined logger but not using it anywhere. Should we still keep it?

digimaun: Removed, logging happening in the provider layer. It's cookie cutter from some other command modules that have the same issue (like commands_mq, commands_assets etc) won't propagate it here.

ketkimnaik (approved):

PR #822, opened 2026-03-05
feat(ns): add iot ops ns mgmt-endpoint remove command

digimaun:
* Also fixes some --help line wrapping issues.
* Adds --no-progress plumbing for mgmt-actions.
* Increments CLI version.

### DIFF
--- a/azext_edge/edge/providers/adr/_help.py
+++ b/azext_edge/edge/providers/adr/_help.py
@@ -515,6 +515,45 @@ def load_iotops_adr_help():
     """
 
     helps[
+        "iot ops ns mgmt-endpoint"
+    ] = """
+        type: group
+        short-summary: Manage management endpoints on Device Registry namespaces.
+        long-summary: |
+          Management endpoints are configured by `az iot ops mgmt-actions enable` and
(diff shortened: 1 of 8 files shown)


azext_edge/edge/providers/adr/_help.py
```diff
@@ -514,6 +514,45 @@ def load_iotops_adr_help():
...
+          tearing down the full infrastructure.
+
+          Use `az iot ops ns show` to inspect available endpoint keys under
+          properties.management.endpoints.
+
+        examples:
+        - name: Remove a management endpoint entry by key.
+          text: >
+            az iot ops ns mgmt-endpoint remove -n mynamespace -g myResourceGroup
+            --endpoint-key $CUSTOM_LOCATION_RESOURCE_ID
```
HangyiWang, line 547: feels like a bit weird using `$VAR` here

HangyiWang (approved):

azext_edge/edge/providers/adr/_help.py
digimaun: Chatted offline about consistent convention for resource Id.

PR #830, opened 2026-03-12
feat: add missing namespace asset/endpoint params, onvif event cmds, and opcua start-instance support

ketkimnaik:
• Add missing CLI params to OPCUA/ONVIF/SSE asset and endpoint commands
• Add --filter-clause, --filter-type, --filter-server params for OPCUA event filter feature
• Add event add/list/remove sub-commands for onvif asset type (was missing, other types had it)
• Add --start-inst param for OPCUA dataset add/update
• Switch OPCUA dataset config processing from v1 → v2 schema to support startInstance

Testing:
Added tests and performed integration test locally:

image

image


azext_edge/edge/providers/adr/params.py
```diff
@@ -954,6 +961,34 @@ def load_adr_arguments(self, _):
             arg_type=get_three_state_flag(),
             arg_group="Configuration",
         )
+        context.argument(
+            "sync_properties_into_state_store",
+            options_list=["--sync-properties-into-dss", "--spidss"],
```
digimaun: nit/thought, shorten properties to props or spell it out? like `--sync-props-into-dss`?

digimaun:
It would be super helpful for users to see some example in --help for the new parameters and options!

azext_edge/edge/providers/adr/params.py
ketkimnaik: Renamed it. I used that value because it was mentioned in the Epic :) but props looks better.

ketkimnaik:
> It would be super helpful for users to see some example in --help for the new parameters and options!

I have added examples in help :)

azext_edge/edge/providers/adr/_help.py
```diff
@@ -848,6 +848,10 @@ def load_iotops_adr_help():
         - name: Add an OPC UA endpoint with security settings and asset discovery enabled
           text: >
             az iot ops ns device endpoint inbound add opcua --device mydevice --instance myInstance -g myInstanceResourceGroup --name myOPCUAEndpoint --endpoint-address "opc.tcp://192.168.1.100:4840" --security-policy "Basic256Sha256" --security-mode "SignAndEncrypt" --run-asset-discovery
+
+        - name: Add an OPC UA endpoint with asset discovery and property sync to state store enabled
+          text: >
+            az iot ops ns device endpoint inbound add opcua --device mydevice --instance myInstance -g myInstanceResourceGroup --name myOPCUAEndpoint --endpoint-address "opc.tcp://192.168.1.100:4840" --run-asset-discovery --sync-props-into-dss true
```
digimaun: nit, --sync-props-into-dss implies true, the true isn't actually needed

azext_edge/edge/providers/adr/_help.py
```diff
@@ -2001,6 +2005,64 @@ def load_iotops_adr_help():
...
+          text: >
+            az iot ops ns asset onvif event add --asset myonvifasset --instance myInstance
+            -g myInstanceResourceGroup --event-group motionEvents --name intrusion --data-source "camera.intrusion"
+            --dest topic="factory/onvif/events" retain=Keep qos=Qos1 ttl=3600
+
+        - name: Replace an ONVIF event with same name
+          text: >
+            az iot ops ns asset onvif event add --asset myonvifasset --instance myInstance
+            -g myInstanceResourceGroup --event-group motionEvents --name motion --data-source "camera.motion.updated"
+            --replace
```
digimaun: In this replace example are --dest values set from the prior example lost?

digimaun (approved):

azext_edge/edge/providers/adr/_help.py
ketkimnaik: --replace creates the event from scratch, so any previously set properties (including --dest) are lost unless re-specified — the updated example now demonstrates this.

azext_edge/edge/providers/adr/_help.py
ketkimnaik: Updated this.

digimaun (approved):

PR #836, opened 2026-03-13
feat(adr): add export/import for namespace asset sub-resources

HangyiWang:


azext_edge/edge/providers/adr/_help.py
```diff
@@ -3445,3 +3445,201 @@ def load_iotops_adr_help():
             -g myInstanceResourceGroup --name sensorData
             --dest topic="updated/mqtt/topic" retain=Never qos=Qos0 ttl=1800
     """
+
+    # OPC UA event commands
+    helps[
+        "iot ops ns asset opcua event"
```
digimaun: These changes include help that already exists, why are they being redefined?

azext_edge/tests/edge/conftest.py
```diff
@@ -167,10 +170,10 @@ def settings_with_rg(settings):
 def tracked_resources():
     resources = []
     yield resources
-    for res in resources:
+    for res in reversed(resources):
```
digimaun, line 173: What was the reason for this ?

azext_edge/edge/providers/adr/_help.py
HangyiWang: Not redefined, those are added by Ketki a while ago after I created the original import/export PR. let me fix them

azext_edge/tests/_log.py
digimaun, line 1: What are the reasons this module adds value over idiomatic pytest patterns? Seems to me it may be reinventing native functionality.

azext_edge/tests/_log.py
```diff
@@ -0,0 +1,202 @@
...
+            return False
+
+    def step(self, num: int, description: str) -> "_Step":
+        """Create a numbered step context."""
+        return self._Step(self, num, description)
+
+    # Default timeout per CLI command (seconds). Override via TESTLOG_CMD_TIMEOUT env var.
+    DEFAULT_CMD_TIMEOUT = int(os.environ.get("TESTLOG_CMD_TIMEOUT", 300))
+
+    def run_command(
```
digimaun, line 154: This module is called _log, but there is `run_command()`, and there already is a wrapper around subprocess.run in tests/helpers where ITs make use of, we should try to unify on one unless there is significantly different requirements

azext_edge/tests/edge/conftest.py
HangyiWang: the `tracked_resources` list accumulates resource IDs in creation order. Using `reversed(resources)` ensures cleanup deletes resources in reverse creation order, this avoids failures when a later-created resource depends on an earlier one

azext_edge/tests/_log.py
HangyiWang: just adding step-tracking and colored output for long integration tests is the primary motivation, we can also simplify this but it's good for human to read the test details in temrinal
image

azext_edge/tests/_log.py
HangyiWang: good call, we should try to unify

azext_edge/tests/edge/conftest.py
digimaun: Makes sense in theory, but what are concrete examples of resources that were causing issues before this change?

azext_edge/edge/providers/adr/_help.py
HangyiWang: fixed

azext_edge/tests/_log.py
HangyiWang: fixed

azext_edge/tests/edge/conftest.py
HangyiWang: few examples:
`test_asset_lifecycle_int.py` creates an asset endpoint profile first (idx 0), then one or more assets referencing that endpoint (idx 1+). Deleting the endpoint before its dependent assets fails with an ARM dependency error. `test_namespace_assets_int.py` / `test_namespace_asset_export_import_int.py`  creates a device first (idx 0), then assets (custom, opcua, rest, sse, mqtt) that reference the device (idx 1+). Deleting the device before its assets similarly fails.

azext_edge/tests/_log.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Any, List, Optional
+
+from azure.cli.core.azclierror import CLIInternalError
+
+from .helpers import run as _helpers_run
+from knack.log import get_logger
+
+logger = get_logger(__name__)
+
+_PRETTY = os.environ.get("PRETTY_LOG")
```
digimaun: This could be a function instead like
```python
def _is_pretty() -> bool:
    """Check at call time, so monkeypatch / late env changes work."""
    return bool(os.environ.get("PRETTY_LOG"))
```

So it is composable with pytest fixtures, and supports the env check post module import

HangyiWang: nice updated

azext_edge/edge/commands_namespaces.py
```diff
@@ -1704,93 +1704,691 @@ def remove_namespace_asset_dataset(
     )
 
 
+# DATASET EXPORT/IMPORT
+def export_namespace_custom_asset_dataset(
```
digimaun, line 1708: Experimental idea ... there appears to be notable duplication of functions amongst import/export wrt the asset type.

We could try eliminating the duplication by doing a factory/generator pattern like

```python
def _make_export_func(method_name: str):
    """Factory that generates a typed export command function."""
    def export_func(cmd, asset_name, instance_name, instance_resource_group,
                    extension="json", output_dir=".", replace=False):
        return getattr(NamespaceAssets(cmd), method_name)(
            asset_name=asset_name, instance_name=instance_name,
            instance_resource_group=instance_resource_group,
            extension=extension, output_dir=output_dir, replace=replace,
        )
    return export_func

def _make_import_func(method_name: str):
    """Factory that generates a typed import command function."""
    def import_func(cmd, asset_name, instance_name, instance_resource_group,
                    file_path, replace=False, **kwargs):
        return getattr(NamespaceAssets(cmd), method_name)(
            asset_name=asset_name, instance_name=instance_name,
            instance_resource_group=instance_resource_group,
            file_path=file_path, replace=replace, **kwargs,
        )
    return import_func

# Generate all dataset export/import functions
for _asset_type in ["custom", "opcua", "rest", "sse", "mqtt"]:
    globals()[f"export_namespace_{_asset_type}_asset_dataset"] = _make_export_func("export_datasets")
    globals()[f"import_namespace_{_asset_type}_asset_dataset"] = _make_import_func("import_datasets")
```

Would be interested to see if something like that would work. Think about it.

Doesn't have to be in this PR.

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -717,27 +955,386 @@ def remove_dataset_datapoint(
...
+        imported_datapoints = _process_namespace_sub_points_file_path(
+            file_path=file_path,
+            original_items=original_datapoints,
+            point_key="name",
+            replace=replace,
+            csv_converter=_convert_sub_points_from_csv_namespace
+        )
+
+        # Validate imported datapoints
+        try:
```
digimaun, line 1079: This validator try/except pattern is repeated a few times, an idea is to extract it in a helper method like (doesn't need to be in this PR)

```python
def _validate_imported_items(
    self,
    items: List[dict],
    validate_fn,
    resource_label: str,
    asset: dict,
    instance_name: str,
    instance_resource_group: str,
    post_validate_fn=None,
):
    """Run connector-metadata validation on imported items, with graceful fallback."""
    try:
        validator = ConnectorMetadataValidator.from_asset(
            cmd=self.cmd, asset=asset,
            instance_name=instance_name,
            instance_resource_group=instance_resource_group,
        )
        for item in items:
            validate_fn(validator, item)
            if post_validate_fn:
                post_validate_fn(item)
        logger.info(f"{resource_label} validated successfully.")
    except ValidationError:
        raise
    except Exception as e:
        logger.warning(
            f"{resource_label} validation skipped: {e}. "
            "This may occur if the connector is not deployed or the cluster is not connected. "
            f"The {resource_label.lower()} will be imported but may fail at runtime."
        )
        if post_validate_fn:
            for item in items:
                post_validate_fn(item)
```

Which could then be used like

```python
self._validate_imported_items(
    items=imported_datasets,
    validate_fn=lambda v, d: v.validate_dataset(d),
    resource_label="Datasets",
    asset=asset, instance_name=instance_name,
    instance_resource_group=instance_resource_group,
)
```

azext_edge/edge/commands_namespaces.py
HangyiWang: yep i got this one yesterday but it required some code changes/tests, I may file another PR after this one.

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: sure I will put this with the refactoring above into one separate PR.

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -1265,6 +1858,121 @@ def update_stream(
...
+        except ValidationError:
+            raise
+        except Exception as e:
+            logger.warning(
+                f"Stream validation skipped: {e}. "
+                "This may occur if the connector is not deployed or the cluster is not connected. "
+                "The streams will be imported but may fail at runtime if the configuration is invalid."
+            )
+            # Still auto-assign destinations even if validation is skipped
+            for stream in imported_streams:
```
digimaun: Please review this logic to ensure it is optimal, there is dupe here with the validation loop in the try block, but also if there is a ValidationError the stream destination may be set on a subset of streams. Maybe this block could be pulled out of the Exception block and run after validation so it uniformly works across streams...for example:

```python
try:
    validator = ConnectorMetadataValidator.from_asset(...)
    for stream in imported_streams:
        validator.validate_stream(stream)
    logger.info("Streams validated successfully.")
except ValidationError:
    raise
except Exception as e:
    logger.warning(f"Stream validation skipped: {e}. ...")

# Always auto-assign destinations uniformly
for stream in imported_streams:
    if default_destinations and ("destinations" not in stream or not stream["destinations"]):
        stream["destinations"] = deepcopy(default_destinations)
```

azext_edge/tests/edge/adr/test_namespace_asset_export_import_int.py
```diff
@@ -0,0 +1,1035 @@
...
+import pytest
+from typing import List
+
+from ...generators import generate_random_string
+from ..._log import TestLog
+
+pytestmark = [pytest.mark.rpsaas, pytest.mark.long_running]
+
+# Module-level caches for shared resource reuse
+_shared_device_name = None
```
digimaun, line 65: We should consider a module scoped fixture with shared KPIs to integrate with the pytest lifecycle and avoid unintended mutations 

```python
@pytest.fixture(scope="module")
def shared_device(require_namespace_init, tracked_resources):
    """Single shared device for all tests in this module."""
    instance_name = require_namespace_init["instanceName"]
    resource_group = require_namespace_init["resourceGroup"]
    device_name = f"dev-{generate_random_string(8, force_lower=True)}"
    run(f"az iot ops ns device create --name {device_name} --instance {instance_name} -g {resource_group}")
    tracked_resources.append(device_name)
    yield device_name
```

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: kk let me check

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -2304,3 +3279,39 @@ def _update_asset_props(
...
+
+    if not original_items:
+        original_items = []
+
+    original_points = {point[point_key]: point for point in original_items}
+    file_points_dict = {point[point_key]: point for point in file_points}
+
+    for key in file_points_dict:
+        if key in original_points and not replace:
+            logger.warning(f"{key} is already present in the asset and will be ignored.")
```
digimaun: In this case when replace is False, what if we emitted a single warning with all entries that were skipped at the end of the process, rather than emit them one at a time?

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: You're right, this is a partial destination assignment and duplicates logic. fixed.

azext_edge/tests/edge/adr/test_namespace_asset_export_import_int.py
HangyiWang: good point, checking

HangyiWang: this is a bigger refactor than it looks due to fixture scope mismatch. `require_namespace_init` is function-scoped (as is its dependency `require_init`), and pytest forbids a module-scoped fixture from depending on a function-scoped one, it'll fail at collection with ScopeMismatch. I will file a separate PR for the changes and a full round int test pass.

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: Sounds good - fixed with unit tests.

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -1600,6 +2306,283 @@ def remove_management_group_action(
...
+        except ValidationError:
+            raise
+        except Exception as e:
+            logger.warning(
+                f"Management group validation skipped: {e}. "
+                "This may occur if the connector is not deployed or the cluster is not connected. "
+                "The management groups will be imported but may fail at runtime if the configuration is invalid."
+            )
+            # Still preserve actions even if validation is skipped
+            for mgmt_group in imported_mgmt_groups:
```
digimaun: Similar comment as the import stream fix here, 

Image

azext_edge/tests/_log.py
```diff
@@ -0,0 +1,195 @@
...
+                _log(f"  ⚠ FAILED: {exc_val}", "terra")
+            _log(f"  ⏱ ({elapsed})", "dim")
+            return False
+
+    def step(self, num: int, description: str) -> "_Step":
+        """Create a numbered step context."""
+        return self._Step(self, num, description)
+
+    # Default timeout per CLI command (seconds). Override via TESTLOG_CMD_TIMEOUT env var.
+    DEFAULT_CMD_TIMEOUT = int(os.environ.get("TESTLOG_CMD_TIMEOUT", 300))
```
digimaun: DEFAULT_CMD_TIMEOUT could also benefit as a function, like the refactor towards _is_pretty()

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -1600,6 +2306,283 @@ def remove_management_group_action(
...
+        except ValidationError:
+            raise
+        except Exception as e:
+            logger.warning(
+                f"Action validation skipped: {e}. "
+                "This may occur if the connector is not deployed or the cluster is not connected. "
+                "The actions will be imported but may fail at runtime if the configuration is invalid."
+            )
+            # Still apply defaults even if validation is skipped
+            for action in imported_actions:
```
digimaun: Same here

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -49,6 +51,149 @@
...
+        if sampling_value and str(sampling_value).strip():
+            configuration["samplingInterval"] = int(sampling_value)
+
+        # Move queueSize to configuration if it exists and is not empty
+        queue_value = point.pop("queueSize", None)
+        if queue_value and str(queue_value).strip():
+            configuration["queueSize"] = int(queue_value)
+
+        if configuration:
+            config_key = "dataPointConfiguration" if "dataSource" in point else "eventConfiguration"
```
digimaun, line 142: Is dataSource not relevant for events?

digimaun (approved):

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: `dataSource` is relevant for events — it's just optional, not required. 
For datapoints, dataSource is required. 
For events (and event groups), dataSource is optional,

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: fixed

azext_edge/tests/_log.py
HangyiWang: sure

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: fixed - try/exception + default assignment afterwards without duplicating

azext_edge/edge/_help.py
```diff
@@ -393,7 +393,7 @@ def load_iotops_help():
                   {
                       "method": "Custom",
                       "customSettings": {
-                          "endpoint": "https://auth-server-template",
+                          "endpoint": "`https://auth-server-template`",
```
digimaun: This is meant to be a copy-paste example for users, the tildes breaks that

digimaun (approved):

azext_edge/edge/_help.py
HangyiWang: fixed

azext_edge/tests/edge/conftest.py
```diff
@@ -167,10 +170,10 @@ def settings_with_rg(settings):
...
         try:
-            run(f"az resource delete --id {res} -v")
-        except CLIInternalError:
+            run(f"az resource delete --id {res} -v", timeout=600)
```
digimaun, line 175: Note on timeout to chat about

azext_edge/tests/edge/adr/test_namespace_asset_helpers_unit.py
```diff
@@ -522,8 +523,8 @@ def test_process_configs(mocker, asset_type: str, test_case: dict, default: bool
     # Set up mocks for all the helper functions
     mocks = {}
     for func_name in [
-        "_process_opcua_dataset_configurations_v2",
-        "_process_opcua_event_configurations_v2",
+        "_process_opcua_dataset_configurations_v1",
```
digimaun: @ketkimnaik changed these to v2 in a prior PR https://github.com/Azure/azure-iot-ops-cli-extension/pull/830

azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -1912,10 +2878,10 @@ def _process_configs(
         # allowed: datasets, events, mgmt groups (no schema?), destinations must be mqtt
         # not allowed: streams
         result = {
-            "datasetsConfiguration": _process_opcua_dataset_configurations_v2(
+            "datasetsConfiguration": _process_opcua_dataset_configurations_v1(
```
digimaun: Why are we changing back to _process_opcua_dataset_configurations_v1? @ketkimnaik

azext_edge/tests/edge/conftest.py
HangyiWang: OK

azext_edge/tests/edge/adr/test_namespace_asset_helpers_unit.py
HangyiWang: I reverted it back to v2. I thought v2 is not ready. But it's unrelated with import/export commands, only referenced in unit tests.

azext_edge/edge/providers/adr/namespace_assets.py
HangyiWang: reverted to v2

digimaun (approved):

PR #840, opened 2026-03-16
fix(tools): remove false positives from unlabeled resources script

ketkimnaik:
The output for final summary looks like this:
image

The service team side actions looks like this:

image

---

### General Guidelines

Intent for Production

Basic expectations


HangyiWang:
Nice thanks @ketkimnaik !

HangyiWang (approved):

tools/get_all_unlabeled_resources.sh
```diff
@@ -141,22 +187,27 @@ for keywords_pattern in "${!KEYWORD_LABEL_MAP[@]}"; do
...
+                        UNLABELED_RESOURCES+=("${resource}|${namespace}|${name}|${expected_label}|${label_value}")
                     fi
                 fi
-            done
+            done < <(echo "${all_resources}" | tail -n +2)
         fi
     done
     echo ""
     
     # Check CRDs
```
digimaun: We don't need common labels on CRs from CRDs because we have the knowledge of what CRDs are associated with our service, and any CR from a known CRD is associated with the edge component that owns the CRD. Common labels are primarily for runtime resources (pods, statefulsets, daemonsets, configmaps etc).

ketkimnaik: Done.

digimaun:
We also have support bundle integration tests, what does this script buy us over those integration tests? If net value, should this script run in the IT workflow? Nothing sensitive is captured, correct?

ketkimnaik:
> We also have support bundle integration tests, what does this script buy us over those integration tests? If net value, should this script run in the IT workflow? Nothing sensitive is captured, correct?

We can run it as a non-blocking informational step in the IT workflow that prints results but doesn't fail the pipeline.

Sensitive data:

No — it only reads metadata.labels via jsonpath, never resource content/spec/data. No secrets, no payloads, no PII. Safe to run in CI.

tools/get_all_unlabeled_resources.sh
```diff
@@ -23,6 +24,29 @@ declare -A KEYWORD_LABEL_MAP=(
     ["meso|observability"]="microsoft-iotoperations-observability"
 )
 
+# All valid AIO app.kubernetes.io/name label values.
+# If a resource already has any of these, it is owned by another AIO service
+# and should not be flagged as unlabeled (prevents cross-service name collision false positives).
+VALID_AIO_LABELS=(
+    "microsoft-iotoperations-mqttbroker"
```
digimaun: `app.kubernetes.io/name=microsoft-iotoperations` is a valid label as well

tools/get_all_unlabeled_resources.sh
```diff
@@ -174,76 +225,54 @@ for keywords_pattern in "${!KEYWORD_LABEL_MAP[@]}"; do
...
 fi
 
+echo ""
+echo "=========================================="
+echo "=== SERVICE TEAM ACTION ITEMS ==="
+echo "=== (CRD definitions missing label — not CLI gaps) ==="
+echo "=========================================="
+echo ""
+
+if [[ ${#CRD_DEFINITION_VIOLATIONS[@]} -eq 0 ]]; then
```
digimaun: CRDs are not expected to have the common labels

tools/get_all_unlabeled_resources.sh
digimaun: There are still CRD paths in the code.

tools/get_all_unlabeled_resources.sh
ketkimnaik: Removed it :)

tools/get_all_unlabeled_resources.sh
ketkimnaik: Added in valid labels.

tools/get_all_unlabeled_resources.sh
```diff
@@ -68,33 +92,54 @@ is_label_valid() {
...
+    # If the resource already has any valid AIO label (just a different one), it is
+    # owned by another AIO service — not a labeling gap, just a name collision.
+    for valid_label in "${VALID_AIO_LABELS[@]}"; do
+        if [[ "${label_value}" == "${valid_label}" ]]; then
+            return 0
+        fi
+    done
+
     # Special case 1: 'observability-cluster-metrics' resources can have 'microsoft-iotoperations-observability-cluster-metrics' label
     if echo "${name_lower}" | grep -q "observability-cluster-metrics"; then
```
digimaun: When are these special cases applicable? Doesn't the new block on line 103-107 already cover every case the special cases check for?

tools/get_all_unlabeled_resources.sh
digimaun, line 1: set -e / set -o pipefail  , prevents silently continuing from failure

tools/get_all_unlabeled_resources.sh
```diff
@@ -23,6 +23,30 @@ declare -A KEYWORD_LABEL_MAP=(
     ["meso|observability"]="microsoft-iotoperations-observability"
 )
 
+# All valid AIO app.kubernetes.io/name label values.
+# If a resource already has any of these, it is owned by another AIO service
+# and should not be flagged as unlabeled (prevents cross-service name collision false positives).
+VALID_AIO_LABELS=(
```
digimaun: Observation that this is another collection of labels (other being KEYWORD_LABEL_MAP) which can drift

ketkimnaik: Fixed — VALID_AIO_LABELS is now derived at runtime from KEYWORD_LABEL_MAP values, with only microsoft-iotoperations-observability-cluster-metrics kept as an explicit addition since it's a sub-component variant not in the map.

tools/get_all_unlabeled_resources.sh
ketkimnaik: set -o pipefail added; skipping set -e since kubectl calls intentionally return non-zero for missing resource types and we already guard with || true / $? checks — set -e would require || true noise throughout without adding real safety.

tools/get_all_unlabeled_resources.sh
ketkimnaik: you're right, VALID_AIO_LABELS (lines 103–107 at the time) already covers every case the special cases handled. Removed them.

tools/get_all_unlabeled_resources.sh
```diff
@@ -23,6 +24,26 @@ declare -A KEYWORD_LABEL_MAP=(
     ["meso|observability"]="microsoft-iotoperations-observability"
 )
```
digimaun, line 29: nit, since jq is required, we could check for it up front

```bash
command -v jq >/dev/null 2>&1 || { echo "jq is required but not installed."; exit 1; }
```

ketkimnaik: Aded check for jq and kubectl.

digimaun (approved):

PR #849, opened 2026-03-19
perf: parallel universe (4h → <2h)

HangyiWang:
Give 2+hours back. 
55+% time reduction.

Less than 2 hours now (tested multiple times)
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/23446021916
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/23419453308
image

Currently 4+hours in `dev`
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/23429871961
image

- **Switch to `worksteal` scheduler and increase parallelism from 4 to 6 workers.** Work-stealing dynamically rebalances tests across workers, preventing idle workers when test durations are uneven.
- Promote `shared_device`, `endpoint_cache`, and `require_namespace_init` from module to session scope. Devices and endpoints are created once per test session, eliminating redundant provisioning across modules and significantly reducing total CI wall time.
- Add `asset_factory` fixture that creates and caches assets+endpoints on first use per module.
- Skip granular resource cleanup in CI (cluster is destroyed anyway) 
- Delete monolithic `test_namespace_asset_export_import_int.py` (~1150 lines) and split into focused modules: dataset, event, stream, and mgmt export/import 
-  `wait_for_expected_count` now uses exponential backoff (3s → 20s cap) instead of fixed 15s intervals 
- Replace bare `next(`) calls in `namespace_assets.py` with explicit None fallback + CLIInternalError — prevents cryptic `StopIteration` on missing datasets/streams/mgmt-groups after update 
- Add retry logic to `connect-arc` and replace fixed sleeps in `create-k3s-cluster` with proper API server/node readiness polling to reduce flaky CI failures.

---

JFYI 

File | Tests | Total (s) | Total (m) | Avg/test
-- | -- | -- | -- | --
event_export_import_int.py | 13 | 4,685 | **78.1m** | 360s
dataset_export_import_int.py | 11 | 3,942 | 65.7m | 358s
mgmt_export_import_int.py | 9 | 3,308 | 55.1m | 368s
datasets_int.py | 5 | 2,240 | 37.3m | 448s
events_int.py | 4 | 1,768 | 29.5m | 442s
mgmt_int.py | 3 | 1,276 | 21.3m | 425s
streams_int.py | 2 | 1,128 | 18.8m | 564s
stream_export_import_int.py | 2 | 624 | 10.4m | 312s
TOTAL | 49 | 18,971 | 316.2m | 387s

Test | Duration | File
-- | -- | --
media_stream_lifecycle | **815.8s (13.6m)** | streams_int
opcua_dataset_lifecycle | 571.9s (9.5m) | datasets_int
custom_dataset_lifecycle | 564.9s (9.4m) | datasets_int
sse_event_lifecycle | 530.2s (8.8m) | events_int
custom_mgmt_lifecycle | 509.6s (8.5m) | mgmt_int
custom_event_lifecycle | 505.3s (8.4m) | events_int
opcua_mgmt_lifecycle | 463.3s (7.7m) | mgmt_int
datapoint_export[yaml-opcua] | 434.0s (7.2m) | dataset_export
mgmt_action_export[yaml-opcua] | 425.2s (7.1m) | mgmt_export
datapoint_export[json-custom] | 421.7s (7.0m) | dataset_export
event_export[yaml-opcua] | 420.7s (7.0m) | event_export
datapoint_export[json-opcua] | 419.5s (7.0m) | dataset_export
onvif_event_lifecycle | 417.6s (7.0m) | events_int
event_export[json-sse] | 416.8s (6.9m) | event_export
mgmt_action_export[json-custom] | 416.2s (6.9m) | mgmt_export
datapoint_export[csv-custom] | 413.4s (6.9m) | dataset_export
mgmt_action_export[yaml-onvif] | 410.2s (6.8m) | mgmt_export
event_export[csv-custom] | 409.3s (6.8m) | event_export
mgmt_action_export[json-opcua] | 408.7s (6.8m) | mgmt_export
mgmt_action_export[json-onvif] | 406.8s (6.8m) | mgmt_export
event_export[yaml-sse] | 406.2s (6.8m) | event_export
event_export[json-opcua] | 406.1s (6.8m) | event_export
event_export[json-custom] | 406.1s (6.8m) | event_export
rest_dataset_lifecycle | 374.2s (6.2m) | datasets_int
mqtt_dataset_lifecycle | 369.8s (6.2m) | datasets_int
sse_dataset_lifecycle | 359.6s (6.0m) | datasets_int
dataset_export[custom] | 336.2s (5.6m) | dataset_export
event_group_export[custom] | 335.0s (5.6m) | event_export
datapoint_export[yaml-custom] | 332.7s (5.5m) | dataset_export
event_export[yaml-custom] | 322.1s (5.4m) | event_export
event_group_export[opcua] | 320.1s (5.3m) | event_export
dataset_export[opcua] | 319.8s (5.3m) | dataset_export
dataset_export[sse] | 319.7s (5.3m) | dataset_export
dataset_export[mqtt] | 319.2s (5.3m) | dataset_export
stream_export[media] | 317.4s (5.3m) | stream_export
dataset_export[rest] | 316.9s (5.3m) | dataset_export
mgmt_action_export[yaml-custom] | 315.6s (5.3m) | mgmt_export
event_group_export[onvif] | 315.3s (5.3m) | event_export
opcua_event_lifecycle | 315.1s (5.3m) | events_int
custom_stream_lifecycle | 311.9s (5.2m) | streams_int
event_group_export[sse] | 310.9s (5.2m) | event_export
event_export[csv-opcua] | 310.0s (5.2m) | event_export
mgmt_group_export[custom] | 309.9s (5.2m) | mgmt_export
datapoint_export[csv-opcua] | 308.7s (5.1m) | dataset_export
mgmt_group_export[opcua] | 307.8s (5.1m) | mgmt_export
mgmt_group_export[onvif] | 307.6s (5.1m) | mgmt_export
event_export[csv-sse] | 306.7s (5.1m) | event_export
stream_export[custom] | 306.5s (5.1m) | stream_export
onvif_mgmt_lifecycle | 302.9s (5.0m) | mgmt_int

- 48 of 49 tests fall in a tight 303-572s range (avg ~370s) - 1 outlier: media_stream_lifecycle at 816s
- The lifecycle tests (no parametrize) tend to be slower (avg 448s) than export/import tests (avg 360s) — they do create/update/delete full lifecycle, while export/import is mostly read-oriented
- The [yaml-opcua] format variants are consistently the slowest in their group (~420-434s) because they're the first parametrize variant that triggers format_test_asset_cache creation (creating the shared asset from scratch)
- Group-level exports (`*_group_export`) are consistently faster (~310-336s) than individual-item exports (`*_action/datapoint/event_export` ~306-434s)


.github/workflows/int_test.yml
```diff
@@ -278,7 +278,7 @@ jobs:
         run: |
           PYTEST_ARGS=""
           if [[ "${{ matrix.scenario.parallel }}" == "true" ]]; then
-            PYTEST_ARGS="--dist=loadfile -n auto"
+            PYTEST_ARGS="-n 6 --dist worksteal"
```
cheatsheet1999, line 281: Nice improvement on the performance, this looks really solid.

Non-blocking: one question on the worker utilization: how did we arrive at -n 6?

In general, scaling workers horizontally doesn’t always translate to better performance (e.g. CPU/IO contention, external dependency limits), so I’m wondering if we have any data or experiments backing this number.

digimaun: Also wondering on reasons for arriving at 6. The standard tier github runners have 4 vCPU, so there should be reason(s) the oversubscription is worth it. Is hardcoding 6 better than setting `-n auto`? If so, lets add a comment in the code that rationalizes it. What about when ITs run in misc environments including locally with a mix of CPU cores.

.github/actions/connect-arc/action.yml
```diff
@@ -48,18 +48,35 @@ runs:
...
+          if [ $attempt -eq $max_retries ]; then
+            echo "Failed to connect cluster to ARC after $max_retries attempts."
+            exit 1
+          fi
+
+          echo "Connection attempt failed. Retrying in ${retry_delay}s..."
+          sleep $retry_delay
+        done
+
+        az connectedk8s enable-features --features custom-locations cluster-connect \
```
digimaun, line 75: While `az connectedk8s connect` is now in a retry loop, what about enable-features? We don't think it warrants a retry loop due to difference in behavior compared to connect?

.github/workflows/int_test.yml
HangyiWang: Thanks for the comment. 

Some context here:

| Config | Workers | Distribution | Wall Clock |
|--------|---------|-------------|------------|
| Dev (`-n auto`) | 4 (gw0-gw3) | Only 2 active — gw0: 35 tests, gw1: 14, gw2-gw3: idle | ~2h |
| This PR (`-n 6 --dist worksteal`) | 6 (gw0-gw5) | All active — 7-9 tests each, evenly distributed | ~69m |

image

Over the past few days I tested scaling from `-n auto` (resolved to 2) → 4 → 6. The 2→4 jump was the biggest win (~4h → ~2h). The 4→6 gain was more about worker balance than raw speedup — all workers stay busy with no idle/wait issues. 6 also gives us headroom as we add more tests.

These tests are I/O-bound, spending ~95% of the time waiting on az CLI → ARM API → K8s round-trips. The GitHub-hosted runner (4 vCPU / 16 GB) barely uses CPU, so worker contention isn’t a concern.

**The real constraint is cluster-side concurrency:** 6 workers means up to 6 simultaneous API calls hitting the same AIO instance. ARM rate limits aren't a concern, but the K8s apiserver and AIO operators can slow under concurrent CRD create/delete pressure. We observed a **~45%** per-test slowdown going from `loadfile` (some workers idle) to `worksteal` (all workers busy) — suggesting we're approaching the point where more cluster concurrency starts to offset parallelism gains. (but 6-way parallelism still delivers a 3× net wall-time improvement)

Metric | Dev (loadfile) | Improved (worksteal)
-- | -- | --
Wall time | 12,339s | 3,989s
Total CPU time | ~17,274s | ~23,934s
Avg per test | ~352s | ~509s
Per-test change | — | +45% slower



With `worksteal` + 6 workers, aggregate test time is ~317 min across 49 tests. Theoretical floor at perfect balance: ~54 min. Actual: 69 min (gap mostly due to one 816s stream integration test outlier test). Going to more workers (8+) would theoretically help but adds session setup overhead (~60s per extra worker) and more cluster pressure.

.github/actions/connect-arc/action.yml
HangyiWang: Good call so the `connect` command is flaky because it handles initial cluster registration, which is prone to transient failures — I've hit this many times, so the retry loop in this PR.

`enable-features` is a simpler operation on a ready cluster, so it's much less likely to fail transiently. And since it runs immediately after connect, there's a small chance of propagation delays. I can wrap it in a lighter retry loop (e.g., 2 attempts) for extra robustness if you think it's worth it.

digimaun: Makes sense. I don't have an opinion right now on adding the loop for enable-features, can close as is and follow up if needed.

digimaun (approved):

cheatsheet1999 (approved):

.github/workflows/int_test.yml
cheatsheet1999: Thanks @HangyiWang, sounds like the worker number has already been validated through multiple runs, I don’t have any strong objections to it.

HangyiWang: Thanks, no problem. Actually, we can use either 4 or 6; it won’t significantly affect the runtime.

digimaun: What about `-n auto --dist worksteal`? The worksteal distribution already addresses the idle worker problem you observed. -n auto would give us 4 workers on the current runner, a small hit vs 6, but it automatically adapts if we ever change runner size or GitHub changes the runner specs. Seems like less maintenance burden for a marginal performance difference, especially since the big win was going from 2 to 4.

HangyiWang: Yep, I haven’t tried it yet—mainly because I’m not sure how the auto flag works. Not sure how many workers it uses (guessing ~4). I’ll test it later.

HangyiWang: BTW @digimaun for your previous comment — this isn’t for local IT runs. I intentionally avoided changing local setup since we usually run only the specific ITs we care about locally, not all of them, so parallelism isn't a big issue.

HangyiWang:
https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/23454996832 
latest: 1h53m
image

PR #852, opened 2026-03-19
fix(check): add deviceregistry Device evaluation for namespaced API

ketkimnaik:
We have namespace devices now  in `az iot ops check --svc deviceregistry` :

image

With status object:

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/check/deviceregistry.py
```diff
@@ -43,22 +44,39 @@ def check_deviceregistry_deployment(
...
+
+    results.extend(check_post_deployment(
+        api_info=NAMESPACED_DEVICEREGISTRY_API_V1,
+        check_name="enumerateNamespacedDeviceRegistryApi",
+        check_desc="Enumerate Namespaced Device Registry API resources",
+        resource_name=resource_name,
+        evaluate_funcs=namespaced_evaluate_funcs,
+        as_list=as_list,
+        detail_level=detail_level,
+        resource_kinds=resource_kinds,
```
digimaun: The command supports filtering by resource type via `--resources`. One part is adding the namespace devices and assets to that list. But here, what is the experience if the user passes in `az iot ops check --svc deviceregistry --resources assetendpointprofile`  or in the higher up `check_post_deployment(
        api_info=DEVICEREGISTRY_API_V1, ...)` processes `az iot ops check --svc deviceregistry --resources device`?

digimaun: Or another way to put is if it's better to have conditions based on if the resource kind belongs to the API.

azext_edge/tests/edge/checks/conftest.py
```diff
@@ -66,6 +66,15 @@ def mock_generate_deviceregistry_asset_target_resources(mocker):
     yield patched
 
 
+@pytest.fixture
+def mock_generate_deviceregistry_device_target_resources(mocker):
+    patched = mocker.patch(
+        "azext_edge.edge.providers.check.deviceregistry.generate_target_resource_name",
```
digimaun: an observation that this is the same patch as the fixture above it, so if there is a case that they both need to be used at the same time, the last patch wins

azext_edge/tests/edge/checks/test_deviceregistry_checks_unit.py
```diff
@@ -494,3 +501,110 @@ def test_asset_endpoint_profiles_checks(
...
+def test_devices_checks(
+    mocker,
+    devices,
+    namespace_conditions,
+    namespace_evaluations,
+    mock_generate_deviceregistry_device_target_resources,
+    detail_level,
+    resource_name,
+):
+    mocker = mocker.patch(
```
digimaun: `mocker` is a special name in this context, lets call this variable something else. Maybe `mock_get_resources`?

azext_edge/edge/providers/check/deviceregistry.py
ketkimnaik: Fixed — resource_kinds is now pre-filtered per API group before being passed to each check_post_deployment call. If a requested kind doesn't belong to that group, the entire call is skipped rather than inadvertently running all evaluations with an empty filter.

azext_edge/tests/edge/checks/conftest.py
ketkimnaik: Good catch! I merged both fixtures into a single one using a side_effect that dispatches on api_info.group, so the correct target name is returned regardless of which API group is under test. mock_generate_deviceregistry_device_target_resources is now just an alias pointing to the same fixture, so there's only ever one patch registered on generate_target_resource_name per test.

azext_edge/tests/edge/checks/test_deviceregistry_checks_unit.py
ketkimnaik: Renamed to mock_get_resources, but since it was unused (no call assertions needed)  dropped the assignment entirely — the patch is still active.

digimaun (approved):

PR #855, opened 2026-03-23
feat(scripts): add management actions quickstart scripts

digimaun:
Quickstart scripts (bash + PowerShell) that provision a complete
management actions scenario end-to-end: device, endpoint, asset with
management group and action, EG namespace, and mgmt-actions enablement.

- Auto-creates EG namespace from instance metadata when not provided
- Location sourced from ADR namespace (devices/assets must be co-located)
- Protocol-extensible config (OPC UA default, ONVIF commented out)
- PS 5.1 compatible via Write-TempJson + @filepath pattern
- README, .env.example files, .gitignore, .gitattributes included


scripts/mgmt-actions/.env.example.ps1
```diff
@@ -0,0 +1,25 @@
+# Personal overrides — copy this file to .env.ps1 and fill in your values.
```
HangyiWang, line 1: No strong opinion, but why do we need a separate env script to set up the env if they all require manual runs?

HangyiWang (approved):

scripts/mgmt-actions/.env.example.ps1
digimaun: Chatted offline, but putting summary here:

The env file separates user-specific config (instance name, resource group, etc.) from script logic. Users copy .env.example.sh to .env.sh (or .env.example.ps1 to .env.ps1), fill in their values, and run bash quickstart.sh (or ./quickstart.ps1). The quickstart script auto-sources the env file so no separate manual step needed. This keeps personal config out of the main script and gitignored, so users don't accidentally commit their values or lose them on git pull.

PR #858, opened 2026-03-30
feat(adr): add generalized inbound endpoint apply command with schema discovery

ketkimnaik:
Overview:

1. New `az iot ops ns device endpoint inbound apply` command — schema-driven generalized alternative to type-specific subcommands; auto-resolves endpoint version from connector template, validates --endpoint-config against the connector schema, and supports all connector types including OPC UA (special-cased via bundled metadata).
2. --show-template flag with config — fetches and displays a ready-to-use config template for the given --connector-type without making an ARM call (OPC UA reads bundled file directly; others fetch from connectorMetadataRef OCI artifact). --show-template flag with schema — fetches and displays the scehma with type, default values, minimum and maximum values.
3. Integration tests covering happy-path lifecycle, schema discovery, and mutual-exclusion error cases.

Testing:

1. Added integration tests and tested it
2. Tested live endpoints
3. Ran unit tests

Results from live endpoints:

Connector template absence error:
image

If value is null we get what type is the field we get hint:

image

Skip connector check and provided endpoint config error:

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/adr/helpers.py
```diff
@@ -183,41 +182,110 @@ def get_default_dataset(asset: dict, dataset_name: str, create_if_none: bool = F
     return matched_datasets[0]
 
 
+def _detect_shell() -> str:
```
HangyiWang, line 185: why do we need `_detect_shell` here

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +401,236 @@ def add_inbound_endpoint(
...
+            name=instance_name,
+            resource_group_name=instance_resource_group,
+        )
+        opcua_mode = (
+            instance.get("properties", {})
+            .get("features", {})
+            .get("opcua", {})
+            .get("mode")
+        )
+        if opcua_mode == "Disabled":
```
HangyiWang, line 431: could `Disabled` be lowercase?

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +401,236 @@ def add_inbound_endpoint(
...
+        is_opcua = connector_type.lower() == DeviceEndpointType.OPCUA.value.lower()
+
+        # --show-schema: return schema and exit early (no device/name/address needed)
+        if show_schema:
+            if is_opcua:
+                import os as _os
+                _schema_file = _os.path.join(
+                    _os.path.dirname(_os.path.abspath(__file__)),
+                    "schemas",
+                    "opcua_connector_metadata.json",
```
HangyiWang: line 446 has an exception check for same file open? perhaps we can extract a helper function here for load opcua metadata

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +401,236 @@ def add_inbound_endpoint(
...
+            )
+            # slim the endpointConfig portion if present
+            if isinstance(raw, dict) and "endpointConfig" in raw:
+                raw["endpointConfig"] = _slim_schema(raw["endpointConfig"])
+            return raw
+
+        # Provider-level required arg guards (CLI enforces these too, but provider
+        # may be called directly in tests or other code paths).
+        from azure.cli.core.azclierror import RequiredArgumentMissingError
+        if not device_name:
```
HangyiWang, line 662: sorry why not check required arguments at the beginning of the function?

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +401,236 @@ def add_inbound_endpoint(
...
+            raise RequiredArgumentMissingError("--name is required.")
+        if not endpoint_address:
+            raise RequiredArgumentMissingError("--endpoint-address is required.")
+
+        # --skip-connector-check only makes sense when endpoint_config is absent;
+        # if the user supplies endpoint_config we must validate a template exists.
+        if skip_connector_check and endpoint_config:
+            raise InvalidArgumentValueError(
+                "--skip-connector-check cannot be used when --endpoint-config is provided.\n"
+                "Create or verify a connector template first: az iot ops connector template create ..."
```
HangyiWang, line 682: nit: is it a convention to use three dots at the end of the command? (in the error message)

azext_edge/edge/providers/adr/params.py
```diff
@@ -839,6 +839,56 @@ def _register_export_import_args(
...
+        )
+        context.argument(
+            "endpoint_name",
+            options_list=["--name", "-n"],
+            help="Endpoint name.",
+            required=True,
+        )
+        context.argument(
+            "endpoint_address",
+            options_list=["--endpoint-address", "--address"],
```
HangyiWang, line 870: --ea?

azext_edge/edge/providers/adr/params.py
```diff
@@ -839,6 +839,56 @@ def _register_export_import_args(
...
+        context.argument(
+            "endpoint_config",
+            options_list=["--endpoint-config", "--ec"],
+            help="Inline JSON string or path to a JSON/YAML file (.json, .yaml, .yml) containing "
+            "connector-specific endpoint configuration. The schema can be discovered with --show-schema. "
+            "Cannot be combined with --skip-connector-check.",
+        )
+        context.argument(
+            "show_schema",
+            options_list=["--show-schema"],
```
HangyiWang: 😆  idk when to add short arguments or not - but --ss? looks strange though

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -74,126 +74,38 @@ def get_endpoint_version_for_type(
...
+        template = self.get_connector_template_for_type(
+            instance_name=instance_name,
+            instance_resource_group=instance_resource_group,
+            connector_type=endpoint_type,
+            look_for_mcr=look_for_mcr,
+        )
+        if template is None:
+            logger.info(
+                f"No connector template found for endpoint type '{endpoint_type}'. "
+                "Endpoint version will be None."
```
HangyiWang: why do we mention endpoint version None here

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +544,178 @@ def delete(
             )
             return wait_for_terminal_state(poller=poller, logger=logger, **kwargs)
 
+    def get_connector_template_for_type(
+        self,
+        instance_name: str,
+        instance_resource_group: str,
+        connector_type: str,
+        look_for_mcr: Optional[bool] = None,
```
HangyiWang, line 549: nit: the naming `look_for_mcr` not natural for me

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +544,178 @@ def delete(
...
+
+            matching = []
+            for template in connector_templates:
+                properties = template.get("properties", {})
+
+                # Apply MCR/non-MCR filter when requested
+                if look_for_mcr is not None:
+                    connector_metadata_ref = properties.get("connectorMetadataRef", "")
+                    is_mcr_template = connector_metadata_ref.startswith(DEFAULT_REGISTRY_HOST)
+                    if look_for_mcr and not is_mcr_template:
```
HangyiWang: `if look_for_mcr != is_mcr_template`?

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +544,178 @@ def delete(
...
+                        tag_settings = image_config.get("tagDigestSettings", {})
+                        tag_version = tag_settings.get("tag", "0.0.0")
+                        matching.append((tag_version, template))
+                        break
+
+            if not matching:
+                return None
+
+            if len(matching) == 1:
+                return matching[0][1]
```
HangyiWang, line 606: why is `matching` a 2d array

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +544,178 @@ def delete(
...
+                logger.warning(
+                    f"Failed to fetch endpoint config schema from '{schema_ref}': {e}."
+                )
+
+        # Path 2: fetch connectorMetadataRef and extract additionalConfigurationSchema
+        metadata_ref = properties.get("connectorMetadataRef")
+        if metadata_ref:
+            try:
+                oci_client = get_oci_client()
+                artifact_info = oci_client.fetch_first_layer(
```
HangyiWang, line 707: ot sure if we still need the `fetch_first_layer` from oci_client haha but seems like there's only one layer always? i may misundestand

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +401,236 @@ def add_inbound_endpoint(
...
+            )
+
+        # OPC UA: no connector template — use bundled metadata for enabled-check and version
+        if is_opcua and not skip_connector_check:
+            opcua_metadata = self._get_opcua_info(instance_name, instance_resource_group)
+            if endpoint_version is None:
+                endpoint_version = opcua_metadata.get("version")
```
peolivei2: Did you check if DOE is also setting a version for OPC endpoints? (asking because this is using the connector version, rather than the endpoint version).

azext_edge/edge/providers/adr/params.py
digimaun: Not --ss that is sensitive term

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -500,6 +733,39 @@ def _get_endpoints(device: dict, inbound: bool = True) -> dict:
     return device_endpoints.get("inbound", {}) if inbound else device_endpoints
 
 
+def _slim_schema(schema: dict) -> dict:
```
peolivei2: Couple questions:
-  Are `$ref`s already resolved when we get to this point?
- Did you validate this against the schemas that we have currently? I thought we had schemas using thigs like `OneOf`.

peolivei2 (approved):

azext_edge/edge/providers/adr/params.py
HangyiWang: kk

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Based on my verification:

$ref: Not applicable here — _slim_schema only processes the additionalConfigurationSchema from connector metadata, which is an inline schema embedded directly in the metadata JSON. There are no external $ref pointers to resolve.

oneOf/anyOf: Validated against all current connector schemas — OPC UA (bundled) and MQTT (fetched live from mcr.microsoft.com/azureiotoperations/akri-connectors/mqtt-metadata:1.0.1). Neither uses oneOf/anyOf. The oneOf you may be thinking of exists in the asset/dataset config schemas (e.g. opcua_dataset_config_v2.json), which are separate from endpoint additional configuration and not processed by _slim_schema.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Good catch — updated to follow DOE's pattern: we no longer set endpoint_version from the OPC UA metadata, passing None instead and letting ADR manage the version.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: You're right that OCI metadata artifacts always have exactly one layer (the JSON blob) — the name reflects the OCI spec where content is stored as manifest layers. Since these aren't multi-layer container images, fetching the first (and only) layer is the correct and intentional approach. I didn't change it because it was from earlier PR just moved it.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: It's a list of (tag_version, template) tuples — the version is stored alongside the template so it's available for the semver comparison loop below without re-extracting it each iteration.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Yeah, better option

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: This was introduced in an earlier PR — keeping as-is to avoid unnecessary churn.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Good catch — removed the redundant "Endpoint version will be None" since return None already conveys that.

azext_edge/edge/providers/adr/params.py
ketkimnaik: I agree it would look bit strange. I think short aliases in this codebase are only added for flags users type frequently. Hence, I don't think we need for this case.

azext_edge/edge/providers/adr/params.py
ketkimnaik: Followed the existing convention.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: The guards are a defensive safety net for direct programmatic callers (e.g. unit tests) that bypass CLI parsing. They're placed after show_schema because show_schema=True is a valid call path without device/name/address — moving them to the top would break that path.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Yeah, I checked that ... is used consistently across the codebase in example commands (including in _log.py and connector_templates.py).

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Yes, I created separate function to avoid duplication

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: "Disabled" is PascalCase per the ARM enum.

azext_edge/edge/providers/adr/helpers.py
ketkimnaik: Good point — removed both _detect_shell and _inline_json_quoting_hint; error message now follows our common pattern. I was experimenting something but seems unecessary.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
HangyiWang: thanks I guess the oci client is from my PR lol

peolivei2 (approved):

azext_edge/edge/providers/adr/_help.py
```diff
@@ -735,11 +735,57 @@ def load_iotops_adr_help():
             az iot ops ns device endpoint inbound remove --device mydevice --instance myInstance -g myInstanceResourceGroup --endpoint myEndpoint1 myEndpoint2
     """
 
+    helps[
+        "iot ops ns device endpoint inbound create"
```
digimaun: Proposing a rename of `inbound create` to `inbound apply`. `--replace` makes it an upsert, and the planned behavior overwrites an existing entry on the device. `--show-template` emits a config you submit back via `--endpoint-config` which is an apply style flow. Given the schema driven, declarative nature of `--endpoint-config`, `apply` matches user expectations better than create.

A consideration for this, with apply semantics, `--replace` becomes redundant, `apply` is naturally idempotent/overwriting. We could either (a) drop `--replace` and have `apply` always overwrite, or (b) keep it as a safety guard (default to deny on existing entry, require `--replace` or `--yes` to overwrite).

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,234 @@ def add_inbound_endpoint(
...
+                )
+            template_mode = show_template.lower()
+            if is_opcua:
+                opcua_metadata = _load_opcua_metadata_file()
+                # Also call _get_opcua_info to allow instance feature-check when
+                # instance args are supplied (e.g. in tests that mock this method).
+                try:
+                    opcua_metadata = self._get_opcua_info(instance_name, instance_resource_group)
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
ketkimnaik: Done

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,234 @@ def add_inbound_endpoint(
...
+                "--skip-connector-check cannot be used when --endpoint-config is provided.\n"
+                "Create or verify a connector template first: az iot ops connector template create ..."
+            )
+
+        # OPC UA: no connector template — use bundled metadata for enabled-check only.
+        # Version is intentionally left as None (let ADR manage it), matching DOE behavior.
+        if is_opcua and not skip_connector_check:
+            self._get_opcua_info(instance_name, instance_resource_group)
+
+        # Connector template lookup for all other types (only needed when endpoint_config is provided)
```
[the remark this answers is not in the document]
ketkimnaik: It should not perform connector template check when skip-conector -config is present. the logic is right.

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,234 @@ def add_inbound_endpoint(
...
+                    endpoint_type=connector_type,
+                )
+
+        # Process endpoint config from file
+        additional_configuration = None
+        if endpoint_config:
+            additional_configuration = process_additional_configuration(
+                additional_configuration=endpoint_config,
+                config_type="endpoint",
+            )
```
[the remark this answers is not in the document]
ketkimnaik: I have added validate_data_against_schema for data validation.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +541,176 @@ def delete(
...
+            semver = scoped_semver_import()
+            latest_version = None
+            latest_template = None
+            for tag_version, template in matching:
+                try:
+                    parsed = semver.parse(tag_version)
+                    if latest_version is None or parsed > semver.parse(latest_version):
+                        latest_version = tag_version
+                        latest_template = template
+                except (ValueError, AttributeError):
```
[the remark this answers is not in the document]
ketkimnaik: Done. Applied or clause and added typeerror.

azext_edge/edge/providers/adr/_help.py
```diff
@@ -735,11 +735,57 @@ def load_iotops_adr_help():
...
+    ] = """
+        type: command
+        short-summary: Add an inbound endpoint to a device using a generalized connector-type approach.
+        long-summary: |
+          The generalized add command is schema-driven. For non-OPC UA connector types it looks up
+          an existing connector template for the specified --connector-type, auto-resolves the
+          endpoint version, and optionally validates --endpoint-config against the connector schema.
```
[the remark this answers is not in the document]
ketkimnaik: I have added schema validation now.

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -183,41 +182,110 @@ def get_default_dataset(asset: dict, dataset_name: str, create_if_none: bool = F
...
+
+
+def _inline_json_quoting_hint() -> str:
+    """Return a shell-specific quoting hint for inline JSON."""
+    shell = _detect_shell()
+    if shell == "powershell":
+        return (
+            "Detected shell: PowerShell. "
+            "Wrap the JSON in single quotes and escape inner double quotes with backslash:\n"
+            "  --endpoint-config '{\\\"key\\\": \\\"value\\\", \\\"nested\\\": {\\\"k\\\": 1}}'"
```
HangyiWang: why sometimes 3 \\\ sometimes 1 \? like line 218

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,234 @@ def add_inbound_endpoint(
...
+                }
+            }
+        }
+
+        with console.status(f"Updating inbound endpoints for {device_name}..."):
+            poller = self.ops.begin_update(
+                resource_group_name=namespace["resource_group"],
+                namespace_name=namespace["name"],
+                device_name=device_name,
+                properties=update_payload,
```
HangyiWang, line 753: just double checking - `update_payload` here is `{"properties": {}`, do we need to extract properties and pass to properties parameter in `begin_update`?

azext_edge/edge/providers/adr/namespace_devices.py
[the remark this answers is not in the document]
HangyiWang: @ketkimnaik what if not skip_connector and not endpoint_config?

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,234 @@ def add_inbound_endpoint(
...
+        is_opcua = connector_type.lower() == DeviceEndpointType.OPCUA.value.lower()
+
+        # --show-template: return config template and exit early (no device/name/address needed)
+        if show_template:
+            if endpoint_config:
+                raise InvalidArgumentValueError(
+                    "--show-template and --endpoint-config cannot be used together. "
+                    "--show-template displays the template and exits without creating an endpoint."
+                )
+            template_mode = show_template.lower()
```
HangyiWang: nit: do we need validation for template_mode since it should only allow few constants?

HangyiWang (approved):

azext_edge/edge/providers/adr/helpers.py
HangyiWang: are you removing this function or it's used still

ketkimnaik: I removed it earlier. Afterwards I had discussion with Paymaun and we discussed that it is better to have this function. we have different types of terminals used by users so this detailed error message will help users to understand the exact format.

azext_edge/edge/providers/adr/namespace_devices.py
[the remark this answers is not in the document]
ketkimnaik: I updated the comment. It will not perform connector template check when skip flag is provided. In all other cases it will perform the check.

azext_edge/edge/providers/adr/_help.py
ketkimnaik: Yeah, apply sounds good. I have updated it and I prefer option b to keep replace as safety guard. I have implemented it like that.

azext_edge/edge/providers/adr/helpers.py
ketkimnaik: Different shell accepts different inline json format. That's why we have this fuction to detect and help user with correct json format if they hit the error.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: The properties parameter of begin_update accepts the full patch body ({"properties": {...}}), not just the inner dict — consistent with the other three begin_update calls in this file. No change needed.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: --show-template is declared with arg_type=get_enum_type(EndpointTemplateMode) in params.py, so the CLI already rejects invalid values. No extra validation needed here.

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -183,41 +182,110 @@ def get_default_dataset(asset: dict, dataset_name: str, create_if_none: bool = F
...
+    # Unix-like shells set $SHELL
+    shell_path = os.environ.get("SHELL", "")
+    if "zsh" in shell_path:
+        return "zsh"
+    if "bash" in shell_path:
+        return "bash"
+    # Windows CMD
+    if sys.platform == "win32":
+        return "cmd"
+    return "unknown"
```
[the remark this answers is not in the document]
ketkimnaik: Done

azext_edge/edge/providers/adr/helpers.py
```diff
@@ -183,41 +182,110 @@ def get_default_dataset(asset: dict, dataset_name: str, create_if_none: bool = F
...
-        error_msg = f"{config_type.capitalize()} configuration is not a valid JSON. "
-        if inline_json:
-            error_msg += "For examples of valid JSON formating, please see https://aka.ms/inline-json-examples "
-        raise InvalidArgumentValueError(
-            f"{error_msg}\n{e.msg}"
-        )
+        # Parse the file content — supports JSON and YAML transparently
+        try:
+            parsed = deserialize_file_content(additional_configuration)
+        except Exception:
```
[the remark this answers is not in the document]
ketkimnaik: done

azext_edge/edge/command_map.py
```diff
@@ -297,6 +297,7 @@ def load_iotops_commands(self, _):
     ) as cmd_group:
         cmd_group.command("list", "list_inbound_device_endpoints")
         cmd_group.command("remove", "remove_inbound_device_endpoints")
+        cmd_group.command("apply", "add_inbound_device_endpoint")
```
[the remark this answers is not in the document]
ketkimnaik: Updated

azext_edge/edge/providers/adr/_help.py
```diff
@@ -735,11 +735,57 @@ def load_iotops_adr_help():
...
+        short-summary: Apply an inbound endpoint to a device using a generalized connector-type approach.
+        long-summary: |
+          The generalized apply command is schema-driven. For non-OPC UA connector types it looks up
+          an existing connector template for the specified --connector-type, auto-resolves the
+          endpoint version, and optionally validates --endpoint-config against the connector schema.
+
+          OPC UA (Microsoft.OpcUa) is a special case: it does not use Akri connector templates.
+          Its schema and version are derived from bundled metadata. The instance must have OPC UA
+          enabled (feature.opcua.mode != Disabled).
+
```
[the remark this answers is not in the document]
ketkimnaik: I don't think it is needed. It right currently.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -632,6 +541,176 @@ def delete(
...
+        device_endpoint_types = properties.get("deviceInboundEndpointTypes", [])
+        schema_refs = []
+        for endpoint_type_info in device_endpoint_types:
+            et = endpoint_type_info.get("endpointType")
+            if et and et.lower() == connector_type.lower():
+                raw = endpoint_type_info.get("configurationSchemaRefs", [])
+                schema_refs = raw if isinstance(raw, list) else [raw]
+                break
+
+        if schema_refs:
```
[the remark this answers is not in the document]
ketkimnaik: Done

azext_edge/tests/edge/adr/test_namespace_devices_unit.py
```diff
@@ -2425,3 +2431,914 @@ def test_add_inbound_mqtt_device_endpoint(
...
+    return {
+        "name": f"template-{connector_type.lower()}",
+        "properties": {
+            "connectorMetadataRef": f"mcr.microsoft.com/azureiotoperations/{connector_type.lower()}-metadata:1.0",
+            "deviceInboundEndpointTypes": [
+                {
+                    "endpointType": connector_type,
+                    "version": version,
+                    "configurationSchemaRefs": schema_refs or [],
+                }
```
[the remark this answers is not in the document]
ketkimnaik: done

ketkimnaik:
Successful Integration test run: https://github.com/ketkimnaik/azure-iot-ops-cli-extension/actions/runs/25182308692/job/73857096871

HangyiWang (approved):

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,266 @@ def add_inbound_endpoint(
...
+            if not skip_connector_check:
+                if is_opcua:
+                    from .specs import NAMESPACE_DEVICE_OPCUA_ENDPOINT_SCHEMA as _endpoint_schema
+                else:
+                    _endpoint_schema = connector_templates.get_endpoint_schema(
+                        instance_name=instance_name,
+                        instance_resource_group=instance_resource_group,
+                        connector_type=connector_type,
+                    ).get("endpointConfig", {})
+                if _endpoint_schema:
```
digimaun: If _endpoint_schema comes back as falsey/{}, for example because the OCI fetch in `get_endpoint_schema` failed or no connectorMetadataRef was present. In that case the whole if is skipped silently and the endpoint gets created as if validation passed. Since the user opted into validation by not passing `--skip-connector-check`, I think this should raise `ValidationError` with a message saying schema retrieval failed and that they can rerun with the flag if they want to bypass it. It would also help to log the swallowed exception inside `get_endpoint_schema` at debug level so the underlying cause is recoverable with `--debug`.

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,266 @@ def add_inbound_endpoint(
...
+        # --show-template: return config template and exit early (no device/name/address needed)
+        if show_template:
+            if endpoint_config:
+                raise InvalidArgumentValueError(
+                    "--show-template and --endpoint-config cannot be used together. "
+                    "--show-template displays the template and exits without creating an endpoint."
+                )
+            template_mode = show_template.lower()
+            if is_opcua:
+                opcua_metadata = _load_opcua_metadata_file()
```
digimaun: The bundled metadata gets loaded unconditionally here, and then _get_opcua_info runs and loads the same file again, so the JSON is parsed twice on every show.

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,266 @@ def add_inbound_endpoint(
...
+                # the ARM call may fail with a not-found error; fall back to the bundled
+                # metadata in that case.  ValidationError (OPC UA explicitly Disabled) is
+                # intentionally NOT caught so the disabled-check still propagates.
+                if instance_name and instance_resource_group:
+                    from azure.core.exceptions import HttpResponseError
+
+                    try:
+                        opcua_metadata = self._get_opcua_info(instance_name, instance_resource_group)
+                    except HttpResponseError:
+                        pass  # instance / resource-group not found – use bundled metadata
```
digimaun: This exception handling would cover 401, 403, throttling, and network failures the same as a 404. The catch could be narrowed to just 404/ResourceNotFoundError and let the other errors surface.

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -500,6 +780,104 @@ def _get_endpoints(device: dict, inbound: bool = True) -> dict:
     return device_endpoints.get("inbound", {}) if inbound else device_endpoints
 
 
+def _slim_schema(schema: dict, mode: str = EndpointTemplateMode.CONFIG.value) -> dict:
```
digimaun: Should this handle items, oneOf, anyOf, allOf, $ref, or schema-valued additionalProperties? If a connector schema uses any of those, this functionality would produce a --show-template output that is missing fields?

azext_edge/edge/providers/adr/namespace_devices.py
```diff
@@ -398,6 +403,266 @@ def add_inbound_endpoint(
...
+            device_name=device_name,
+            instance_name=instance_name,
+            resource_group=instance_resource_group,
+        )
+        namespace = parse_resource_id(device["id"])
+        original_endpoints = _get_endpoints(device)
+
+        if endpoint_name in original_endpoints and not replace:
+            raise InvalidArgumentValueError(
+                f"Inbound endpoint '{endpoint_name}' already exists. Use --replace to update it."
```
digimaun: Quick consistency thought: the other apply commands (broker listener apply, broker authn apply, broker authz apply, dataflow apply, dataflow endpoint apply) all upsert directly via begin_create_or_update without a `--replace` flag.

If someone does:

`--show-template config > config.json && apply --endpoint-config config.json` ... that only works the first time, second time `--replace` is necessary.

One thought using `--no-replace` mechanics instead, so `apply` by default remains an upsert but if the user wants to prevent replace, they opt-in.

ketkimnaik: I removed replace and to be safe added no-replace flag.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: In case of anyOf and oneOf case for config option I have added logic to use first non-null value and at the top show warning like in below pic:
image

for allOf I am showing all the fields for config option.

In schema mode everything is shown:

image

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Done

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: Fixed — _load_opcua_metadata_file() is now only called as a fallback when _get_opcua_info wasn't invoked or threw HttpResponseError; otherwise the metadata returned by _get_opcua_info is used directly.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: I have added validationerror.

azext_edge/edge/providers/adr/namespace_devices.py
ketkimnaik: The current implementation handles all of these. _slim_schema now supports $ref (Draft-07 #/definitions/... via _resolve_ref), oneOf (via _slim_oneof — collapses to first non-null in config mode with a warning, preserves all variants in schema mode), allOf (via _slim_allof — merges properties in config mode), and items (array fields produce a concrete example entry). anyOf and schema-valued additionalProperties are intentionally unsupported — matching the Fluent UI form library that renders these schemas in the portal, which also ignores them.

digimaun (approved):

PR #859, opened 2026-04-01
feat(secretsync): add secret management commands

ketkimnaik:
Overview:

- Adds az iot ops secretsync secret command group with set, list, and unset commands to manage individual SecretSync resources and their AKV secret mappings via CLI.

- set creates or merges secrets into a SecretSync (idempotent), validates AKV secrets exist, and keeps the shared SPC in sync.

- unset uses a ref-count guard (ARG query across all SecretSyncs in the custom location) to safely remove secrets from the shared SPC only when no other SecretSync references them. Auto-deletes the SecretSync when the last secret is removed.

Testing:

- Added unit and integration tests

image

- Performed liveendpoints testing for following scenaios:
image

image

image

image

image

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/_help.py
```diff
@@ -2119,6 +2119,90 @@ def load_iotops_help():
...
+        short-summary: Manage individual secrets within SecretSync resources.
+    """
+
+    helps[
+        "iot ops secretsync secret set"
+    ] = """
+        type: command
+        short-summary: Create or update a SecretSync resource that syncs AKV secrets to K8s.
+        long-summary: |
+            Resolves the instance's default secret provider class (SPC), verifies each AKV secret exists,
```
HangyiWang: nit: we should put one complete sentence in a single line in `long-summary`. (python code length restriction shouldn't be applied here) Otherwise a few single words can end up taking a whole line because the user’s terminal wraps text automatically while they also split text based on our newline characters here. 
Like this PR for CMS commands: https://github.com/Azure/azure-iot-cli-extension/pull/807. pylint/flake8 will skip the helper message file so no problems.

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,34 @@ def load_iotops_arguments(self, _):
...
+            help="Name for the SecretSync ARM resource and the resulting K8s secret.",
+        )
+
+    with self.argument_context("iot ops secretsync secret set") as context:
+        context.argument(
+            "secret_names",
+            options_list=["--secret-name"],
+            action="append",
+            help="AKV secret name to sync, format: <akv-name>=<target-key>. "
+            "Both parts are required. Repeatable. The AKV secret must already exist.",
```
HangyiWang: sorry what's `Repeatable`

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -62,6 +64,7 @@
 SERVICE_ACCOUNT_WASM = "aio-wasm-graph-controller"
 KEYVAULT_ROLE_ID_SECRETS_USER = "4633458b-17de-408a-b874-0445c86b69e6"
 KEYVAULT_ROLE_ID_READER = "21090545-7ca7-4776-b22c-e363652d74d2"
+KEYVAULT_URL = "https://{keyvaultName}.vault.azure.net/"
```
HangyiWang: nit: do we need the last `/`?

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,263 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+                if mapping["akv_name"] not in existing_object_names:
+                    secret_entry = {
+                        "objectName": mapping["akv_name"],
+                        "objectType": "secret",
+                    }
+                    entry_text = yaml.safe_dump(secret_entry, indent=6)
+                    objects_obj["array"].append(entry_text)
+                    existing_object_names.add(mapping["akv_name"])
+
+            spc["properties"]["objects"] = yaml.safe_dump(objects_obj, indent=6).replace("\n- |", "\n    - |")
```
HangyiWang, line 605: no concern, but `"\n- |", "\n    - |"` looks a bit weird

azext_edge/edge/_help.py
digimaun: It's more about the characters per line rather than one sentence in a single line. I think greater than 88 chars leads to auto formatting/wrapping.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,263 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+
+            existing_object_names = set()
+            for entry in objects_obj.get("array", []):
+                entry_obj = yaml.safe_load(entry)
+                if entry_obj and "objectName" in entry_obj:
+                    existing_object_names.add(entry_obj["objectName"])
+
+            for mapping in secret_mappings:
+                if mapping["akv_name"] not in existing_object_names:
+                    secret_entry = {
```
peolivei2, line 595: For certificates we need to add the object encoding as well. We should be able to check the encoding field of the source secret in the AKV.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,263 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+            except ResourceNotFoundError:
+                secret_sync = None
+
+            if secret_sync:
+                # Merge new entries into existing SecretSync
+                existing_mapping = secret_sync.get("properties", {}).get("objectSecretMapping", [])
+                existing_source_paths = {m["sourcePath"] for m in existing_mapping}
+                for mapping in secret_mappings:
+                    if mapping["akv_name"] in existing_source_paths:
+                        # Update existing entry's targetKey
```
peolivei2: You shouldn't update the key of an existing mapping.
Secret syncs support mapping the same secret to multiple keys. A user might have done this on purpose (e.g., mapping the same cert for different uses).

To avoid changes in the design, we can do:
- `set` adds the entry for the source/key if it doesn't exist.
- `unset` removes all mappings for the secret (which is still semantically correct, since unset reads as "remove the secret from the secret sync").

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,263 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+
+        with console.status("Working...") as status:
+            # Step 1: Fetch the named SecretSync
+            secret_sync = self.ssc_mgmt_client.secret_syncs.get(
+                resource_group_name=resource_group_name,
+                secret_sync_name=secret_sync_name,
+            )
+            mappings = secret_sync.get("properties", {}).get("objectSecretMapping", [])
+
+            # Find and remove the entry whose sourcePath == secret_name
```
peolivei2: "entries"

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,263 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+                for m in ss_mappings:
+                    if m.get("sourcePath") == secret_name:
+                        still_referenced = True
+                        break
+                if still_referenced:
+                    break
+
+            # Step 3: Only remove from SPC if no other SecretSync references this secret
+            if not still_referenced:
+                spc = self.get_default_spc(instance_name=name, resource_group_name=resource_group_name)
```
peolivei2: It's possible for a secret sync to reference an SPC other than the default one from the instance.
For semantic correctness, you should:
- Get the actual SPC from the secret sync that you just modified.
- When doing the ref count, consider only secret syncs referencing that SPC.
- When doing the SPC cleanup, clean up that specific SPC.

peolivei2 (changes requested):

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Same pattern used in certs.py (lines 498, 516) and migration.py (line 231). It fixes yaml.safe_dump indentation for the SPC objects field — the ARM API expects array items indented under the key.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Copied from existing KEYVAULT_URL in certs.py for consistency. Happy to drop it if preferred.

azext_edge/edge/params.py
ketkimnaik: We can add secret-name param multiple time. I have added set command example in overview you can check that for reference :)

azext_edge/edge/_help.py
HangyiWang: oh we don't have auto format control here in `_help.py`

azext_edge/edge/params.py
HangyiWang: thanks

azext_edge/edge/providers/orchestration/resources/instances.py
HangyiWang: cool

azext_edge/edge/providers/orchestration/resources/instances.py
HangyiWang: sounds good

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/edge/_help.py
ketkimnaik: Fixed — joined sentences to single lines per Hangyi's suggestion. The CLI framework auto-wraps at ~88 chars as Paymaun noted, so manual breaks are unnecessary.

azext_edge/edge/providers/orchestration/resources/instances.py
peolivei2: Can you check the AKV encoding field instead fo the content type?

peolivei2 (dismissed):

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: The AKV get_secret response doesn't have a separate "encoding" field — it has contentType (e.g., application/x-pkcs12), managed (bool), kid (certificate key ID), and optionally tags.file-encoding. Currently we check contentType for pkcs12 or x-pem. Could you clarify which field you mean? Is it contentType with a different check, or tags.file-encoding?

ketkimnaik: Done. I am using file endcoding to verify now.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,269 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+            for mapping in secret_mappings:
+                try:
+                    secret_response = keyvault_client.get_secret(
+                        vault_base_url=vault_url,
+                        secret_name=mapping["akv_name"],
+                        secret_version="",
+                    )
+                    tags = secret_response.get("tags") or {}
+                    mapping["needs_hex_encoding"] = tags.get("file-encoding", "").lower() == "hex"
+                except (ResourceNotFoundError, HttpResponseError) as e:
```
digimaun: ResourceNotFoundError should already be a subclass of HttpResponseError, so it handles the 404 case on its own. Having HttpResponseError in the same except means other status codes/error responses like 403, 429, or 500 all get reported as "AKV secret not found" which can be pretty confusing to debug.

Dropping HttpResponseError from the clause should be enough to get the friendly error then everything else gets bubbled up with whatever the service returned

azext_edge/tests/edge/orchestration/test_secretsync_secret_int.py
```diff
@@ -0,0 +1,392 @@
...
+        "resourceGroup": resource_group,
+        "instanceName": instance_name,
+        "keyvaultName": kv_name,
+        "testSecrets": test_secrets,
+    }
+
+    # Cleanup: remove test secrets from AKV
+    for secret_name in test_secrets:
+        try:
+            run(f"az keyvault secret delete --vault-name {kv_name} --name {secret_name}")
```
digimaun, line 79: This command may soft delete the if the vault is configured for it, so repeated test runs with the same kv could accumulate soft-deleted secrets or hit naming collisions. Might be worth adding a best-effort purge after the delete.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,269 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+
+    def unset_secretsync_secret(
+        self,
+        name: str,
+        resource_group_name: str,
+        secret_sync_name: str,
+        secret_name: str,
+        confirm_yes: Optional[bool] = None,
+        **kwargs,
+    ):
```
digimaun: nit `-> Optional[dict]`

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,269 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+                resource=secret_sync,
+            )
+            return wait_for_terminal_state(ss_poller, **kwargs)
+
+    def list_secretsync_secrets(
+        self,
+        name: str,
+        resource_group_name: str,
+        secret_sync_name: str,
+    ) -> Optional[List[dict]]:
```
digimaun: This returns None when there are no mappings, which will break az ... -o json | jq length and similar pipeline usage. Other list commands in the extension return an empty list in this case.

list_secretsync is has special handling which reflects whether secret sync is enabled or not for the instance...so None for that means "not applicable" rather than "zero items".

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: Something we need to take into consideration moving forward is support for different cloud profiles (non public cloud) ... a solution here is obtaining the URL from control plane key vault GET.

azext_edge/tests/edge/orchestration/resources/test_secretsync_secret_unit.py
```diff
@@ -0,0 +1,701 @@
...
+        content_type="application/json",
+    )
+
+    secretsync_secret_set(
+        cmd=mocked_cmd,
+        instance_name=instance_name,
+        resource_group_name=resource_group_name,
+        secret_sync_name=secret_sync_name,
+        secret_names=[f"{akv_name}=certificate"],
+        wait_sec=0.1,
```
digimaun: I think its fine to have wait_sec=0

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +520,269 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+            # Step 4: Update the SPC via ARM
+            spc_poller = self.ssc_mgmt_client.azure_key_vault_secret_provider_classes.begin_create_or_update(
+                resource_group_name=resource_group_name,
+                azure_key_vault_secret_provider_class_name=spc_name,
+                resource=spc,
+            )
+            wait_for_terminal_state(spc_poller, **kwargs)
+
+            # Step 5: Create or update the SecretSync resource
+            instance = self.show(name=name, resource_group_name=resource_group_name)
```
digimaun: Is this self.show() even necessary? I see its used like instance["location"] and "instance["extendedLocation"] those values should match the default SPC already available on line 533

azext_edge/tests/edge/orchestration/test_secretsync_secret_int.py
```diff
@@ -0,0 +1,392 @@
...
+from knack.log import get_logger
+from typing import List
+
+from azure.cli.core.azclierror import CLIInternalError
+
+from ...generators import generate_random_string
+from ...helpers import run
+
+logger = get_logger(__name__)
+SYNC_WAIT_SEC = 15
```
digimaun: Where are these used? SYNC_WAIT_SEC and MAX_RETRIES

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,34 @@ def load_iotops_arguments(self, _):
...
+        context.argument(
+            "secret_sync_name",
+            options_list=["--secret-sync-name"],
+            help="Name for the SecretSync ARM resource and the resulting K8s secret.",
+        )
+
+    with self.argument_context("iot ops secretsync secret set") as context:
+        context.argument(
+            "secret_names",
+            options_list=["--secret-name"],
```
digimaun: I think we should call this `--secret-map` to clearly signal this is a mapping.

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,34 @@ def load_iotops_arguments(self, _):
...
+            action="append",
+            help="AKV secret name to sync, format: `<akv-name>=<target-key>`. "
```
digimaun: Suggest "AKV secret mapping in the format \<akv-secret-name>=\<target-key>. Repeatable. The AKV secret must exist."

azext_edge/edge/command_map.py
```diff
@@ -59,6 +59,14 @@ def load_iotops_commands(self, _):
         cmd_group.command("disable", "secretsync_disable")
         cmd_group.show_command("list", "secretsync_list")
 
+    with self.command_group(
+        "iot ops secretsync secret",
+        command_type=secretsync_resource_ops,
+    ) as cmd_group:
+        cmd_group.command("set", "secretsync_secret_set")
+        cmd_group.command("list", "secretsync_secret_list")
+        cmd_group.command("unset", "secretsync_secret_unset")
```
digimaun: I know unset pairs well with set verbiage wise. Lets also consider whether "remove" makes sense in place of unset. Consistency wise remove is more common in the code base (asset event remove, ns mgmt-endpoint remove, ops broker listener port remove, connector opcua trust remove, etc (more examples). It is used when stripping an entry from a parent resource that continues to exist (same case here).

unset is wrong, this is more of a stylistic call.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done.

azext_edge/tests/edge/orchestration/test_secretsync_secret_int.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/tests/edge/orchestration/resources/test_secretsync_secret_unit.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done

azext_edge/tests/edge/orchestration/test_secretsync_secret_int.py
ketkimnaik: Earlier using them but then were not needed. I missed to clean up them.

azext_edge/edge/params.py
ketkimnaik: Yeah, that sounds good. Renamed it.

azext_edge/edge/params.py
ketkimnaik: Sure.

azext_edge/edge/command_map.py
ketkimnaik: I choose Remove :)

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done.

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,34 @@ def load_iotops_arguments(self, _):
...
+            options_list=["--secret-map"],
+            action="append",
+            help="AKV secret mapping in the format <akv-secret-name>=<target-key>. "
+            "Repeatable. The AKV secret must exist.",
+        )
+
+    with self.argument_context("iot ops secretsync secret remove") as context:
+        context.argument(
+            "secret_name",
+            options_list=["--secret-map"],
```
digimaun: --secret-map should only be on set not from remove, the original parameter name for remove was correct

azext_edge/edge/_help.py
```diff
@@ -2119,6 +2119,79 @@ def load_iotops_help():
...
+            If the SecretSync already exists, new secret entries are merged into it. Existing entries with the same AKV secret name will have their target key updated.
+
+            The --secret-sync-name value becomes the K8s secret name. Consumers reference it via `<secret-sync-name>/<target-key>` for device endpoints, or just `<secret-sync-name>` for dataflow endpoints.
+
+        examples:
+        - name: Create a SecretSync for device endpoint x509 cert auth.
+          text: >
+            az iot ops secretsync secret set --instance myInstance -g myRG
+            --secret-sync-name my-certs
+            --secret-name my-tls-cert=certificate --secret-name my-tls-key=privateKey
```
digimaun: using old parameter

azext_edge/edge/commands_secretsync.py
```diff
@@ -54,3 +54,52 @@ def secretsync_disable(cmd, instance_name: str, resource_group_name: str, confir
...
+
+def secretsync_secret_remove(
+    cmd,
+    instance_name: str,
+    resource_group_name: str,
+    secret_sync_name: str,
+    secret_name: str,
+    confirm_yes: Optional[bool] = None,
+    **kwargs,
+):
```
digimaun: nit, `-> Optional[dict]`

azext_edge/edge/commands_secretsync.py
```diff
@@ -54,3 +54,52 @@ def secretsync_disable(cmd, instance_name: str, resource_group_name: str, confir
...
+        **kwargs,
+    )
+
+
+def secretsync_secret_list(
+    cmd,
+    instance_name: str,
+    resource_group_name: str,
+    secret_sync_name: str,
+) -> Optional[list[dict]]:
```
digimaun: after the change the type needs to be updated, `-> list[dict]`

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,34 @@ def load_iotops_arguments(self, _):
...
+    with self.argument_context("iot ops secretsync secret") as context:
+        context.argument(
+            "instance_name",
+            options_list=["--instance", "-i", "-n"],
+            help="IoT Operations instance name.",
+        )
+        context.argument(
+            "secret_sync_name",
+            options_list=["--secret-sync-name"],
+            help="Name for the SecretSync ARM resource and the resulting K8s secret.",
```
digimaun, line 1567: remove + list, use --secret-sync-name in a different way, this reads like creation (which is appropriate for set).

azext_edge/edge/_help.py
```diff
@@ -2119,6 +2119,79 @@ def load_iotops_help():
...
+    ] = """
+        type: group
+        short-summary: Manage individual secrets within SecretSync resources.
+    """
+
+    helps[
+        "iot ops secretsync secret set"
+    ] = """
+        type: command
+        short-summary: Create or update a SecretSync resource that syncs AKV secrets to K8s.
```
digimaun:
```suggestion
        short-summary: Set AKV secret mappings on a SecretSync resource.
```

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: I don't see a change

azext_edge/edge/params.py
ketkimnaik: Yeah, I have corrected it now :)

azext_edge/edge/_help.py
ketkimnaik: Corrected it.

azext_edge/edge/commands_secretsync.py
ketkimnaik: Done

azext_edge/edge/commands_secretsync.py
ketkimnaik: Done

azext_edge/edge/params.py
ketkimnaik: Done

azext_edge/edge/_help.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -517,6 +521,276 @@ def get_default_spc(self, instance_name: str, resource_group_name: str) -> dict:
...
+                    secret_sync_name=secret_sync_name,
+                    resource=secret_sync,
+                )
+                modified_secret_sync = wait_for_terminal_state(ss_poller, **kwargs)
+
+            # Step 2: Ref-count guard — check if any other SecretSync referencing the same SPC
+            #         still uses this secret
+            instance = self.show(name=name, resource_group_name=resource_group_name)
+            resource_map = self.get_resource_map(instance)
+            all_secretsyncs = resource_map.connected_cluster.get_cl_resources_by_type(
```
digimaun: Because ARG is eventually consistent this may not be up to date. Consider doing the ref count check before modifying the secret sync resource (and omit the target one)

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: Doesn't have to be in this PR, but KEYVAULT_URL here is duplicated in opcua.certs module, makes sense to move it and reference it in common.

azext_edge/edge/params.py
```diff
@@ -1553,6 +1553,46 @@ def load_iotops_arguments(self, _):
...
+        )
+
+    with self.argument_context("iot ops secretsync secret set") as context:
+        context.argument(
+            "secret_sync_name",
+            options_list=["--secret-sync-name"],
+            help="Name for the SecretSync ARM resource and the resulting K8s secret.",
+        )
+        context.argument(
+            "secret_names",
```
digimaun: Why not also change the binding value to secret_map to be consistent? Mismatch may confuse devs moving forward.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Yeah, done.

azext_edge/edge/params.py
ketkimnaik: Yeah, make sense. Done.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: covered in this PR.

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Changed

digimaun (approved):

PR #870, opened 2026-04-22
ci: add support bundle label coverage check to nightly int-test workflow

ketkimnaik:
---

### General Guidelines

Intent for Production

Basic expectations


.github/workflows/int_test.yml
```diff
@@ -289,6 +289,9 @@ jobs:
           name: coverage-${{ matrix.scenario.name }}
           path: ./.coverage
           include-hidden-files: true
+      - name: "Check support bundle label coverage"
+        continue-on-error: true
+        run: bash ./tools/get_all_unlabeled_resources.sh
```
digimaun: We have a comprehensive support bundle test module(s), can we understand the overlap and delta of this script with the support bundle tests and evaluate whether this one-off script is warranted vs updating the support bundle tests with any valuable deltas.

ketkimnaik: The script and existing tests check different directions — existing tests query by label first (kubectl get -l app.kubernetes.io/name=X) so they're blind to resources that lack the label entirely. The script does the inverse: finds resources by name keyword and asserts the label exists, catching labeling regressions the tests can never see. Adding expected_label to the missing services improves the tests but doesn't close this gap. The script is warranted to keep — it's the only check that detects when a resource exists on the cluster but won't be captured by the bundle due to a missing/wrong label.

digimaun: Fair point on directionality. The current tests are label-first, so a missing-label regression is invisible to them, and that gap is worth closing.

My concern is the shape of the fix. A standalone bash script gives us two parallel validation strategies for the same surface: the pytest support-bundle suite, and a heuristic script with its own keyword list, exclusions, and accepted labels. That second copy will drift as services evolve.

I'd rather fold this into the existing test framework as an inverse check. The service tests already declare canonical prefixes and the expected label, so asserting that every matching cluster resource carries that label fits naturally alongside `check_workload_resource_files`. One strategy in pytest, richer assertions, no duplicated service knowledge in shell. The script is fine as a temporary stopgap, but I wouldn't want it to be the steady state.

HangyiWang:
Hi @ketkimnaik  could you paste link/screenshot of this updated pipeline workflow run? Thanks.

.github/workflows/int_test.yml
ketkimnaik: Replaced the workflow step with a proper check_cluster_label_coverage() helper in the int test framework — it queries all resources without a label filter, matches by name prefix, and asserts each carries the expected label. Called it from MQ, schema registry, dataflow, akri, and meso tests. Secretstore and certmanager are excluded as they don't have a common AIO label yet.

ketkimnaik:
> Hi @ketkimnaik could you paste link/screenshot of this updated pipeline workflow run? Thanks.

We won't be needing new step. I have updated the current test to include inverse logic as Paymaun mentioned in earlier comment.

tools/get_all_unlabeled_resources.sh
```diff
@@ -45,6 +45,9 @@ KNOWN_UNLABELED_EXCLUSIONS=(
     "azure-iot-operations/azure-iot-operations-observability-trust-bundle"
```
digimaun: Still necessary to keep this bash script or we can delete it?

azext_edge/tests/edge/support/create_bundle_int/helpers.py
```diff
@@ -182,6 +182,74 @@ def convert_file_names(files: List[str]) -> Dict[str, List[Dict[str, str]]]:
...
+        "vwc": "validatingwebhookconfiguration",
+        "mwc": "mutatingwebhookconfiguration",
+        "crb": "clusterrolebinding",
+    }
+
+    for resource_type in workload_types:
+        full_type = key_to_full_map.get(resource_type, resource_type)
+        try:
+            kubectl_items = run(f"kubectl get {full_type} -A -o json")
+        except CLIInternalError:
```
digimaun: If the kubectl command errors we won't see it right? At least log entries would help (debug or otherwise)

digimaun (dismissed):

tools/get_all_unlabeled_resources.sh
ketkimnaik: Keeping it — the pytest inverse check covers WORKLOAD_TYPES (pods, deployments, configmaps, etc.) but the script also covers serviceaccounts, ingresses, and networkpolicies which aren't in the bundle scope. It's also useful for ad-hoc local debugging against any cluster without needing to run the full int test suite. It serves a different purpose as a standalone diagnostic tool.

azext_edge/tests/edge/support/create_bundle_int/helpers.py
ketkimnaik: Added warning.

HangyiWang (dismissed):

tools/get_all_unlabeled_resources.sh
digimaun: If serviceaccounts/ingresses/networkpolicies aren't in the support bundle scope, then checking their labels isn't validating bundle coverage. And if we do decide they matter, check_cluster_label_coverage() already takes workload_types, so we pass them in at the call site.

Also the int test already runs against any cluster with a single `pytest -k` invocation and gives richer output than the script. We don't need a second tool for that workflow. My point is the script duplicates the service-to-label knowledge that now lives in pytest, and any new service has to be remembered in two places.

digimaun (approved):

ketkimnaik:
Successful integration test run: https://github.com/ketkimnaik/azure-iot-ops-cli-extension/actions/runs/25137113481/job/73678185384

PR #872, opened 2026-04-30
feat(mgmt-actions): set clientIdPrefix on EG MQTT dataflow endpoint

danewalton-msft:
Sets `clientIdPrefix` to the AIO instance name on the Event Grid MQTT dataflow endpoint created in `_setup_eg_dataflow_endpoint`, matching the bicep template (`clientIdPrefix: aioInstance.name`).

Also extends the GET-then-PUT drift detection to compare `clientIdPrefix` so existing endpoints are updated when the value differs (e.g., endpoints created before this change).

### Changes
- Add `"clientIdPrefix": instance_name` to `mqttSettings` in the resource payload.
- Read `existing_client_id_prefix` and include `client_id_prefix_matches` in the configuration-drift check and log message.


azext_edge/tests/edge/orchestration/test_mgmt_actions_unit.py
```diff
@@ -1013,6 +1048,121 @@ def test_existing_endpoint_same_config(self, mocked_cmd, mocked_responses: respo
         # Only the GET call, no PUT
         assert len(mocked_responses.calls) == 1
 
+    def test_existing_endpoint_missing_client_id_prefix_updates(
```
digimaun: `test_existing_endpoint_missing_client_id_prefix_updates` and `test_existing_endpoint_different_client_id_prefix_updates` seem to differ only in clientIdPrefix value in the GET response A `@pytest.mark.parametrize` over existing_prefix would deduplicate while preserving coverage.

azext_edge/edge/providers/orchestration/mgmt_actions.py
digimaun: Lets update the docstring to include clientIdPrefix to the comparisons, then switch 'either' to 'any' in `If either differs ...`

azext_edge/tests/edge/orchestration/test_mgmt_actions_unit.py
```diff
@@ -905,6 +905,40 @@ def test_create_new_system_assigned(self, mocked_cmd, mocked_responses: response
...
+            eg_ctx=eg_ctx,
+            instance_name=instance_name,
+            instance_resource_id=instance_rid,
+            resource_group_name=rg,
+            extended_location=MOCK_EXTENDED_LOCATION,
+            wait_sec=0,
+        )
+
+        put_body = json.loads(mocked_responses.calls[1].request.body)
+        assert put_body["properties"]["mqttSettings"]["clientIdPrefix"] == instance_name
```
digimaun, line 940: Can we `assert len(mocked_responses.calls) == 2`?

digimaun (approved):

PR #892, opened 2026-06-30
feat: add multi-cloud support for Azure Government and China (#889)

ketkimnaik:
* Add multi-cloud (Fairfax/Mooncake) support ---------

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
+++ b/azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
@@ -22,6 +22,7 @@ from ...instances import SECRET_SYNC_RESOURCE_TYPE, Instances
 from .....orchestration.upgrade2 import calculate_config_delta
 from ......util.file_operations import read_file_content, validate_file_extension
 from ......util.queryable import Queryable
+from ......util.cloud_config import CloudConfig
 from ......util.az_client import (
     get_keyvault_client,
     get_ssc_mgmt_client,
@@ -59,6 +60,7 @@ class OpcUACerts(Queryable):
         )
         self.keyvault_client = get_keyvault_client(
             subscription_id=self.default_subscription_id,
+            keyvault_scope=CloudConfig(self.cmd).keyvault_scope,
         )
         self.instance_name = instance_name
         self.resource_group_name = resource_group_name
--- a/azext_edge/edge/util/cloud_config.py
+++ b/azext_edge/edge/util/cloud_config.py
@@ -1,0 +1,133 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+"""
+Centralized cloud configuration.
+
+Single source of truth for cloud-specific endpoints, suffixes, and token scopes.
+Values that the Azure CLI cloud framework provides (ARM, Microsoft Graph, Key Vault,
+Storage, ACR) are read at runtime from ``cmd.cli_ctx.cloud``. Values the framework does
+not provide (Service Bus suffix, Event Grid token audience) are maintained here as
+explicit cloud-name -> value mappings.
+
+When adding new cloud-dependent functionality, resolve values through ``CloudConfig``
+rather than hardcoding public-cloud URLs.
+"""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from azure.cli.core.commands import AzCliCommand
+
+# Known cloud names as reported by ``cmd.cli_ctx.cloud.name``.
+CLOUD_AZURE_PUBLIC = "AzureCloud"
+CLOUD_AZURE_US_GOVERNMENT = "AzureUSGovernment"
+# NOTE: Azure China (Mooncake) is NOT functional today — the Azure IoT Operations service
+# does not have server-side support in Mooncake, so end-to-end flows cannot be validated
+# there yet. The China endpoint/suffix mappings below are intentionally retained so that
+# support is already in place if/when Mooncake gains server-side support in the future.
+# Until then, treat AzureChinaCloud as best-effort / untested.
+CLOUD_AZURE_CHINA = "AzureChinaCloud"
+
+# Service Bus FQDN suffix per cloud. Not provided by the Azure CLI cloud framework.
+SERVICEBUS_SUFFIX_MAP = {
+    CLOUD_AZURE_PUBLIC: "servicebus.windows.net",
+    CLOUD_AZURE_US_GOVERNMENT: "servicebus.usgovcloudapi.net",
+    CLOUD_AZURE_CHINA: "servicebus.chinacloudapi.cn",
+}
+
+# Event Grid token audience per cloud. Not provided by the Azure CLI cloud framework.
+EVENTGRID_AUDIENCE_MAP = {
+    CLOUD_AZURE_PUBLIC: "https://eventgrid.azure.net",
+    CLOUD_AZURE_US_GOVERNMENT: "https://eventgrid.azure.us",
+    CLOUD_AZURE_CHINA: "https://eventgrid.azure.cn",
+}
+
+# Clouds where Microsoft Fabric OneLake is available.
+FABRIC_ONELAKE_SUPPORTED_CLOUDS = frozenset({CLOUD_AZURE_PUBLIC})
+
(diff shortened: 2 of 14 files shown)


cheatsheet1999 (dismissed):

azext_edge/edge/util/oci_client.py
```diff
@@ -421,8 +423,15 @@
...
     @staticmethod
-    def _is_acr_registry(registry: str) -> bool:
+    def _is_acr_registry(registry: str, cmd=None) -> bool:
         """Check if the registry is an Azure Container Registry."""
+        if cmd is not None:
+            from .cloud_config import CloudConfig
+
+            try:
+                return registry.endswith(CloudConfig(cmd).acr_suffix)
+            except Exception:  # pragma: no cover - cloud ACR suffix not set
```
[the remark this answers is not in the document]
ketkimnaik: Fixed it.

cheatsheet1999 (approved):

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -59,6 +60,7 @@ def __init__(
         )
         self.keyvault_client = get_keyvault_client(
             subscription_id=self.default_subscription_id,
+            keyvault_scope=CloudConfig(self.cmd).keyvault_scope,
```
digimaun, line 63: The token scope is cloud aware now, but the vault base URL these calls hit still comes from `KEYVAULT_URL` in common.py, which is hardcoded to `.vault.azure.net`. In Gov and China we would be sending a correct token to the wrong host, so the OPC UA cert operations at lines 438, 723, 927 and 933 should fail. You already added `keyvault_dns_suffix` to CloudConfig for exactly this, so we should build the vault URL from the cloud suffix so the data plane matches the token scope.

azext_edge/edge/util/cloud_config.py
```diff
@@ -0,0 +1,133 @@
...
+        self._cloud = cmd.cli_ctx.cloud
+
+    @property
+    def name(self) -> str:
+        return self._cloud.name
+
+    # --- Values provided by the Azure CLI cloud framework ---
+
+    @property
+    def arm_endpoint(self) -> str:
```
digimaun, line 69: `arm_endpoint` is defined here but nothing in production consumes it. The preflight in `host.py:95` (`verify_cli_client_connections`) still probes the hardcoded public `ARM_ENDPOINT`, so in a sovereign cloud we would be checking connectivity to public ARM rather than the cloud the user is targeting, which may give a misleading pass or fail. We should wire this property into that preflight. If we are not going to use it, we should drop it.

azext_edge/edge/util/cloud_config.py
```diff
@@ -0,0 +1,133 @@
...
+        """Microsoft Graph token resource (no trailing slash)."""
+        return self._cloud.endpoints.microsoft_graph_resource_id.rstrip("/")
+
+    @property
+    def storage_suffix(self) -> str:
+        """Storage endpoint suffix (e.g. core.windows.net)."""
+        return self._cloud.suffixes.storage_endpoint
+
+    @property
+    def keyvault_dns_suffix(self) -> str:
```
digimaun, line 99: `keyvault_dns_suffix` is defined but nothing consumes it. It maps exactly to the High finding, the vault base URL in certs.py that is still hardcoded to `.vault.azure.net`. We should either wire it in there or drop it so the surface we expose matches what we actually use.

azext_edge/edge/util/cloud_config.py
```diff
@@ -0,0 +1,133 @@
...
+# Service Bus FQDN suffix per cloud. Not provided by the Azure CLI cloud framework.
+SERVICEBUS_SUFFIX_MAP = {
+    CLOUD_AZURE_PUBLIC: "servicebus.windows.net",
+    CLOUD_AZURE_US_GOVERNMENT: "servicebus.usgovcloudapi.net",
+    CLOUD_AZURE_CHINA: "servicebus.chinacloudapi.cn",
+}
+
+# Event Grid token audience per cloud. Not provided by the Azure CLI cloud framework.
+EVENTGRID_AUDIENCE_MAP = {
+    CLOUD_AZURE_PUBLIC: "https://eventgrid.azure.net",
```
digimaun, line 44: This public audience now duplicates `MGMT_ACTIONS_EG_AUDIENCE` in `common.py:442`. Production reads the audience from `CloudConfig.eventgrid_audience` now, but that constant still lingers and the mgmt actions tests still assert against it, so we have two definitions of the public audience that can drift apart. We should collapse to one, either drop the constant and have the tests assert against this map, or keep a single canonical definition.

azext_edge/edge/util/oci_client.py
```diff
@@ -421,8 +423,15 @@ def _get_acr_access_token(self, cmd, registry: str, repository: str) -> Optional
...
     @staticmethod
-    def _is_acr_registry(registry: str) -> bool:
+    def _is_acr_registry(registry: str, cmd=None) -> bool:
         """Check if the registry is an Azure Container Registry."""
+        if cmd is not None:
+            from .cloud_config import CloudConfig
+
+            try:
+                return registry.endswith(CloudConfig(cmd).acr_suffix)
+            except Exception as ex:  # pragma: no cover - cloud ACR suffix not set
```
digimaun, line 433: This new fallback catches bare `Exception` and logs at debug. The realistic failure from resolving `acr_suffix` is an `AttributeError` when the cloud context is missing, so we should catch that specifically to keep the fallback without masking anything unexpected, in line with the usual guidance to avoid broad excepts.

azext_edge/tests/edge/orchestration/test_mgmt_actions_unit.py
```diff
@@ -5726,3 +5726,43 @@ def test_namespace_not_found(self, mocked_cmd, mocked_responses: responses, mock
...
+
+        assert "not available in this cloud environment" in str(exc.value)
+
+    @pytest.mark.parametrize("cloud_name", ["AzureCloud", "AzureUSGovernment"])
+    def test_mgmt_actions_enable_not_blocked_in_supported_clouds(self, cloud_name):
+        provider = self._make_mgmt_actions(cloud_name)
+
+        # In supported clouds the cloud gate must NOT raise; the call proceeds past the gate
+        # and fails later for an unrelated reason (no mgmt client wired in this lightweight setup).
+        with pytest.raises(Exception) as exc:
```
digimaun, line 5761: This proves the supported cloud path by catching any `Exception` and asserting the gate message is absent, which relies on enable failing later for an unrelated reason. That is a little fragile. If enable ever fails before reaching the gate for a different reason, this test would still pass while no longer exercising the gate. We may want to assert it reaches a specific known point past the gate instead.

azext_edge/tests/helpers.py
```diff
@@ -441,3 +441,59 @@ def wait_for_expected_count(
...
+    fixture = CLOUD_CONFIG_FIXTURES[cloud_name]
+
+    class _Stub:
+        pass
+
+    cloud = _Stub()
+    cloud.name = cloud_name
+    cloud.endpoints = _Stub()
+    cloud.endpoints.resource_manager = fixture["resource_manager"]
+    cloud.endpoints.active_directory_resource_id = fixture["active_directory_resource_id"]
```
digimaun, line 490: `active_directory_resource_id` is set on this mock but CloudConfig never reads it, and `build_mock_cmd_for_cloud` is only used by the new cloud readiness tests, so this field is unused here and could be dropped as noise. The matching value correction in conftest.py is fine to keep.

azext_edge/edge/util/queryable.py
```diff
@@ -66,19 +62,24 @@ def query(self, query: str, first: bool = False) -> Optional[Union[dict, List[di
     def get_resource_group(self, name: str) -> dict:
         return self.resource_client.resource_groups.get(resource_group_name=name)
 
-    def get_sp_id(self, app_id: str, token_resource: str = "https://graph.microsoft.com", **kwargs) -> Optional[str]:
+    def get_sp_id(self, app_id: str, token_resource: Optional[str] = None, **kwargs) -> Optional[str]:
         """
         Attempts to fetch the service principal Id by app Id from the Microsoft Graph API.
         """
         from azure.cli.core.util import send_raw_request
+        from .cloud_config import CloudConfig
```
digimaun, line 70: CloudConfig is imported inline here and in oci_client.py and mgmt_actions.py, but at module top in certs.py, dataflows.py and instances.py. cloud_config.py imports nothing from azext_edge so there is no circular risk, so we may want to standardize on a top level import for consistency.

PR #893, opened 2026-07-02
fix(cloud): wire cloud-aware endpoints and dedupe EG audience

ketkimnaik:
Review Comment (from #892) Fixes

- Key Vault URL hardcoded to public suffix: Now build the vault URL from the cloud's Key Vault DNS suffix so it matches the token scope in Gov/China.
- ARM endpoint defined but preflight probed public ARM: Preflight now checks connectivity against the active cloud's ARM endpoint.
- Key Vault DNS suffix defined but unused: Now consumed when building the vault URL.
- Duplicate public Event Grid audience: Collapsed to a single canonical definition.
- Broad exception catch on ACR suffix resolution: Narrowed to the specific expected error so nothing unexpected is masked.
- Fragile cloud-gate test: Now asserts it reaches a specific point past the gate instead of catching any exception.
- Unused field on the test cloud mock: Removed as noise.
- Inconsistent cloud config import style: Standardized on top-level imports.

---

### General Guidelines

Intent for Production

Basic expectations


azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
```diff
@@ -71,6 +73,10 @@ def __init__(
         self.location = instance["location"]
         self.opcua_mode = instance.get("properties", {}).get("features", {}).get("opcua", {}).get("mode")
 
+    def _build_vault_url(self, keyvault_name: str) -> str:
+        """Build the Key Vault data-plane base URL for the active cloud."""
+        return f"https://{keyvault_name}{self._keyvault_dns_suffix}/"
```
digimaun, line 78: The vault URL resolves per cloud now, but the only test coverage is the public suffix since the opcua cert tests all run through the shared public `mocked_cmd`. A regression back to a hardcoded public host would not be caught in Gov or China, which is the case this change exists for. #892 already added cloud parametrized tests for the dataflow hosts and CloudConfig resolution, so we should add a small parametrized test asserting `_build_vault_url` returns the `.vault.usgovcloudapi.net` and `.vault.azure.cn` hosts to close the loop.

azext_edge/tests/edge/orchestration/test_mgmt_actions_unit.py
```diff
@@ -5756,13 +5758,14 @@ def test_mgmt_actions_enable_blocked_in_china(self):
...
+        # because no mgmt client is wired in this lightweight setup.
+        with pytest.raises(AttributeError) as exc:
             provider.enable(
                 name=generate_random_string(),
                 resource_group_name=generate_random_string(),
                 eg_resource_id=generate_random_string(),
             )
 
-        assert "not available in this cloud environment" not in str(exc.value)
+        assert "iotops_mgmt_client" in str(exc.value)
```
digimaun, line 5777: One tradeoff to flag with this assertion. It couples the test to the internal attribute name `iotops_mgmt_client` and relies on `WorkflowDisplay` and the semver import not raising first, so if the enable flow is reordered later this may fail in a confusing way. It does satisfy the ask to reach a specific point past the gate. We may want a brief comment noting the instance lookup is the intended first client interaction, so the coupling is obvious to the next reader.

azext_edge/edge/providers/orchestration/resources/connector/opcua/certs.py
ketkimnaik: Added Test

azext_edge/tests/edge/orchestration/test_mgmt_actions_unit.py
ketkimnaik: Added brief comment

digimaun (approved):

PR #897, opened 2026-07-13
feat: generalize ns asset mgmt-group and mgmt-action commands

ketkimnaik:
- Adds connector-agnostic `az iot ops ns asset mgmt-group` and `az iot ops ns asset mgmt-action` command groups that automatically detect the connector type from the asset's device endpoint, so users no longer need to specify it in the command path.
- Provides the full lifecycle across both groups — add, update, show, list, remove, and export/import round-trips for management groups and their actions — with `--show-template` for discovering the connector-specific configuration schema.

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/params.py
+++ b/azext_edge/edge/params.py
@@ -49,7 +49,7 @@ from .providers.orchestration.common import (
     TlsKeyAlgo,
     TlsKeyRotation,
 )
-from .providers.adr.common import EndpointTemplateMode, FileType
+from .providers.adr.common import ActionType, EndpointTemplateMode, FileType
 
 
 def load_iotops_arguments(self, _):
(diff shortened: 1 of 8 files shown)


azext_edge/edge/providers/adr/namespace_assets.py
```diff
@@ -4114,6 +4114,459 @@ def import_streams(
...
+                instance_name=instance_name,
+                instance_resource_group=instance_resource_group,
+                is_action=False,
+            )
+            if "managementGroupConfiguration" in config_result:
+                group["managementGroupConfiguration"] = config_result["managementGroupConfiguration"]
+
+        if default_topic == "":
+            group.pop("defaultTopic", None)
+        elif default_topic:
```
cheatsheet1999, line 4448: This is probably by design, but we can clear defaultTopic with  `--default-topic ""` , but there's no way to clear `dataSource / typeRef` (since they only set when is not None, right?).

azext_edge/edge/params.py
```diff
@@ -2490,6 +2490,206 @@ def load_iotops_arguments(self, _):
...
+        )
+        context.argument(
+            "default_topic",
+            options_list=["--default-topic", "--dt"],
+            help="Default topic for management group actions.",
+        )
+        context.argument(
+            "default_timeout",
+            options_list=["--default-timeout", "--dto"],
+            help="Default timeout in seconds for management group actions. Minimum: 0",
```
cheatsheet1999, line 2517: Did we check a negative number is rejected by the server? Nit, non-blocking 🙂

azext_edge/edge/providers/adr/namespace_assets.py
ketkimnaik: By design — this matches the existing connector-specific mgmt-group update, where only defaultTopic is clearable via "" and dataSource/typeRef are set only when a value is provided. Keeping the generalized command consistent with that behavior.

azext_edge/edge/params.py
ketkimnaik: Verified against the live server — a negative value is rejected with ManagementGroupDefaultTimeoutInSecondsMinValueValidationError ("cannot be less than 0 seconds") and the resource isn't created, so validation is server-authoritative here (consistent with the existing mgmt commands).

cheatsheet1999 (approved):

PR #899, opened 2026-07-16
fix: match FIC issuer to token iss

cheatsheet1999:
## Summary

SecretSync used the OIDC issuer reported by ARM when creating a federated identity credential (FIC). In some configurations the ARM issuer ends with `/` while the service account token `iss` does not. Entra requires an exact match, so secret retrieval fails with `AADSTS700211` (HTTP 401).

## Fix
Resolve the canonical issuer from the cluster's **public OIDC discovery document** (`{issuer}/.well-known/openid-configuration`). By the OIDC contract, the discovery `issuer` is the same value the token carries as `iss` - and the same document Entra fetches unauthenticated to validate the FIC.

- Resolve via a plain unauthenticated HTTPS GET - **no kubeconfig, cluster RBAC, or service account token**. This also works from an ARM-only / CI context and in the clone flow before the target cluster exists.
- Use the discovery issuer **verbatim** in the FIC; accept it only when it is exact or differs from the ARM value by exactly one trailing slash.
- Probe the verbatim URL and, for trailing-slash ARM issuers, also the one-slash-removed URL, use whichever returns a matching issuer (the single-slash form is what succeeds against Azure's OIDC endpoint).
- Warn and keep the ARM value when discovery cannot verify - no new hard-fail on an ARM-only command.
- Harden the request: HTTPS-only, TLS verification, short timeout, bounded response body, and no redirect following (the URL originates from ARM data).
- Use the resolved issuer consistently for FIC lookup, naming, and creation, and apply the same handling to the clone flow.

## Some clarifications...

**Q: Why read the discovery document instead of minting a service account token?**
A: The discovery `issuer` equals the token `iss` by contract, so it yields the same canonical value without requiring cluster access, kubeconfig, RBAC (`serviceaccounts/token`), or the service account to exist. It also works from CI / ARM-only contexts and in clone (before the cluster is deployed).

**Q: Why not always remove the trailing slash?**
A: This is not a failure that happens every time. Some clusters include `/` in their issuer. Normalizing unconditionally could break working configurations, so the canonical discovery value is used verbatim.

**Q: Does this change the common case?**
A: No. When ARM and the discovery issuer already match, the same issuer is used with no warning.

**Q: Why does the FIC lookup remain exact?**
A: Entra treats issuers with and without `/` as different values. The resolved issuer is used for both lookup and creation.

**Q: Does this update existing FICs?**
A: No. Repairing already-created FICs is outside this change's scope (tracked as a follow-up).

### Before:
token iss and ARM issuerUrl mismatch:

iss:
image

ARM issueUrl (and FIC):
image

Note the trailing slash after `fc`

Result: 401
image

### After
token ISS and ARM IssuerUrl remains unchanged, but FIC is normalized
Screenshot 2026-07-15 190404
No trailing slash

Result:
image

### DIFF
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -116,7 +116,12 @@ def _get_service_account_issuer(namespace: str, service_account_name: str) -> Op
             response,
             client.AuthenticationV1TokenRequest,
         )
-    return decode_access_token(token_response.status.token).get("iss")
+
+    try:
+        issuer = decode_access_token(token_response.status.token).get("iss")
+    except (AttributeError, IndexError, ValueError):
+        return None
+    return issuer if isinstance(issuer, str) else None
 
 
 def resolve_oidc_issuer(arm_issuer: str, namespace: str, service_account_name: str) -> str:
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -4,12 +4,14 @@
 # Licensed under the MIT License. See License file in the project root for license information.
 # ----------------------------------------------------------------------------------------------
 
+import json
 import re
 from contextlib import nullcontext
 from typing import Dict, Iterable, List, Optional
+from urllib.parse import urlsplit
 
+import requests
 import yaml
-from azure.cli.core.auth.util import decode_access_token
 from azure.cli.core.azclierror import (
     AzureResponseError,
     InvalidArgumentValueError,
@@ -17,14 +19,9 @@ from azure.cli.core.azclierror import (
 )
 from azure.core.exceptions import HttpResponseError, ResourceNotFoundError
 from knack.log import get_logger
-from kubernetes import client
-from kubernetes.client.exceptions import ApiException
-from kubernetes.config.config_exception import ConfigException
 from rich import print
 from rich.console import Console
-from urllib3.exceptions import HTTPError
 
-from ...base import load_config_context
 from ....util.az_client import (
     ResourceIdContainer,
     get_iotops_mgmt_client,
@@ -74,6 +71,9 @@ KEYVAULT_ROLE_ID_SECRETS_USER = "4633458b-17de-408a-b874-0445c86b69e6"
 KEYVAULT_ROLE_ID_READER = "21090545-7ca7-4776-b22c-e363652d74d2"
 
 COMPAT_FEAT_KEY_SET = {"opcua.mode"}
+OIDC_DISCOVERY_PATH = "/.well-known/openid-configuration"
+OIDC_DISCOVERY_TIMEOUT_SECONDS = 5
+OIDC_DISCOVERY_MAX_RESPONSE_BYTES = 64 * 1024  # cap the response size to 64KB
 
 
 def get_user_msg_warn_ra(prefix: str, principal_id: str, scope: str) -> str:
@@ -98,46 +98,76 @@ def get_cred_subject(namespace: str, service_account_name: str):
     return f"system:serviceaccount:{namespace}:{service_account_name}"
 
 
-def _get_service_account_issuer(namespace: str, service_account_name: str) -> Optional[str]:
-    token_request = client.AuthenticationV1TokenRequest(
-        spec=client.V1TokenRequestSpec(
-            audiences=["api://AzureADTokenExchange"],
-        )
-    )
-    core_v1_api = client.CoreV1Api()
-    with core_v1_api.create_namespaced_service_account_token(
-        name=service_account_name,
-        namespace=namespace,
-        body=token_request,
-        _preload_content=False,
-        _request_timeout=5,
-    ) as response:
-        token_response = core_v1_api.api_client.deserialize(
-            response,
-            client.AuthenticationV1TokenRequest,
-        )
+def oidc_issuers_match(first: str, second: str) -> bool:
+    if first == second:
+        return True
+    return first == f"{second}/" or second == f"{first}/"
 
+
+def _is_https_url(url: str) -> bool:
     try:
-        issuer = decode_access_token(token_response.status.token).get("iss")
-    except (AttributeError, IndexError, ValueError):
+        parsed = urlsplit(url)
+    except (AttributeError, ValueError):
+        return False
+    return parsed.scheme.lower() == "https" and parsed.hostname is not None
+
+
+def _read_public_discovery_issuer(discovery_url: str) -> Optional[str]:
+    try:
+        with requests.get(
+            discovery_url,
+            allow_redirects=False,
+            stream=True,
+            timeout=OIDC_DISCOVERY_TIMEOUT_SECONDS,
+            verify=True,
+        ) as response:
+            if response.status_code != 200:
+                return None
+            # Bound the decoded body since the URL comes from ARM data.
+            content = bytearray()
+            for chunk in response.iter_content(chunk_size=OIDC_DISCOVERY_MAX_RESPONSE_BYTES):
+                content.extend(chunk)
+                if len(content) > OIDC_DISCOVERY_MAX_RESPONSE_BYTES:
+                    return None
+            issuer = json.loads(content).get("issuer")
+    except (requests.RequestException, ValueError, AttributeError, TypeError):
         return None
-    return issuer if isinstance(issuer, str) else None
+    return issuer if isinstance(issuer, str) and issuer else None
 
 
-def resolve_oidc_issuer(arm_issuer: str, namespace: str, service_account_name: str) -> str:
-    try:
-        load_config_context()
-        kubernetes_issuer = _get_service_account_issuer(
-            namespace=namespace,
-            service_account_name=service_account_name,
-        )
-    except (ApiException, ConfigException, HTTPError, OSError, yaml.YAMLError):
-        return arm_issuer
+def _get_public_discovery_issuer(arm_issuer: str) -> Optional[str]:
+    if not _is_https_url(arm_issuer):
+        return None
+    discovery_urls = [f"{arm_issuer}{OIDC_DISCOVERY_PATH}"]
+    if arm_issuer.endswith("/"):
+        discovery_urls.append(f"{arm_issuer[:-1]}{OIDC_DISCOVERY_PATH}")
 
-    if not kubernetes_issuer or kubernetes_issuer.rstrip("/") != arm_issuer.rstrip("/"):
-        return arm_issuer
+    for discovery_url in discovery_urls:
+        issuer = _read_public_discovery_issuer(discovery_url)
+        if issuer and oidc_issuers_match(arm_issuer, issuer):
+            return issuer
+    return None
 
-    return kubernetes_issuer
+
+def resolve_oidc_issuer(arm_issuer: str) -> str:
+    discovery_issuer = _get_public_discovery_issuer(arm_issuer)
+
+    if discovery_issuer:
+        if discovery_issuer != arm_issuer:
+            logger.warning(
+                "The OIDC issuer reported by ARM '%s' differs from the cluster issuer '%s' by a trailing slash. "
+                "The federated identity credential will use the cluster issuer.",
+                arm_issuer,
+                discovery_issuer,
+            )
+        return discovery_issuer
+
+    logger.warning(
+        "Could not verify the OIDC issuer reported by ARM '%s'. The ARM value will be used for the federated "
+        "identity credential, but workload identity token exchange may fail if it differs from the cluster issuer.",
+        arm_issuer,
+    )
+    return arm_issuer
 
 
 def get_enable_syntax(instance_name: str, resource_group_name: str) -> str:
@@ -441,11 +471,7 @@ class Instances(Queryable):
                 )
                 return
 
-            oidc_issuer = resolve_oidc_issuer(
-                arm_issuer=oidc_issuer,
-                namespace=namespace,
-                service_account_name=SERVICE_ACCOUNT_SECRETSYNC,
-            )
+            oidc_issuer = resolve_oidc_issuer(arm_issuer=oidc_issuer)
             if not federated_credential_name:
                 federated_credential_name = get_fc_name(
                     cluster_name=cluster_resource["name"],
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -75,6 +75,7 @@ COMPAT_FEAT_KEY_SET = {"opcua.mode"}
 OIDC_DISCOVERY_PATH = "/.well-known/openid-configuration"
 OIDC_DISCOVERY_TIMEOUT_SECONDS = 5
 OIDC_DISCOVERY_MAX_RESPONSE_BYTES = 64 * 1024  # cap the response size to 64KB
+OIDC_DISCOVERY_CHUNK_SIZE_BYTES = 8 * 1024
 
 
 def get_user_msg_warn_ra(prefix: str, principal_id: str, scope: str) -> str:
@@ -129,7 +130,7 @@ def _read_public_discovery_issuer(discovery_url: str, deadline: float) -> Option
                 return None
             # Bound the decoded body since the URL comes from ARM data.
             content = bytearray()
-            for chunk in response.iter_content(chunk_size=OIDC_DISCOVERY_MAX_RESPONSE_BYTES):
+            for chunk in response.iter_content(chunk_size=OIDC_DISCOVERY_CHUNK_SIZE_BYTES):
                 if time.monotonic() > deadline:
                     return None
                 content.extend(chunk)
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -6,6 +6,7 @@
 
 import json
 import re
+import time
 from contextlib import nullcontext
 from typing import Dict, Iterable, List, Optional
 from urllib.parse import urlsplit
@@ -112,13 +113,16 @@ def _is_https_url(url: str) -> bool:
     return parsed.scheme.lower() == "https" and parsed.hostname is not None
 
 
-def _read_public_discovery_issuer(discovery_url: str) -> Optional[str]:
+def _read_public_discovery_issuer(discovery_url: str, deadline: float) -> Optional[str]:
+    remaining = deadline - time.monotonic()
+    if remaining <= 0:
+        return None
     try:
         with requests.get(
             discovery_url,
             allow_redirects=False,
             stream=True,
-            timeout=OIDC_DISCOVERY_TIMEOUT_SECONDS,
+            timeout=remaining,
             verify=True,
         ) as response:
             if response.status_code != 200:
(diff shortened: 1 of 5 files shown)


azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,43 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+        name=service_account_name,
+        namespace=namespace,
+        body=token_request,
+        _preload_content=False,
+        _request_timeout=5,
+    ) as response:
+        token_response = core_v1_api.api_client.deserialize(
+            response,
+            client.AuthenticationV1TokenRequest,
+        )
```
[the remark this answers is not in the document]
cheatsheet1999: @copilot The installed runtime is Kubernetes 36.0.2, the current request constructs and serializes successfully. In addition, a freshly downloaded kubernetes-36.0.2 PyPI wheel exports `AuthenticationV1TokenRequest` and does not contain/export `V1TokenRequest` .

Verify again and check if the above review is valid.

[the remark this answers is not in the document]
cheatsheet1999: Invalid, resolving.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,43 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+
+
+def resolve_oidc_issuer(arm_issuer: str, namespace: str, service_account_name: str) -> str:
+    try:
+        load_config_context()
+        kubernetes_issuer = _get_service_account_issuer(
+            namespace=namespace,
+            service_account_name=service_account_name,
+        )
+    except (ApiException, ConfigException, HTTPError, OSError, yaml.YAMLError):
```
ketkimnaik: `_get_service_account_issuer` ends with `decode_access_token(token_response.status.token).get("iss")`, but `resolve_oidc_issuer` only catches (ApiException, ConfigException, HTTPError, OSError, yaml.YAMLError). A malformed/empty token would raise a decode error (or AttributeError if status/token is None), which escapes the try and aborts the whole enable/clone flow. Since this is meant to be best-effort with a fallback, consider broadening the caught exceptions or wrapping the decode.

cheatsheet1999: Thanks. I initially assumed a successful TokenRequest would always contain a valid JWT, so I only handled Kubernetes and configuration errors. 

I agree missing or malformed token data should also fall back to the ARM issuer. I’ve updated the decode path accordingly.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,48 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+        )
+
+    try:
+        issuer = decode_access_token(token_response.status.token).get("iss")
+    except (AttributeError, IndexError, ValueError):
+        return None
+    return issuer if isinstance(issuer, str) else None
+
+
+def resolve_oidc_issuer(arm_issuer: str, namespace: str, service_account_name: str) -> str:
```
digimaun: I want to float a different mechanism, because I think it removes the cluster dependency entirely and resolves most of the concerns at the root.

The value Entra compares against is the issuer published at the OIDC discovery doc, `{issuer}/.well-known/openid-configuration`, which by the discovery contract must equal the token `iss`. That is the same document Entra fetches unauthenticated to validate the FIC, and for a conformant Kubernetes issuer the discovery issuer and the token `iss` come from the same configured value. So we can read the canonical issuer with a plain HTTPS GET, no kubeconfig, no cluster RBAC, no `aio-ssc-sa`, and it works from CI or an ARM only context.

**Resolution**
- Fetch the discovery doc for the ARM issuer. Mind the URL, if the ARM value ends in `/` then `{arm_issuer}/.well-known/...` becomes `//.well-known`, so probe the verbatim form and the one slash removed form.
- Use the discovery issuer verbatim in the FIC. Do not `rstrip` and store a normalized value. The broad `rstrip("/")` here also treats multiple slashes as equal.
- Accept the discovery issuer only when it is exact or differs from ARM by exactly one terminal slash. Harden the GET, https only, TLS verify, short timeout, bounded body, limit redirects, since the URL comes from ARM data.

**How the user becomes aware**
- Exact match, use it, no noise.
- One slash difference, use the canonical value and warn that ARM and the cluster issuer differed and we corrected the FIC. That warning at enable time is how the user finds out, they no longer have to discover it through a downstream 401.
- Discovery unreachable or unverifiable, warn and keep the ARM value and mention the risk. I would not hard fail here, that would add a failure mode to a command that is ARM only today.

**Cluster fallback, gated and non intrusive**
- Default is discovery only, we never touch kubeconfig.
- If discovery is blocked from the CLI host, the user can opt in to a cluster read with `--context`. Only an explicitly passed `--context` triggers cluster access, never the ambient current-context. Worth noting `--context` is optional elsewhere and just selects a context, here its presence also means consent to cluster access, so a user opts in with something like `--context default`.
- Even the cluster fallback can read `/.well-known/openid-configuration` from the cluster API rather than minting a token, which avoids the `serviceaccounts/token` RBAC and the SA existence dependency.

**Repair for users already affected**
- enable returns before resolution when secretsync resources already exist (`instances.py:436`), so someone who already hit the 401 cannot fix the bad FIC by re-running. If we resolve before that early return and update the existing slash equivalent FIC by name instead of creating a new one, re-running enable becomes the repair path. clone benefits too, since discovery works before the target is deployed whereas the token lookup runs before the target could even have the SA.

If we would rather keep the token approach for this PR and defer the redesign, the two must fixes are to honor an explicit `--context` instead of current-context, and to log the fallback so it is not invisible. But I think discovery is the better default and worth doing here. Was the token route chosen with a specific self hosted or private issuer scenario in mind? If the issuer must be reachable by Entra anyway, discovery from the CLI should cover the managed case and `--context` covers the CLI egress blocked case.

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
```diff
@@ -77,6 +78,15 @@ def mocked_get_tenant_id(mocker):
     )
 
 
+@pytest.fixture(autouse=True)
```
digimaun, line 80: This autouse fixture patches `resolve_oidc_issuer` to identity for the whole enable suite, and the clone suite does the same, so `_get_service_account_issuer` is never actually executed in tests. That is the root of the last two rounds here. The `AuthenticationV1TokenRequest` class name could not be validated by CI because that line never runs, and the decode fallback that was just added has no test either. Whichever resolution mechanism we land on, a focused test that exercises the real resolver against a mocked transport, covering a matching issuer, a one slash difference, and an unreachable or malformed response, would lock down the path in one place.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: Thanks, this direction makes sense. The token route was chosen because it was the most direct way to reproduce the issue and obtain the exact token `iss` , not because of a specific private or self-hosted issuer scenario. I’m aligned with using discovery as the default and an explicitly supplied  --context  for cluster fallback. 

One scope check: would you like this PR to include the full repair path for existing slash-equivalent FICs and wire `--context` into both SecretSync enable and clone, or would you prefer discovery plus focused tests here and track the repair path separately? My reading is that you prefer all of it in this PR.

digimaun: For this PR I'd scope it to the discovery resolver in both enable and clone, with tests. We take the discovered issuer verbatim, accept it only on an exact or single trailing slash match, warn on a correction, and keep ARM when we can't verify. That removes the token path, and with it kubeconfig, --context, RBAC, and the ambient context concern.

Repair for existing FICs I'd keep as a fast follow up since it changes the early return and deserves its own review. Same for the --context fallback, it can wait until an egress blocked user actually needs it.

cheatsheet1999: Got it, thanks.

cheatsheet1999:
@digimaun Thanks for the detailed guidance. I’ve refactored this to use the discovery resolver we agreed on. Mapping the changes to your points:

1. Discovery in both flows, resolve_oidc_issuer(arm_issuer=...) now runs in both SecretSync enable (instances.py) and clone (clone.py). The token-based path has been removed.
2. Verbatim issuer - The issuer returned by discovery is used as-is. No rstrip or normalized value is stored.
3. Exact match or one trailing slash only - `oidc_issuers_match()` returns true only when the values are identical or differ by exactly one trailing slash. Values with multiple trailing slashes or unrelated values are rejected.
4. URL handling - I first probe the verbatim discovery URL. If the ARM issuer ends with a trailing slash, I also probe the URL with that slash removed, then use whichever endpoint returns a matching issuer. The single-slash form succeeds against Azure’s OIDC endpoint, while the verbatim //.well-known form returns a 404.
5. Hardened GET - The request is restricted to HTTPS, uses `verify=True`, has a five-second timeout, limits the response body to 64 KiB, and sets `allow_redirects=False`.
6. User awareness - An exact match is silent. A one-slash correction produces a warning that the ARM and cluster issuers differed and that the FIC was corrected. If discovery is unreachable or cannot be verified, we warn the user and retain the ARM value, so this does not introduce a new hard failure.
7. Deferred based on your guidance - Repairing an existing FIC, which would require changing the early-return behavior, and adding an opt-in --context cluster fallback will be tracked as separate follow-up items.

Could you take another look when you have a moment?

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
cheatsheet1999: Added `test_oidc_issuer_unit.py` that exercises the real resolver against a mocked transport (matching, one-slash, unreachable, malformed, + edge cases). Resolving.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,78 @@ def get_cred_subject(namespace: str, service_account_name: str):
     return f"system:serviceaccount:{namespace}:{service_account_name}"
 
 
+def oidc_issuers_match(first: str, second: str) -> bool:
+    if first == second:
+        return True
+    return first == f"{second}/" or second == f"{first}/"
+
```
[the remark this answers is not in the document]
cheatsheet1999: It’s true that `oidc_issuers_match("…/x//", "…/x/")` returns `True`, but this cannot result in an incorrect FIC.

The match is only used as a trust gate. When the issuers match, we store the discovery document’s `issuer` verbatim, never the ARM value. Under the OIDC discovery contract, that published issuer is the value Entra validates against the token’s `iss` claim, so leniency in the comparison cannot change the value stored in the FIC.

The `…//` form is also not expected to be reachable, since neither ARM nor a conformant discovery document should emit an issuer with two trailing slashes. 

I’m leaving this as-is for now, but I’m happy to add a `not endswith("/")` guard if other folks would prefer the function to reject that case explicitly. Resolving.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,78 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+    return parsed.scheme.lower() == "https" and parsed.hostname is not None
+
+
+def _read_public_discovery_issuer(discovery_url: str) -> Optional[str]:
+    try:
+        with requests.get(
+            discovery_url,
+            allow_redirects=False,
+            stream=True,
+            timeout=OIDC_DISCOVERY_TIMEOUT_SECONDS,
```
digimaun: The `timeout` here is a scalar, so requests applies it to the connection and to inactivity between reads, not as a total deadline. With `stream=True` and the `iter_content` loop, a server that sends a byte just under the interval keeps the read alive well past 5 seconds, and the two probes run in sequence. Benign slowness is handled since a real stall trips the timeout and we fall back to ARM. A slow drip endpoint though could stall enable and clone for a long time. Since we are treating this URL as untrusted ARM data, could we add a wall clock budget across both probes and bail once it is exceeded?

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,78 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+            if response.status_code != 200:
+                return None
+            # Bound the decoded body since the URL comes from ARM data.
+            content = bytearray()
+            for chunk in response.iter_content(chunk_size=OIDC_DISCOVERY_MAX_RESPONSE_BYTES):
+                content.extend(chunk)
+                if len(content) > OIDC_DISCOVERY_MAX_RESPONSE_BYTES:
+                    return None
+            issuer = json.loads(content).get("issuer")
+    except (requests.RequestException, ValueError, AttributeError, TypeError):
```
digimaun: This catch is a bit narrower than the best effort contract the function promises. `json.loads` on a deeply nested body raises `RecursionError`, which is a `RuntimeError` and not in this tuple, so it escapes through `resolve_oidc_issuer` into enable and clone and aborts the flow. That is the same shape as the decode error from the earlier round. Since anything here should fall back to ARM, could we catch `Exception` and log at debug, or add the missing types, so a malformed discovery doc cannot take down the command?

azext_edge/tests/edge/orchestration/resources/test_oidc_issuer_unit.py
```diff
@@ -0,0 +1,153 @@
...
+    assert resolve_oidc_issuer(arm_issuer) == arm_issuer
+    warning.assert_called_once()
+
+
+def test_resolve_oidc_issuer_rejects_oversized_discovery(mocked_responses: responses, mocker):
+    arm_issuer = "https://issuer.example/cluster"
+    mocked_responses.add(
+        method=responses.GET,
+        url=f"{arm_issuer}{OIDC_DISCOVERY_PATH}",
+        body=b"x" * (OIDC_DISCOVERY_MAX_RESPONSE_BYTES + 1),
```
digimaun: This body is a run of `x`, which is not valid JSON, so the fallback here comes from the parse error rather than the size guard. The test passes even if the 64KB cap is removed. Could we use oversized valid JSON instead, something like a matching issuer plus padding that pushes it past the cap, so the test actually fails if the guard goes away?

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -92,6 +98,78 @@ def get_cred_subject(namespace: str, service_account_name: str):
...
+            logger.warning(
+                "The OIDC issuer reported by ARM '%s' differs from the cluster issuer '%s' by a trailing slash. "
+                "The federated identity credential will use the cluster issuer.",
+                arm_issuer,
+                discovery_issuer,
+            )
+        return discovery_issuer
+
+    logger.warning(
+        "Could not verify the OIDC issuer reported by ARM '%s'. The ARM value will be used for the federated "
```
digimaun: Keeping this warning makes sense since we are about to write auth state whose issuer we could not confirm. One wording thought. The common way to hit this is a CLI host that cannot reach the public issuer, where the ARM value is often correct and federation still works because Entra reaches the issuer rather than the CLI. Could we word it so the user understands the issuer could not be reached from this host and the ARM value may well be fine, rather than reading it as a broken credential?

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: I see, thanks for pointing this out.

I’m thinking of using a `time.monotonic()` deadline as a single five-second budget across both probes. The deadline would be:

* Checked before each probe
* Passed as the request `timeout`
* Rechecked in the read loop so that a slow-drip response also triggers it

I’ll also fetch the normalized `/.well-known` URL first for trailing-slash issuers. This ensures that the probe we expect to succeed runs first and is not starved by the `//` probe that we know most likely returns a 404.

We would fall back to ARM only if the actual endpoint is unreachable within the shared budget.

I’m open to using a different total timeout if you prefer.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: I'll update the handler to catch Exception, log the failure at debug level, and return `None` so it falls back to the ARM value. Thank you for bringing this up.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: That makes sense, thanks for clarifying. I’ll keep the warning, but update the wording so it is clear that the issuer could not be reached from the current CLI host, rather than implying that the federated credential is invalid.

I'm moving forward with below message:

“Could not verify the OIDC issuer ‘%s’ from this host. The ARM-reported value will be used for the federated identity credential. This may still work if Microsoft Entra can reach the issuer.”

Let me know if you’d prefer different wording or emphasis.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: In the latest commit, I implemented a shared monotonic deadline with `requests`, but it does not fully address the slow-drip case because `iter_content()` can remain blocked beyond the deadline.

A strict wall-clock timeout appears to require using something like `aiohttp.ClientTimeout(total=remaining)`. That would enforce the shared five-second budget across both probes, including response-body reads, but it would also introduce a new dependency and require some additional unit-test changes.

Before I make the larger change, would you prefer that approach, or would the simpler best-effort deadline with `requests` be sufficient for this case?

azext_edge/tests/edge/orchestration/resources/test_oidc_issuer_unit.py
cheatsheet1999: You're right. I switched it to valid oversized JSON. Thank you.

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: Let's stay with requests and skip aiohttp. A best effort resolve time probe does not warrant a new runtime dependency in the wheel, and the safe fallback to the ARM value with a warning already covers a failed probe. You are right that requests gives no hard total deadline here. The timeout is per read inactivity, and iter_content with a large chunk can stay inside one read past the deadline on a slow drip, so this stays best effort rather than a guaranteed bound. If you want to narrow that window without a dependency, streaming with a small chunk size makes the deadline check run on each read and bounds the body drip more tightly. I am fine either way given the fallback, so treat that as optional.

azext_edge/edge/providers/orchestration/resources/instances.py
digimaun: This wording works well.

cheatsheet1999: Resolving. Thank you.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: Sounds good. I’ll keep the current `requests` implementation and treat the shared deadline as a best-effort bound. Thanks for confirming.

digimaun: chunk_size is currently the same value as the 64KB cap, so content can grow to roughly twice the cap before the size check rejects it. A smaller chunk size would tighten the size guard and the deadline check together, so it is the same optional change.

digimaun (dismissed):

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: Actually, thanks for pointing this out. I think this is a good safeguard to have. 

I’ll use a smaller 8 KB chunk size so the deadline is checked more frequently, and the buffer can only grow slightly beyond the 64 KB cap before the request is rejected.

cheatsheet1999: Updated the code, now it grows to at most roughly 72KB.

digimaun (approved):

PR #902, opened 2026-07-16
fix(check): report cluster access-denied per resource instead of false negatives

ketkimnaik:
- `az iot ops check` now detects Kubernetes 401/403 responses when reading Azure IoT Operations resources and reports a precise per-resource "Access denied (HTTP 403)" (with a permissions footer in the summary), instead of misleading "Unable to fetch … in any namespace" false negatives.
- Handling is per resource, so partial-access principals still see valid results for resources they can read, while only the forbidden ones are flagged — no full-command abort.

Before change overall output:

image

It does not show clear message and fails the check.

After change applied:
1. For overall command:
image

2. For per service:
image

It shows detailed error.

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/providers/base.py
+++ b/azext_edge/edge/providers/base.py
@@ -46,6 +46,15 @@ def load_config_context(context_name: Optional[str] = None):
     DEFAULT_NAMESPACE = current_config.get("namespace") or "azure-iot-operations"
 
 
+class ClusterAccessDeniedError(Exception):
+    """Raised when a cluster API call is rejected with HTTP 401/403 due to missing permissions."""
+
+    def __init__(self, status: Optional[int], resource: Optional[str] = None):
+        self.status = int(status or 403)
+        self.resource = resource
+        super().__init__(f"Access denied (HTTP {self.status})" + (f" for '{resource}'" if resource else ""))
+
+
 _namespaced_service_cache: dict = {}
 
 
@@ -110,7 +119,12 @@ _custom_object_cache: dict = {}
 
 
 def get_custom_objects(
-    group: str, version: str, plural: str, namespace: Optional[str] = None, use_cache: bool = True
+    group: str,
+    version: str,
+    plural: str,
+    namespace: Optional[str] = None,
+    use_cache: bool = True,
+    raise_on_access_error: bool = False,
 ) -> Union[dict, None]:
     target_resource_key = (group, version, plural, namespace)
     if use_cache:
@@ -128,6 +142,8 @@ def get_custom_objects(
         _custom_object_cache[target_resource_key] = f(**kwargs)
     except ApiException as ae:
         logger.debug(str(ae))
+        if raise_on_access_error and int(ae.status or 0) in (401, 403):
+            raise ClusterAccessDeniedError(status=ae.status, resource=plural)
     else:
         return _custom_object_cache[target_resource_key]
 
--- a/azext_edge/edge/providers/base.py
+++ b/azext_edge/edge/providers/base.py
@@ -48,7 +48,7 @@ def load_config_context(context_name: Optional[str] = None):
 
 
 class ClusterAccessDeniedError(Exception):
-    """Raised when a cluster API call is rejected with HTTP 401/403 due to missing permissions."""
+    """Raised when a cluster API call is rejected with HTTP 401/403 (authentication/authorization failure)."""
 
     def __init__(self, status: Optional[int], resource: Optional[str] = None):
         self.status = int(status or 403)
@@ -73,7 +73,7 @@ def reraise_cluster_access_errors() -> Iterator[None]:
         _reraise_access_errors.reset(token)
 
 
-def _reraise_if_access_denied(ae: ApiException, resource: Optional[str] = None) -> None:
+def reraise_if_access_denied(ae: ApiException, resource: Optional[str] = None) -> None:
     if _reraise_access_errors.get() and int(getattr(ae, "status", 0) or 0) in (401, 403):
         raise ClusterAccessDeniedError(status=ae.status, resource=resource)
 
@@ -98,7 +98,7 @@ def get_namespaced_service(name: str, namespace: str, as_dict: bool = False) -> 
         _namespaced_service_cache[target_service_key] = v1_service
     except ApiException as ae:
         logger.debug(str(ae))
-        _reraise_if_access_denied(ae, resource=f"service/{name}")
+        reraise_if_access_denied(ae, resource=f"service/{name}")
     else:
         return retrieve_namespaced_service_from_cache(target_service_key)
 
@@ -134,7 +134,7 @@ def get_namespaced_pods_by_prefix(
         _namespaced_pods_cache[target_pods_key] = pods_list.items
     except ApiException as ae:
         logger.debug(str(ae))
-        _reraise_if_access_denied(ae, resource="pods")
+        reraise_if_access_denied(ae, resource="pods")
         return []
     else:
         return filter_pods_from_cache(target_pods_key)
@@ -166,7 +166,7 @@ def get_custom_objects(
         _custom_object_cache[target_resource_key] = f(**kwargs)
     except ApiException as ae:
         logger.debug(str(ae))
-        _reraise_if_access_denied(ae, resource=plural)
+        reraise_if_access_denied(ae, resource=plural)
     else:
         return _custom_object_cache[target_resource_key]
 
@@ -188,7 +188,7 @@ def get_cluster_custom_api(group: str, version: str, raise_on_404: bool = False)
         logger.debug(msg=str(ae))
         if int(ae.status) == 404 and raise_on_404:
             raise ResourceNotFoundError(f"{group}/{version} resource API is not detected on the cluster.")
-        _reraise_if_access_denied(ae, resource=f"{group}/{version}")
+        reraise_if_access_denied(ae, resource=f"{group}/{version}")
     else:
         return _cluster_resource_api_cache[target_resource_api_key]
 
@@ -317,7 +317,7 @@ def get_namespaced_secret(namespace: str, secret_name: str) -> dict:
     except ApiException as ae:
         error_msg = str(ae)
         logger.debug(msg=error_msg)
-        _reraise_if_access_denied(ae, resource=f"secret/{secret_name}")
+        reraise_if_access_denied(ae, resource=f"secret/{secret_name}")
     else:
         if result:
             return generic.sanitize_for_serialization(obj=result)
--- a/azext_edge/edge/providers/base.py
+++ b/azext_edge/edge/providers/base.py
@@ -6,6 +6,7 @@
 
 import socket
 from contextlib import contextmanager
+from contextvars import ContextVar
 from typing import Dict, Iterator, List, Optional, Union
 from urllib.request import urlopen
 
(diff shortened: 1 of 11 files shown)


azext_edge/edge/providers/check/base/resource.py
```diff
@@ -117,7 +117,9 @@ def get_resources_by_name(
     resource_name: str,
     namespace: str = None,
 ) -> List[dict]:
-    resources: list = (api_info.get_resources(kind=kind, namespace=namespace) or {}).get("items", [])
+    resources: list = (
+        api_info.get_resources(kind=kind, namespace=namespace, raise_on_access_error=True) or {}
```
digimaun: This is the one place the 401/403 raise is wired in, so the precise access denied result only happens for custom resource list reads. The other cluster reads in the check flow still turn a 403 into a false negative. Pods go through `get_namespaced_pods_by_prefix` which returns `[]`, so akri gets none of this even though it is in the summary, and the pod health checks across services are in the same boat. API discovery in `get_cluster_custom_api` swallows the error and reports "API resources not detected", references via `get_valid_resource_names` and secrets return empty, and the version, storage, and node prechecks keep the generic handling. Could we push the 401/403 detection down to the shared read helpers, or centralize it, so the whole command benefits rather than just the CR list path? As is, a partial access principal denied on pods or references still sees the old misleading output.

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -119,7 +146,14 @@ def check_post_deployment(
             append_resource = True
 
         if append_resource:
-            results.append(evaluate_func(detail_level=detail_level, as_list=as_list, resource_name=resource_name))
+            try:
+                results.append(
+                    evaluate_func(detail_level=detail_level, as_list=as_list, resource_name=resource_name)
+                )
+            except ClusterAccessDeniedError as access_error:
```
digimaun, line 165: This catch only covers `ClusterAccessDeniedError`, and one nearby read path is not routed through it. `get_config_map` rethrows a plain `ApiException` on a non 404, so a 403 on a referenced configmap raises past this handler and aborts rather than producing the graceful result. The secret path returns None instead, so the two references behave differently from each other and from the CR path. This abort predates this PR, but since the goal here is a clean per resource access denied with no full command abort, routing those reference reads through the same `ClusterAccessDeniedError` path would make it consistent.

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -119,7 +146,14 @@ def check_post_deployment(
             append_resource = True
 
         if append_resource:
-            results.append(evaluate_func(detail_level=detail_level, as_list=as_list, resource_name=resource_name))
+            try:
+                results.append(
+                    evaluate_func(detail_level=detail_level, as_list=as_list, resource_name=resource_name)
```
digimaun, line 163: The catch is per evaluate_func, and a few of them read more than one kind. Asset eval reads endpoint profiles, dataflow eval reads profiles and endpoints, opcua runtime conditionally reads instances. If the inner read is the one denied, we surface a single access denied result labeled with the outer kind and drop whatever the func had already built. So the "per resource" message can name the wrong resource. Not sure it is worth restructuring now, but flagging that the granularity is per evaluator, not per read.

azext_edge/edge/providers/base.py
```diff
@@ -46,6 +46,15 @@ def load_config_context(context_name: Optional[str] = None):
     DEFAULT_NAMESPACE = current_config.get("namespace") or "azure-iot-operations"
 
 
+class ClusterAccessDeniedError(Exception):
+    """Raised when a cluster API call is rejected with HTTP 401/403 due to missing permissions."""
```
digimaun: Small precision point. We treat 401 and 403 the same and tell the user they lack permission and to grant read access. A 401 is usually an authentication problem, an expired or invalid kubeconfig token, not an authorization one, so that remediation can be wrong in the one case this feature exists to make clear. Could we distinguish them, or at least soften the 401 wording to access denied without asserting a permissions cause?

azext_edge/edge/providers/check/base/deployment.py
```diff
@@ -25,6 +25,33 @@
 logger = get_logger(__name__)
 
 
+def _build_access_denied_result(resource: ListableEnum, error: ClusterAccessDeniedError, as_list: bool = False) -> dict:
```
digimaun, line 34: This is a behavior change to a critical diagnostic and there are no tests for it. The check providers have solid existing coverage, and `test_resource_unit.py:1002` still models a 403 as returning None. A focused test that mocks an `ApiException` with status 403 through `get_resources_by_name` and asserts the access denied result shape, the `accessDenied` marker, and the summary footer flip would lock the propagation and the summary rollup in place. A 401 case and a nested read case would cover the two points above too.

ketkimnaik: Done

azext_edge/edge/providers/base.py
ketkimnaik: 401 now uses authentication-specific wording ("Could not authenticate… kubeconfig token may be expired or invalid") while 403 keeps the "lacks permission / grant read access" remediation, and the neutral summary footer no longer asserts a permissions cause.

azext_edge/edge/providers/check/base/deployment.py
ketkimnaik: Now the access-denied message names the actual denied read (error.resource) instead of the outer evaluator kind, so multi-read evaluators like asset/dataflow report the correct resource.

azext_edge/edge/providers/check/base/deployment.py
ketkimnaik: Now routed the configmap read through the same access-denied path, so a 403 on a referenced configmap now degrades gracefully like the secret and CR paths instead of aborting.

azext_edge/edge/providers/check/base/resource.py
ketkimnaik: Good call — I've centralized the 401/403 detection so pods, API discovery, references, and secrets are all covered, and the whole check command now reports a precise access-denied result instead of false negatives.

azext_edge/edge/providers/check/base/deployment.py
digimaun: Naming error.resource fixes the mislabel. One regression to flag though. At 26e8bb3 only the CR list raised. This follow-up also makes the listener service read and the secret reference read raise, and since the catch is per evaluator, a denial on either now unwinds the whole evaluator and drops what it already built. A listener service 403 used to show the listener findings plus an Unable to fetch service note, and now it returns only the access denied result for that service. The secret reference read regressed the same way. The configmap path is fine since it used to abort and now degrades. Could we catch ClusterAccessDeniedError at those secondary read sites and add an inline access denied row to the current check_manager, so the evaluator keeps its other findings? The per evaluator catch stays a fine backstop for reads that are the whole subject, like pods or the CR list, where there is nothing else to preserve.

azext_edge/edge/providers/check/base/resource.py
digimaun: Moving this to a context gate is a clean way to cover pods, API discovery, references, and secrets in one place. One gap remains. check_pre_deployment runs before the with block in run_checks, and the version and node prechecks read the cluster directly instead of through the shared helpers, so a 403 there still falls through to the generic output. Post-deployment is precise now while the prechecks are not, so a single run can show both. Could we fold those into the same handling, or leave the prechecks to a follow-up?

azext_edge/edge/providers/check/base/deployment.py
digimaun: The new tests lock down the main paths, custom objects, the per evaluator catch, discovery, the 401 and 403 wording, and the summary flip. One gap worth closing. The service, pod, secret, and configmap raise branches in the shared helpers are not exercised, so a future edit to one of those except blocks would not trip a test. A small parametrized test that drives each helper through reraise_if_access_denied and asserts it raises inside the context and swallows outside would cover them.

azext_edge/edge/providers/check/base/deployment.py
ketkimnaik: Fixed — the listener service and secret/configmap reference reads now catch the denial at the read site and add an inline access-denied row, so the evaluator keeps its other findings instead of unwinding. The per-evaluator catch stays as the backstop for whole-subject reads like pods and the CR list.

azext_edge/edge/providers/check/base/resource.py
ketkimnaik: Added the gate for pre deployment runs as well.. It is better to cover in this same PR.

azext_edge/edge/providers/check/base/deployment.py
ketkimnaik: Added tests that exercise each shared helper's access-denied branch, so those paths are now covered.

azext_edge/edge/providers/check/mq.py
```diff
@@ -1008,9 +1032,30 @@ def _evaluate_listener_service(
...
 
-    associated_service: dict = get_namespaced_service(
-        name=listener_spec_service_name, namespace=namespace, as_dict=True
-    )
+    associated_service: dict = None
+    try:
+        associated_service = get_namespaced_service(
+            name=listener_spec_service_name, namespace=namespace, as_dict=True
+        )
+    except ClusterAccessDeniedError as access_error:
```
digimaun, line 1059: This is the right shape for the listener service read. The diagnostics service read in `_evaluate_broker_diagnostics_service` is the same kind of secondary `get_namespaced_service` call inside `evaluate_brokers`, and it is not wrapped the same way. Since the broker findings are built first, a 401 or 403 there unwinds the whole evaluator and drops them, which is the case we are trying to avoid. Could we wrap that read too, or render it inline, so a denied diagnostics service keeps the broker results? The pod health reads in the same evaluator have the same shape, though those returned empty before so they are a smaller concern.

ketkimnaik: Wrapped the diagnostics service read in same way.

azext_edge/edge/providers/check/mq.py
```diff
@@ -1143,9 +1188,28 @@ def _evaluate_broker_diagnostics_service(
...
     detail_level: int = ResourceOutputDetailLevel.summary.value,
 ) -> None:
-    diagnostics_service = get_namespaced_service(
-        name=AIO_BROKER_DIAGNOSTICS_SERVICE, namespace=namespace, as_dict=True
-    )
+    try:
+        diagnostics_service = get_namespaced_service(
+            name=AIO_BROKER_DIAGNOSTICS_SERVICE, namespace=namespace, as_dict=True
+        )
+    except ClusterAccessDeniedError as access_error:
```
digimaun, line 1214: This is the right fix for the diagnostics read, thanks. To finish the consistency pass, the pod reads in `evaluate_brokers`, the `get_namespaced_pods_by_prefix` loop, are the last secondary read that still unwinds. A denial there would drop the broker findings the same way the diagnostics read did before this change. Now that the listener, diagnostics, secret, and configmap reads all preserve findings, wrapping the pod reads to report inline and keep going would make `evaluate_brokers` fully consistent. Fine as a fast follow up if you would rather not grow this PR.

digimaun (dismissed):

azext_edge/edge/providers/check/mq.py
ketkimnaik: Covered in this same PR.

digimaun (approved):

PR #905, opened 2026-07-22
fix: align 2607 fresh-install configuration defaults

cheatsheet1999:
## Summary
Align the 2607 configuration behavior between fresh installations and upgrades:

- Keep `connectors.values.gdsManager.enabled` opt-in by omitting it from default deployment settings. Users can enable it explicitly through `--ops-config`.
- Keep the deployment-specific `connectors.values.securityPki.subjectName` for fresh installations without backfilling it during upgrades.
- Set the following Cert Manager settings to `false` on fresh installations while allowing overrides through `--cm-config`:
  - `trust-manager.secretTargets.enabled`
  - `trust-manager.secretTargets.authorizedSecretsAll`
- Preserve existing extension configuration during upgrades without automatically adding any of these four settings.

## Validations:
- Verified a live 2.7.0 -> 2.8.0 upgrade:
  - Existing IoT Operations and Cert Manager configuration was preserved.
  - None of the four settings was backfilled.
  - The existing OPC UA certificate identity remained unchanged.
  - GDS state remained absent.
- Verified a fresh 2.8.0 installation:
  - GDS was not enabled by default.
  - A unique OPC UA broker certificate subject was generated.
  - Both Cert Manager settings were explicitly set to `false`.

* **`yuelin-pr905-verify-up-rg` / `pr905-verify-up`**: Verified a healthy upgrade from **2.7.0 to 2.8.0**. All four settings remain absent, the existing generic certificate and serial number are unchanged, no GDS state is present, and post-deployment health checks pass.

* **`yuelin-pr905-verify-fresh-rg` / `pr905-verify-fresh`**: First verified the fresh-install defaults: GDS is absent, the subject is unique, and both Cert Manager values are set to `false`. Then reinstalled with install-time overrides: `--cm-config` set both values to `true`, and `--ops-config` enabled GDS. The retained state shows the unique subject, all explicit overrides set to `true`, GDS state present, a GDS-issued certificate, and passing health checks.


azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1175,7 +1237,7 @@ def _has_delta_in_config(self) -> bool:
                         has_mqtt_migration = True
                         break
 
-        return self.override.config or self.config_delta or has_mqtt_migration
+        return self.override.config or self.config_delta or has_mqtt_migration or self._get_2607_config_delta()
```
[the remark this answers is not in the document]
cheatsheet1999: Done

cheatsheet1999:
IT Workflow in progress

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1108,6 +1124,52 @@ def _get_mqtt_migration_config(self) -> dict:
...
+            return {}
+
+        target_version = self.semver.parse(self.desired_version[0])
+        current_config = self.extension.get("properties", {}).get("configurationSettings", {})
+        target_config = {}
+
+        if self.moniker == EXTENSION_MONIKER_OPS and target_version >= self.semver.parse(
+            MIN_INSTANCE_VERSION_FOR_GDS_MANAGER
+        ):
+            target_config[OPS_GDS_MANAGER_CONFIG] = OPS_GDS_MANAGER_DEFAULT
```
digimaun: Can we drop connectors.values.gdsManager.enabled from the automatic backfill and keep it opt-in? Same for the deployment side where the template sets it by default, so fresh installs and upgrade don't set it unless the user explicitly adds it. I'll follow up with the context separately.

cheatsheet1999: For sure, we can definitely take this approach.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1108,6 +1122,51 @@ def _get_mqtt_migration_config(self) -> dict:
...
 
+    def _get_2607_config_delta(self) -> Dict[str, str]:
+        if not self.extension or not self.desired_version[0]:
+            return {}
+
+        target_version = self.semver.parse(self.desired_version[0])
+        current_config = self.extension.get("properties", {}).get("configurationSettings", {})
+        target_config = {}
+
+        if self.moniker == EXTENSION_MONIKER_OPS and target_version >= self.semver.parse(
```
digimaun: Could we omit the subjectName backfill in this release? This value is the broker's application instance certificate subject, and the connector requires it to match its certificate. On an existing cluster, applying a new subject either reissues that cert or leaves the config subject and the cert out of sync, and either way can prevent new OPC UA sessions until trust and config are realigned. Fresh installs are unaffected since there is no prior trust. A subject unique to each cluster is the right end state, but on existing clusters it is a deliberate cert migration, better handled on its own than as an automatic backfill here.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1108,6 +1122,51 @@ def _get_mqtt_migration_config(self) -> dict:
...
+            if (
+                len(suffix) == OPS_EXTENSION_SUFFIX_LENGTH
+                and suffix.isascii()
+                and suffix.isalnum()
+            ):
+                target_config[OPS_SUBJECT_NAME_CONFIG] = f"{OPS_SUBJECT_NAME_PREFIX}{suffix}"
+        elif self.moniker == EXTENSION_MONIKER_CM and target_version >= self.semver.parse(
+            MIN_CM_VERSION_FOR_SECRET_TARGETS
+        ):
+            target_config = DEFAULT_CM_SECRET_TARGET_CONFIG
```
digimaun: Both match the trust-manager default of false, so backfilling them on an existing cluster changes nothing, no new permissions and no behavior change. Extension config is sticky though, so an explicit false persists as an override we would have to clear later if the default ever changes, for no gain. I think we could keep these on fresh install and drop them from the upgrade path. With subjectName also leaving the backfill, _get_2607_config_delta becomes empty, so the method and its two call sites can go.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: We can move forward with this approach. Thank you.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: Thanks for the thorough review and for helping remove the unnecessary changes. I really appreciate it. The PR is ready for another review.

digimaun (approved):

PR #907, opened 2026-07-27
fix: repair existing SecretSync FIC issuer mismatch

cheatsheet1999:
## Summary

#899 ensures newly created SecretSync FICs use the canonical issuer from OIDC discovery.

Existing installations may still have a FIC created from the ARM issuer whose value differs from the discovery issuer by one trailing slash. Rerunning `az iot ops secretsync enable` did not repair it because the command returned when it found existing SecretSync resources, before resolving the issuer or checking the FIC.

## Fix

**For the affected case:**

 ARM issuer:       https://issuer/       (old FIC copied this)
 Discovery issuer: https://issuer        (canonical token issuer)
 Existing FIC:     same SecretSync subject, issuer=https://issuer/

---
Resolve the discovery issuer before the existing-resource early return and repair an existing slash-equivalent FIC in place.

- Attempt repair only when discovery verifies an issuer that differs from ARM.
- Match the SecretSync FIC by service-account subject.
- Keep the existing subject and audiences, replace only the issuer.
- Do not create a new FIC when no matching existing credential is found.
- Keep the existing early return so SecretSync resources are not recreated.

## Some clarifications...

**Q: What happens when the existing FIC has a trailing-slash mismatch?**  
A: The same FIC resource is updated with the discovery issuer. Its name, subject, and audiences remain unchanged.

**Q: What happens when the existing FIC is already correct?**  
A: It is found as an exact match and no update is made.

**Q: Does this change first-time SecretSync enablement?**  
A: No. New enablement continues through the existing creation path and creates the FIC using the resolved discovery issuer.

**Q: What happens when discovery cannot verify the issuer?**  
A: The ARM value is retained and no repair is attempted. This avoids rewriting an existing FIC toward an unverified value.

**Q: Can this create a duplicate FIC?**  
A: No. The repair path only updates a matching existing credential by name and does nothing when no match is found.

## E2E details:
[yuelin-pr907-fic-e2e-rg](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/a386d5ea-ea90-441a-8263-d816368c84a1/resourceGroups/yuelin-pr907-fic-e2e-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/yuelin-pr907-fic-e2e-mi/federatedcredentials)

### DIFF
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -8,7 +8,7 @@ import json
 import re
 import time
 from contextlib import nullcontext
-from typing import Dict, Iterable, List, Optional
+from typing import Dict, Iterable, List, Optional, Tuple
 from urllib.parse import urlsplit
 
 import requests
@@ -164,7 +164,7 @@ def _get_public_discovery_issuer(arm_issuer: str) -> Optional[str]:
     return None
 
 
-def resolve_oidc_issuer(arm_issuer: str) -> str:
+def resolve_oidc_issuer(arm_issuer: str) -> Tuple[str, bool]:
     discovery_issuer = _get_public_discovery_issuer(arm_issuer)
 
     if discovery_issuer:
@@ -175,14 +175,14 @@ def resolve_oidc_issuer(arm_issuer: str) -> str:
                 arm_issuer,
                 discovery_issuer,
             )
-        return discovery_issuer
+        return discovery_issuer, True
 
     logger.warning(
         "Could not verify the OIDC issuer '%s' from this host. The ARM-reported value will be used "
         "for the federated identity credential. This may still work if Microsoft Entra can reach the issuer.",
         arm_issuer,
     )
-    return arm_issuer
+    return arm_issuer, False
 
 
 def get_enable_syntax(instance_name: str, resource_group_name: str) -> str:
@@ -473,14 +473,14 @@ class Instances(Queryable):
             namespace = custom_location["properties"]["namespace"]
             cred_subject = get_cred_subject(namespace=namespace, service_account_name=SERVICE_ACCOUNT_SECRETSYNC)
             arm_oidc_issuer = self._ensure_oidc_issuer(cluster_resource, use_self_hosted_issuer)
-            oidc_issuer = resolve_oidc_issuer(arm_issuer=arm_oidc_issuer)
+            oidc_issuer, issuer_verified = resolve_oidc_issuer(arm_issuer=arm_oidc_issuer)
 
             secretsync_resources = resource_map.connected_cluster.get_cl_resources_by_type(
                 custom_location_id=instance["extendedLocation"]["name"],
                 resource_types={SPC_RESOURCE_TYPE, SECRET_SYNC_RESOURCE_TYPE},
             )
             if secretsync_resources:
-                if oidc_issuer != arm_oidc_issuer:
+                if issuer_verified:
                     self._repair_federated_cred_issuer(
                         mi_resource_id_container=mi_resource_id_container,
                         issuer_url=oidc_issuer,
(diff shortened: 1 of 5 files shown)


azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -472,21 +472,27 @@ def enable_secretsync(
...
-            oidc_issuer = self._ensure_oidc_issuer(cluster_resource, use_self_hosted_issuer)
+            arm_oidc_issuer = self._ensure_oidc_issuer(cluster_resource, use_self_hosted_issuer)
+            oidc_issuer = resolve_oidc_issuer(arm_issuer=arm_oidc_issuer)
 
             secretsync_resources = resource_map.connected_cluster.get_cl_resources_by_type(
                 custom_location_id=instance["extendedLocation"]["name"],
                 resource_types={SPC_RESOURCE_TYPE, SECRET_SYNC_RESOURCE_TYPE},
             )
             if secretsync_resources:
+                if oidc_issuer != arm_oidc_issuer:
```
digimaun: This gate ties the repair to ARM and discovery disagreeing, but the thing we care about is whether discovery verified an issuer. resolve_oidc_issuer returns a plain string, so a verified issuer that equals the ARM value looks identical to the fallback where discovery failed. A credential still holding the trailing slash is then skipped in the one case where we have the most confidence, which is discovery succeeding and confirming ARM. Could we gate on verification instead, either by calling _get_public_discovery_issuer directly since it already returns Optional[str], or by returning the resolved value along with whether it was verified?

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -1049,6 +1055,39 @@ def federate_msi(
             },
         )
 
+    def _repair_federated_cred_issuer(
```
digimaun, line 1058: The list and update calls here are unguarded, so a repair the user did not ask for can fail the command. Re-running enable on an instance that already has secret sync resources used to warn and return, and now a permission problem or a transient ARM error surfaces as a raw failure with no indication that a credential repair was what failed. _attempt_keyvault_role_assignments already handles this shape on this same command, catching HttpResponseError and re-raising a ValidationError with a message that names what the user needs to do. Could we follow that pattern here and name the credential plus the manual az identity federated-credential update remediation?

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: You’re right, I based the gate on the originally reproduced ARM/discovery mismatch, so I missed the case where discovery confirms ARM but an existing FIC still has the trailing-slash variant. I’ll update the resolver to return the issuer with its verification status, gated repair on that status, and added coverage for verified discovery versus fallback. Thanks for catching this.

azext_edge/edge/providers/orchestration/resources/instances.py
cheatsheet1999: That makes sense. Thank you. I updated both the FIC list and update paths to catch `HttpResponseError` and raise an actionable `ValidationError` instead of surfacing the raw SDK error. Could you please review the error messages when you get a chance?

digimaun: Two thoughts on the messages. Could we lead with the underlying error and close with the remediation? get_user_msg_warn_ra in this file follows that shape, opening with the failure plus the error and ending on "Please handle this step before continuing", which also leaves the actionable line closest to the prompt.

On the update failure, "then rerun the command:" sits directly above the az identity command, so it reads as though that is the command to rerun. Running it already repairs the credential, so rerunning enable finds an exact match and does nothing. Could we drop the rerun and let the command stand as the fix, something like "Update its issuer manually with:"? The list failure is different since there the rerun is what applies the repair once permission is sorted.

cheatsheet1999: For sure, I reordered both messages to lead with the underlying error and close with the remediation. 

The list failure now directs the user to resolve the error and rerun enable, while the update failure ends with the manual update command

digimaun (approved):

PR #908, opened 2026-07-30
ci: enable subprocess coverage for integration tests

ketkimnaik:
## Summary
Integration tests invoke the CLI as child `az iot ops ...` processes (via `helpers.run` -> `subprocess.run`). coverage.py only traces the process it starts in, so handler execution in those child processes was invisible — ops integration coverage read as ~0%.

This enables coverage.py subprocess measurement so those child processes are traced and combined into the report.

## Changes
- **`.coveragerc`**: add `parallel = True` (per-process data files), switch to `source_pkgs = azext_edge.edge`, and add a `[paths]` mapping so the installed-extension path and repo source combine into one dataset.
- **`scripts/coverage/sitecustomize.py`** (new): startup hook that calls `coverage.process_startup()` to arm measurement inside each `az` subprocess. No-op when `COVERAGE_PROCESS_START` is unset or coverage is unavailable.
- **`tox.ini`**: append `scripts/coverage` to `PYTHONPATH` and set `COVERAGE_PROCESS_START` in the integration env so subprocesses self-arm; add `coverage combine --append` to the `report` env to merge per-process data files.

## Validation
- Verified against a live cluster: a real `az iot ops check` run produced per-process data files that combined into non-zero coverage for check-provider modules that previously reported 0% (e.g. `providers/check/common.py` 89%, `base/node.py` 79%, `base/check_manager.py` 73%, `summary.py` 31%).
- `flake8 scripts/coverage/sitecustomize.py --config=setup.cfg` passes.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,13 @@
 # .coveragerc to control coverage.py
 [run]
 branch = True
-source = azext_edge/edge
+# parallel writes a uniquely-suffixed data file per process so that the pytest
+# process and every `az` subprocess spawned by integration tests can be measured
+# without clobbering each other. Combined later via `coverage combine`.
+parallel = True
+# Measure by importable package name (not path) so the installed extension used
+# by `az iot ops ...` subprocesses is measured wherever it lives on disk.
+source_pkgs = azext_edge.edge
 omit =
     # ignore all tests
     azext_edge/tests/*
@@ -12,3 +18,11 @@ omit =
     azext_edge/edge/command_map.py
     # skip params and command map
     azext_edge/edge/providers/rpsaas/adr/params.py
+
+[paths]
+# Map the various on-disk locations of the extension (repo source, tox target,
+# `az extension add` cliextensions dir) back to the repo source so data from the
+# pytest process and the `az` subprocesses combines into a single set of files.
+source =
+    azext_edge/edge
+    */azext_edge/edge
--- a/scripts/coverage/sitecustomize.py
+++ b/scripts/coverage/sitecustomize.py
@@ -1,0 +1,27 @@
+# coding=utf-8
+# ----------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License file in the project root for license information.
+# ----------------------------------------------------------------------------------------------
+
+"""
+sitecustomize hook to enable coverage.py measurement inside subprocesses.
+
+Integration tests invoke the CLI as child ``az iot ops ...`` processes (see
+``azext_edge/tests/helpers.py::run`` -> ``subprocess.run``). coverage.py only
+traces the process it starts in, so those child processes would otherwise be
+invisible to coverage, which is why integration coverage reads as ~0%.
+
+When this directory is on ``PYTHONPATH`` and ``COVERAGE_PROCESS_START`` points at
+the coverage config, Python imports this module at interpreter startup and
+``coverage.process_startup()`` arms measurement for the child process. It is a
+no-op when ``COVERAGE_PROCESS_START`` is unset and when coverage is unavailable,
+so it is safe for every process that inherits ``PYTHONPATH``.
+"""
+
+try:
+    import coverage
+
+    coverage.process_startup()
+except Exception:  # pragma: no cover - never break a subprocess if coverage is missing
+    pass
--- a/tox.ini
+++ b/tox.ini
@@ -84,7 +84,11 @@ passenv =
 setenv =
     # You can temporarily add variables here to modify your tests
     # azext_edge_skip_init=true
-    PYTHONPATH={envsitepackagesdir}/azure-cli-extensions/azure-iot-ops
+    # scripts/coverage is appended so `az` subprocesses import sitecustomize.py and
+    # arm coverage via coverage.process_startup() (see COVERAGE_PROCESS_START below).
+    PYTHONPATH={envsitepackagesdir}/azure-cli-extensions/azure-iot-ops{:}{toxinidir}{/}scripts{/}coverage
+    # enable coverage.py measurement inside the `az iot ops ...` subprocesses the int tests spawn
+    COVERAGE_PROCESS_START={toxinidir}{/}.coveragerc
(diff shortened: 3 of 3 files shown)


.coveragerc
```diff
@@ -1,7 +1,13 @@
...
+# parallel writes a uniquely-suffixed data file per process so that the pytest
+# process and every `az` subprocess spawned by integration tests can be measured
+# without clobbering each other. Combined later via `coverage combine`.
+parallel = True
+# Measure by importable package name (not path) so the installed extension used
+# by `az iot ops ...` subprocesses is measured wherever it lives on disk.
+source_pkgs = azext_edge.edge
 omit =
     # ignore all tests
     azext_edge/tests/*
```
digimaun, line 13: These omit patterns are relative, so they only match files measured under the current directory. The `az` subprocesses now being measured import the extension from the tox install target, so the vendored SDK, params.py and command_map.py get recorded at that path instead, the patterns miss, and `[paths]` remaps them back into the report. That is around 151k lines of generated vendor code plus the two declarative files landing in the report as covered, which moves the number the opposite way from what this PR is after. I reproduced it on coverage 7.15.2 and confirmed that prefixing each pattern with `*/` restores the omission for both the installed path and the in repo unit runs. The adr/params.py entry can go at the same time, that file no longer exists.

scripts/coverage/sitecustomize.py
```diff
@@ -0,0 +1,27 @@
...
+``coverage.process_startup()`` arms measurement for the child process. It is a
+no-op when ``COVERAGE_PROCESS_START`` is unset and when coverage is unavailable,
+so it is safe for every process that inherits ``PYTHONPATH``.
+"""
+
+try:
+    import coverage
+
+    coverage.process_startup()
+except Exception:  # pragma: no cover - never break a subprocess if coverage is missing
```
digimaun, line 26: The docstring says this is a no op when coverage is unavailable, which is the ImportError case. Catching Exception also swallows a bad COVERAGE_PROCESS_START or a coverage failure, and then we are back to an empty report with no signal about why. Could we narrow this to ImportError so the intended case stays quiet and a real misconfiguration surfaces?

scripts/coverage/sitecustomize.py
```diff
@@ -0,0 +1,27 @@
...
+invisible to coverage, which is why integration coverage reads as ~0%.
+
+When this directory is on ``PYTHONPATH`` and ``COVERAGE_PROCESS_START`` points at
+the coverage config, Python imports this module at interpreter startup and
+``coverage.process_startup()`` arms measurement for the child process. It is a
+no-op when ``COVERAGE_PROCESS_START`` is unset and when coverage is unavailable,
+so it is safe for every process that inherits ``PYTHONPATH``.
+"""
+
+try:
```
digimaun, line 22: Coverage 7.10 added `[run] patch = subprocess`, which arms subprocesses without a startup hook and would replace this file and the scripts/coverage entry on PYTHONPATH with one config line. The catch is that coverage is unpinned in dev_requirements, so it would need a floor. The current approach works on any version, so this is just raising the simpler end state if we want to take the pin.

zzcodework:

Reviewing the **full diff** (head `a9d31242`, base `dev`).

The diagnosis is right and the mechanism is the standard one: integration tests shell out to `az iot ops ...` via `subprocess.run`, coverage only traces its own process, hence ~0% integration coverage. `sitecustomize.py` + `COVERAGE_PROCESS_START` + `parallel = True` + `coverage combine` is exactly the documented coverage.py recipe, and the comments explaining each knob are genuinely useful. Two things look likely to make it still report no data.

**Blocking**

_None identified — but see finding 1, which I believe will prevent the change from actually working._

**Suggestions**

1. `tox.ini` / `.coveragerc` — with `parallel = True`, each process writes `.coverage.<host>.<pid>.<random>` **into its own current working directory**, and `coverage combine` only looks in the directory of the data file (repo root here). The `az` subprocesses spawned by the integration tests do not necessarily run with the repo root as CWD, so their data files can land somewhere `coverage combine` never sees — and the failure is silent because the leading `-` on `- coverage combine --append` swallows "no data to combine". Pin the location explicitly so every process writes to the same file set:

   ```ini
   setenv =
       COVERAGE_FILE={toxinidir}{/}.coverage
   ```

   (set it in both the test env and the `coverage` env). Worth verifying end to end that the reported integration coverage actually moves off ~0% before merging — that is the only real proof this works.

2. `.coveragerc` — switching `[run] source` to `source_pkgs = azext_edge.edge` is the right call for measuring the installed extension, but the `omit` entries are still repo-relative globs (`azext_edge/tests/*`, `azext_edge/edge/params.py`, ...). Those are matched against the file paths in the data, and files measured inside the `cliextensions` install dir are absolute paths that will not match. The `[paths]` block only remaps `*/azext_edge/edge`, so nothing rewrites the test paths. Prefixing the omit patterns with `*/` (e.g. `*/azext_edge/tests/*`) makes them match regardless of install location.

3. `scripts/coverage/sitecustomize.py` — Python imports **only the first** `sitecustomize` found on `sys.path`. Appending `scripts/coverage` to `PYTHONPATH` is safe today, but if any other component on the path ships its own `sitecustomize`, one of the two is silently dropped depending on ordering. A short comment noting this (or defensively chaining to a pre-existing `sitecustomize` via `importlib`) would save a future debugging session. Also, `except Exception: pass` swallows a genuine misconfiguration of `COVERAGE_PROCESS_START` — a `sys.stderr` note gated behind a debug env var would make "why is coverage still zero" answerable.

**Nits**

4. `.coveragerc` still ends without a trailing newline (the diff shows `\ No newline at end of file` on both sides).

5. `tox.ini` — the `- coverage combine --append` step is only in the `coverage` env. Anyone running `coverage report` directly after a local test run will now see no data (because `parallel = True` means nothing is ever written to plain `.coverage`), which is a confusing regression for local workflows. Worth a line in `docs/tox-testing.md` noting that `coverage combine` is now required.

**Summary** — right approach and well documented, but I would not expect it to produce non-zero subprocess coverage without pinning `COVERAGE_FILE` (finding 1), and the `omit` globs likely stop matching for the installed-extension paths (finding 2). Please confirm the measured integration coverage actually changes before merging.

PR #909, opened 2026-08-10
docs: add AIO CLI maintainer agent and template sync skill

cheatsheet1999:
## Summary

- Add a repository-local `aio-cli-maintainer` Copilot agent.
- Add the `sync-aio-bicep-templates` skill.
- Guide safe generation of embedded ARM templates from AIO Bicep release branches.
- Detect and report configuration or behavioral changes before editing.

## Usage
Run /agent or /skills inside copilot CLI or VS Code Copilot
image
image

## TL;DR
1. It reproduces a shipped release bit-for-bit. Run against the 2608 branch, it regenerated PR #913's `template.py`  byte-for-byte, including both `metadata._generator.templateHash` values. That hash is computed by the Bicep compiler over template content. Matching it means the same source produced the same output.
2. A context-free agent ran it end-to-end and got it right. A fresh agent with no session history, given only this file plus the standard inputs, produced an ENABLEMENT template with 0 differences from `dev` 
3.  It independently caught the bug #914 had to fix.


zzcodework:

Reviewing `full diff` (head `8c92070`) — 2 new files under `.github/agents/` and `.github/skills/`, documentation only, no runtime, command-surface, or packaging changes. The existing Copilot comment about the stray "Agency" wording appears already addressed at this head, so it is not repeated.

**Blocking**

- None.

**Suggestions**

- `.github/skills/sync-aio-bicep-templates/SKILL.md`, section 2: the statement that the optimizer's Python-output mode "invokes Black as a single executable string and does not work reliably" is correct — `tools/template_optimizer.py` calls `run(f"black {output_template_path} ...", check=True)` with `subprocess.run` and no `shell=True`, so it raises `FileNotFoundError` on a string command. Rather than permanently documenting a workaround around a two-line bug, consider fixing the optimizer (pass a list, or `[sys.executable, "-m", "black", ...]`) and simplifying this section. If the fix is out of scope for this PR, a `TODO`/issue link here would keep the workaround from becoming permanent.
- Note that the JSON-mode workaround silently drops the optimizer's own round-trip assertion (`assert payload_to_integrate == data` / `assertDictEqual`) that the Python path performs, since you re-serialize with `repr()` outside the tool. Section 5 step 1 re-parses `template.py` and compares, which covers this — worth saying so explicitly so a reader does not think the check was lost.
- Required inputs vs. section 6: the input is described as "Local path to `azure-iot-operations-tests`", while the PR description and the wiki links refer to the "Azure IoT Operations deployment repository" / `Azure/azure-iot-operations`. Since section 1 validates `Deployment/*.bicep`, `Deployment/release.json` and `AioExtension/helm/aio/values.yaml` against that path, please name the exact repository once and use it consistently — a wrong-repo input fails late and confusingly.
- Section 4 updates `template.py`, `EXTENSION_CONFIGS` in `test_template_unit.py`, and `AIO_RELEASE`/`VERSION` in `azext_edge/constants.py`, but does not mention `HISTORY.rst`. A release-moniker bump is normally user-visible; consider adding a step to prompt for (or draft) the changelog entry so it is not forgotten in the release flow.
- The `VERSION` policy says "`stable` with an already stable version: ask for an explicit version". The current `azext_edge/constants.py` is at `VERSION = "2.8.0"` (stable) with `AIO_RELEASE = "2607"`, so the very next stable run lands in the "ask" branch. Worth confirming that is the intended default behavior rather than an accidental dead end.

**Nits**

- Section 3 says "Derive each `commit_id` using the latest commit at the selected ref that changed its top-level Bicep file", then adds "If imported Bicep changes make that provenance insufficient, ask for an explicit override". Since imported modules changing without the top-level file changing is the common case, consider making the import-aware check an explicit step rather than a conditional caveat.
- Consistency with the sibling PR in azure-iot-cli-extension#877: that agent file omits the `tools:` frontmatter key that this one declares. Worth aligning the two agent profiles.

Summary: clean, well-structured, and the "report behavioral changes before editing" framing (with the 2607 lesson called out) is the right instinct. Main improvement is fixing the underlying `template_optimizer.py` Black invocation instead of documenting around it, plus tightening the source-repository naming.

HangyiWang (dismissed):

zzcodework:

Reviewing `changes since 8c92070` (head `9aedfc8`) — one commit, "clarify AIO template source tracking", touching `.github/agents/aio-cli-maintainer.agent.md` and `.github/skills/sync-aio-bicep-templates/SKILL.md` only. Documentation-only; no runtime, command-surface, or packaging changes.

**Blocking**

- None.

**Suggestions**

- Two items from the previous review are still open, both deliberate-looking, so just flagging rather than re-arguing:
  - The underlying `tools/template_optimizer.py` Black invocation (`run(f"black {path} ...")` via `subprocess.run` with no `shell=True`, which raises `FileNotFoundError`) is still unfixed, and section 2 still documents the JSON-mode workaround around it. If fixing it is out of scope here, a `TODO` or issue link next to that paragraph would stop the workaround from becoming permanent.
  - Section 4 still does not mention `HISTORY.rst`. A release-moniker/`VERSION` bump is user-visible, so a prompt-for-changelog step would close the loop with the repo's release conventions.

**Nits**

- Section 3 step 3 says "Compare every referenced file at that top-level commit with the same file at the selected ref. Treat a missing file as a difference." Worth stating explicitly that comparison is by file *content* at the two refs (not commit identity), since a file touched and reverted between the two points is not a real provenance divergence and would otherwise send the workflow into the step-5 "ask for an explicit `commit_id`" branch unnecessarily.

Summary: this iteration addresses the naming, provenance, and round-trip-assertion feedback well. The import-aware provenance rewrite in section 3 is the substantive improvement — it turns the previous conditional caveat into a deterministic procedure, which is exactly what an agent following this skill needs. Confirming the other resolutions: the source repository is now consistently named `azure-iot-operations-tests` in both the agent file and the skill's required inputs, the JSON-mode round-trip coverage is explicitly attributed to section 5, and the `stable`-on-`stable` VERSION rule now states its rationale.

zzcodework:

Reviewing `changes since 9aedfc8` (head `d290925`)

The incremental change is documentation-only: commits `f2943d0` ("docs: harden AIO template sync skill") and `b091751` ("fix: correct release validation"), both touching `.github/skills/sync-aio-bicep-templates/SKILL.md`. The rest of the compare range is `dev` merged in, not PR content. No CLI command surface, parameters, or help text change, so no `_help.py` / `_params.py` / changelog follow-up applies.

**Blocking**

None.

**Suggestions**

- `.github/skills/sync-aio-bicep-templates/SKILL.md` (section 5 literal-currency check, `lit = str(node)` / `any(s == lit or f"'{lit}'" in s for s in shipped)`): tightening the bare-substring test to an exact or quoted-token match is a real improvement — the old `str(node) in s` would pass on any short value. But the comparison is still done on `str(node)`, and the values read out of `values.yaml` are not all strings. A YAML boolean gives `True` while the template carries `true`, and an int/float or a nested node gives a Python repr that can never appear in the ARM output, so those reads now report `FAIL ... is current` even when the substitution is correct. Since the check is meant to be trusted enough to gate a release, it is worth normalizing the scalar (e.g. compare against a JSON-serialized form) and explicitly handling the non-scalar case as "not checkable" rather than as a failure — otherwise the first false FAIL trains the operator to ignore this check.
- Same block: `shipped = set(strings(emb["INSTANCE"]))` yields dictionary **keys** as well as values, so the new `s == lit` arm can be satisfied by a literal that merely coincides with a key name in the blueprint. The quoted-token arm is unaffected. Restricting `strings()` to values for this particular check (or dropping the bare-equality arm) would remove that path.

**Nits**

- The `test_upgrade2_unit.py` addition to the section 5 pytest block is consistent with the surrounding prose ("Run every file listed…", the constant-vs-tag guard rationale). I checked the remaining "three test files" wordings at lines ~336 and ~638 — those refer specifically to the resource-key-set assertions, which are still three files, so they are correct and do not need updating.
- `SKILL.md` is now ~900 lines. The usual convention for skills is a shorter entry file that links out to detailed procedure files loaded on demand; several of the long narrative subsections here (redaction rationale, CLI version policy, "what validation cannot cover") would be natural split points if this keeps growing.

**Summary**

Doc-only iteration that mostly adds hard-won rationale to the sync workflow and one genuine correctness fix to the literal-currency validation. Nothing blocking; the two suggestions are about the validation snippet still comparing non-string YAML scalars via `str()` and matching against dictionary keys.

.github/skills/sync-aio-bicep-templates/SKILL.md
```diff
@@ -0,0 +1,901 @@
...
+  they can rule on it. Removing a parameter without also removing its readers leaves a reference to a parameter that
+  no longer exists: after deleting one, `grep` its name and remove every remaining occurrence, including entries in
+  `defaultAioConfigurationSettings`.
+
+  Confine that `grep` to the compiled closure — the two top-level templates and the files they transitively import.
+  The deployment repository also contains sibling templates that this workflow never compiles, and a redacted name
+  routinely appears in them. Editing those changes nothing in `template.py`, so leave them alone and say you did;
+  a repository-wide `grep` invites a reviewer to "finish the job" in files that are out of scope.
+
+Never satisfy an `$fxv#N` reference by keeping the blob. Before collapsing one, enumerate **every** reference to it
```
HangyiWang, line 280: is `$fxv#N` always fixed name or it's only a random string

HangyiWang (approved):

.github/skills/sync-aio-bicep-templates/SKILL.md
```diff
@@ -0,0 +1,901 @@
+---
+name: sync-aio-bicep-templates
+description: Generate and safely integrate updated template.py blueprints from the azure-iot-operations-tests deployment repository and an explicit release ref.
```
HangyiWang, line 3: `azure-iot-operations-tests` is not sensitive but yes it's not open source repo

.github/skills/sync-aio-bicep-templates/SKILL.md
```diff
@@ -0,0 +1,901 @@
+---
```
HangyiWang, line 1: Since the SKILL is relatively long, later we can improve the structure by referencing the files in the folder but not the same md file:
```
.github/skills/sync-aio-bicep-templates/
├── SKILL.md                    # Steps, decisions, and links
├── scripts/
│   ├── validate_literals.py    # Executable validation logic
│   └── validate_sync.sh
├── templates/
│   └── wiki-update.md
└── references/
    ├── release-policy.md
    └── design-rationale.md
```

Then  SKILL.md  would say things like:
Run `scripts/validate_literals.py`.
Follow `references/release-policy.md`.
Generate the update using `templates/wiki-update.md`.

This reduces the amount loaded as primary instructions, makes validation executable/testable, and separates required steps from lengthy background explanation.

.github/skills/sync-aio-bicep-templates/SKILL.md
cheatsheet1999: `$fxv` is a fixed Bicep-internal prefix. The `#N` suffix (`$fxv#0`, `$fxv#1`, etc.) is positional and can change when earlier `loadContent` calls are added, so the skill matches `$fxv#0, $fxv#1, $fxv#2....` etc, instead of a fixed index.

It can be removed because Bicep inlines the entire YAML just to read one image tag.

* **Section 4:** replaces the load with the resolved literal and recompiles. Any missed reference fails with `BCP057`.
  [https://github.com/cheatsheet1999/azure-iot-ops-cli-extension/blob/d290925dae21395d85f8882bffca305bc346c98e/.github/skills/sync-aio-bicep-templates/SKILL.md?plain=1#L238](https://github.com/cheatsheet1999/azure-iot-ops-cli-extension/blob/d290925dae21395d85f8882bffca305bc346c98e/.github/skills/sync-aio-bicep-templates/SKILL.md?plain=1#L238)

* **Section 5:** verifies that no `$fxv` remains.
  [https://github.com/cheatsheet1999/azure-iot-ops-cli-extension/blob/d290925dae21395d85f8882bffca305bc346c98e/.github/skills/sync-aio-bicep-templates/SKILL.md?plain=1#L494](https://github.com/cheatsheet1999/azure-iot-ops-cli-extension/blob/d290925dae21395d85f8882bffca305bc346c98e/.github/skills/sync-aio-bicep-templates/SKILL.md?plain=1#L494)

.github/skills/sync-aio-bicep-templates/SKILL.md
cheatsheet1999: Agree with the direction. Moving the wiki payload into templates/ is a clear win.

I’m a bit more cautious about references/, since some non-obvious follow-ups were caught from prose that a fresh agent may not load if it sits behind a link. I’d take that reorg incrementally as a follow-up.

Same for validate_literals.py: making it executable/testable is useful, but it also adds code to maintain, so I’d prefer that as a separate follow-up rather than part of this PR.

.github/skills/sync-aio-bicep-templates/SKILL.md
cheatsheet1999: Right, it has to be named because it's a required input to the workflow: `template.py` is compiled from that repo's Bicep, so the description can't state what this skill does without it. Only the repo name and paths appear in this file, no content from it does. So I'd keep like this if that's ok.

cheatsheet1999:
Synced with @ketkimnaik - I’ll take the remaining items as follow-ups.

PR #910, opened 2026-08-13
feat: refresh iot ops instance api client. Uses 2026-07-01

cheatsheet1999:
## Summary
Updates the IoT Operations management client to API version `2026-07-01`.

The 2607 deployment templates moved to `2026-07-01`, but the vendored management client was still pinned to `2026-03-01`. 

## Changes

- Regenerate the vendored `iotopsmgmt` client at `2026-07-01` (`v20260301` → `v20260701`)
- Add `IoTOpsMgmtApiVersion.V20260701` and make it the default
- Add the missing `< 1.4.0` branch so older instances keep routing to `2026-03-01` while 1.4.x uses `2026-07-01`
- Raise `CLONE_INSTANCE_VERS_MAX` to `1.5.0`

## Verification

Tested against a live 1.4.41 instance:
Before:
image

After:
image

Also:
- `az iot ops clone` now succeeds; all management calls use `2026-07-01` and the emitted template contains no stale API versions
- `--add-insecure-listener` now emits `2026-07-01`


zzcodework:

Reviewing `full diff` (head `ccac692`)

Scope note: 17 files, but ~18.7k of the additions are the regenerated vendored client under `azext_edge/edge/vendor/clients/iotopsmgmt/v20260701/`. I reviewed the hand-written changes (`clone.py`, `common.py`, `az_client.py`, the two vendor shim files, `test_clone_unit.py`) and spot-checked the generated tree for import-surface breakage.

**Blocking**
- None found. The version gating in `clone.py` (`<1.1.0` → 2024-11-01, `<1.2.0` → 2025-04-01, `<1.3.0` → 2025-10-01, new `<1.4.0` → 2026-03-01, else 2026-07-01) matches the assertion chain added in `test_clone_unit.py::_assert_instance_apis`, and both shim files (`iotopsmgmt/__init__.py`, `iotopsmgmt/operations.py`) were repointed to `v20260701` — those were the only two references to `v20260301` outside the version directory itself, so nothing is left dangling. Packaging is fine too (`pyproject.toml` uses `packages.find` with `azext_edge.*`, so the new version package is picked up automatically), and `_help.py` interpolates `CLONE_INSTANCE_VERS_MAX`, so the clone help text tracks the `1.5.0` bump without a separate edit.

**Suggestions**
1. `azext_edge/edge/providers/orchestration/clone.py` (~L1815) — `InstanceApiConfig.v2_enabled` is an allowlist that now has to grow with every API refresh. The failure mode is silent and wrong-direction: forget to append the next version and `v2_enabled` quietly falls back to `False` for the newest instances. Inverting it to exclude the known pre-v2 versions (`V20241101`, `V20250401`) fails safe — a new API version defaults to v2 instead of silently degrading.
2. `azext_edge/edge/util/az_client.py` (L214) — the `DEFAULT_IOTOPS_MGMT_API_VERSION` bump is global; it changes the api-version for every non-clone `az iot ops` call (show/list/update, broker/dataflow providers), not just the clone path. The verification section covers one live 1.4.41 instance in (presumably) public cloud. Worth confirming `2026-07-01` is registered for `Microsoft.IoTOperations` in the sovereign clouds (Fairfax/Mooncake) before this ships, since a missing api-version there surfaces as a hard failure on basic read commands rather than a degraded clone.
3. `clone.py` `get_api_config` — with `v20260301` deleted, requests for older instances (2024-11-01 … 2026-03-01) are now built and deserialized by the 2607-generated code with only the api-version query param differing. That pattern predates this PR, but the blast radius grows each refresh. If the 2603→2607 swagger diff renamed/removed any property on the instance/broker/dataflow types that clone reads for `<1.4.0` instances, the older-version paths would break in a way none of the unit tests would catch (they assert the api-version string, not deserialization). A quick clone run against a 1.3.x instance would close that gap.
4. `azext_edge/tests/edge/orchestration/test_clone_unit.py` (~L1930-1946) — `_assert_instance_apis` re-implements the same `if/elif` chain as production, so it passes for any consistent-but-wrong gating (e.g. if both sides had used `1.3.0` as the new boundary). A static `[(version, expected_iotops_api, expected_adr_api), ...]` table parametrized over representative versions (1.0.34, 1.2.5, 1.3.9, 1.4.0, 1.4.41) would actually pin the contract. Also nothing asserts `DEFAULT_IOTOPS_MGMT_API_VERSION`, which is the change with the widest reach here.

**Nits**
- None. I checked `HISTORY.rst` and `azext_edge/constants.py` deliberately: the prior equivalent refresh (#798, `f36c468`) also added no changelog entry, and `constants.py` is already at `VERSION = 2.8.0` / `AIO_RELEASE = 2607` on `dev`, so neither needs touching in this PR.

Summary: the mechanical parts of the refresh are complete and self-consistent. My only real ask before merge is confirming the new default api-version is available everywhere the extension runs (suggestion 2) and a smoke clone of a `<1.4.0` instance (suggestion 3); the rest is hardening.

_This is an automated review and may be incomplete — please treat findings as input, not as a gate._

azext_edge/tests/edge/orchestration/test_clone_unit.py
```diff
@@ -1041,7 +1041,7 @@ def test_clone_manager_instance_v2(
         instance_name=model_instance_name,
         cluster_name=model_cluster_name,
         add_resources_map=add_resources_map,
-        instance_version="1.3.0",
+        instance_version="1.4.0",
```
digimaun: Moving this to 1.4.0 covers the new api version but drops the only 1.3.x case, and nothing else in the file uses one. That leaves the `< 1.4.0` branch added in clone.py untested, which is also the only path pairing 2026-03-01 with the 2026-07-01 client. Could we parametrize over both instead of moving it?

digimaun (dismissed):

azext_edge/tests/edge/orchestration/test_clone_unit.py
cheatsheet1999: Thanks, done, parametrized over both 1.3.0 and 1.4.0.

This also needed a fix in the mock setup: it registered the instance re-fetch only when not v2_enabled, but the right condition is whether the instance's API differs from the default, those coincided until the default moved to 2026-07-01, so 1.3.0 hit an unregistered URL

zzcodework:

Reviewing `changes since ccac692` (head `c31fe76`) — 1 commit, test-only: `azext_edge/tests/edge/orchestration/test_clone_unit.py` (+6/-3).

No blocking issues. The update addresses the earlier review feedback correctly:

- `test_clone_manager_instance_v2` is now parametrized over both `1.3.0` and `1.4.0`, so the `< 1.4.0` branch in `clone.py` — the only path pairing instance API `2026-03-01` with the `2026-07-01` client — is exercised again rather than dropped.
- The mock fix in `add_instance` is the right condition: keying the extra instance re-fetch off `self.api_config.iotops_mgmt_api not in instance_fetch_by_apis` (with `DEFAULT_IOTOPS_MGMT_API_VERSION` as the baseline) tracks "the instance API differs from the default", which is what the production code actually branches on. The previous `not self.api_config.v2_enabled` only coincided with that while the default was below `2026-07-01`, so it would have started registering the wrong responses as the default moved — good catch to fix it here rather than pinning the test to the old default.

digimaun (approved):

PR #911, opened 2026-08-13
feat: backfill default OPC UA connector template on upgrade

ketkimnaik:
- Starting with the 2608 release, the OPC UA supervisor requires a default connector template resource on the instance, but instances upgraded from earlier releases never got one — leaving the OPC UA connector without its control-plane representation (health and allocated-devices reporting).
- `az iot ops upgrade` now detects when this template is missing and creates it, but only for instances targeting the release where it became required (2608+); earlier upgrades are unaffected. The operation is idempotent — it's skipped when the template already exists.
- Complements the create path (covered separately by the release blueprint), so both fresh installs and upgrades end up consistent regardless of when the instance was first deployed.

---

### General Guidelines

Intent for Production

Basic expectations


HangyiWang (dismissed):

zzcodework:

Reviewing `full diff` (head `6b11dc0`) — 4 files, +407/-6.

Copilot already flagged the `default_opcua_template_name` docstring vs. implementation mismatch and the "any `Microsoft.OpcUa` template satisfies the check" case; I agree with both and won't repeat them. Additional points below.

**Suggestions**

- `azext_edge/edge/providers/orchestration/common.py` — `OPCUA_CONNECTOR_VERSION = "1.4.10"` is a hand-maintained value ("Bump per release from the publishing capture step's `connectors.image.tag`"), and it is *not* derived from anything the release-increment flow touches: sibling PR #912 (2608 increments) updates `constants.py` + both template blueprints + `test_template_unit.py`, but not this constant. If it drifts, `upgrade` will stamp a supervisor image tag that does not match the release the instance was just upgraded to — a silent, cluster-side failure rather than a CLI error. Two mitigations worth considering: (a) add it to whatever checklist/script drives the release version bump, and (b) assert it in a unit test next to the blueprint version assertions so a stale value is at least visible in CI.
- Same file — `OPCUA_CONNECTOR_AIO_MIN_VERSION = "1.2.100"` sits next to `MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE = "1.4.72"`. The two look contradictory at a glance (a template stamped `aioMinVersion 1.2.100` that is only ever created on 1.4.72+). Assuming `1.2.100` is copied verbatim from the product Bicep, a one-line comment saying so would prevent a future reader from "fixing" it to 1.4.72.
- Merge ordering with #912: the gate version `1.4.72` is exactly the `iotOperations` version #912 introduces. If this lands first it is inert until #912 merges; if #912 lands first, upgrades performed in the gap reach 2608 without the backfill and will need a second `az iot ops upgrade` later. Worth landing them close together.
- `azext_edge/edge/providers/orchestration/upgrade2.py` — `_check_opcua_connector_template_needed` swallows `HttpResponseError` and returns `False` at `logger.debug` level. This matches the existing `_check_default_registry_needed` pattern so I'm not asking to change it here, but the consequence is different: a transient/403 list failure silently produces an upgrade that *looks* complete while the OPC UA connector stays without its control-plane representation, which is precisely the symptom this PR exists to fix. A `logger.warning` on that path would make the skip visible without changing behavior.

**Nits**

- The version-gate refactor (`_get_target_ops_version` + `_is_target_version_at_least`) is a clean generalization of `_is_target_version_above_migration_threshold` and preserves the existing override > `init_version_map` > current precedence — no issue, just noting it was checked.

Summary: the design (gate on target version, idempotent check, create after the extension/instance updates and before secretsync migration) is sound, and `assert_operation_order` genuinely asserts the new ordering rather than passing vacuously. My main concern is the hand-maintained `OPCUA_CONNECTOR_VERSION` drifting from the release it is supposed to track.

digimaun:
This is a `feat:`, 2608 hasn't been released yet, and this is a new functional requirement for deployment.

azext_edge/edge/providers/orchestration/common.py
```diff
@@ -429,6 +429,22 @@ class TrustedSigningKeyType(Enum):
 MIN_INSTANCE_VERSION_V2 = "1.2.36"
 MIN_INSTANCE_VERSION_V1_FOR_V2_UPGRADE = "1.1.59"
 MIN_INSTANCE_VERSION_FOR_CM_MIGRATE = "1.2.83"
+# Target version at/above which `az iot ops upgrade` backfills the default OPC UA
+# akriConnectorTemplates resource. The requirement landed in 2608 (1.4.72)
+MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE = "1.4.72"
```
digimaun, line 434: This gate is 1.4.72 but the CLI's own pinned target is 1.4.41 (template.py:1468, `VERSIONS.iotOperations`). With no `--ops-version`, `_get_target_ops_version` resolves the manifest value, so a default upgrade resolves 1.4.41, fails the gate and skips the backfill. Today the path only runs when someone passes `--ops-version 1.4.72` or higher.

I assume that is intentional until the 2608 manifest bump, but nothing ties the two values together, so a release that bumps one and not the other leaves this inert with no signal. Could we derive the gate from the manifest, or add a test that fails once the pinned version crosses it?

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -189,6 +195,34 @@ def _check_default_registry_needed(self) -> bool:
...
+
+        The OPC UA supervisor requires a default ``akriConnectorTemplates`` resource (2608+) that
+        no CLI install path created historically.
+        """
+        try:
+            existing_templates = self.connector_templates.list(
+                instance_name=self.instance_name, resource_group_name=self.resource_group_name
+            )
+            for template in existing_templates:
+                if (template.get("connectorType") or "").lower() == OPCUA_CONNECTOR_ENDPOINT_TYPE.lower():
```
digimaun: This treats any template with connectorType `Microsoft.OpcUa` as satisfying the requirement, but the supervisor adopts by the `azureiotoperationsconnectorforopcua-` prefix, per `OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX` and the docstring on `create_default_opcua_template`. A user created OPC UA template under a different name would skip the backfill and still leave the supervisor without a template it will adopt, which is the case this is meant to fix. Matching on the prefix would line detection up with the adoption mechanism.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
```diff
@@ -888,13 +998,29 @@ def assert_operation_order(target_scenario: UpgradeScenario, upgrade_result: Lis
...
+    if expects_connector_template:
+        ct_index = upgrade_result.index(connector_template_creates[0])
+        if expects_registry:
+            assert ct_index > upgrade_result.index(
+                registry_creates[0]
+            ), "Connector template should be created after registry endpoint"
+        if expects_secretsync:
+            assert ct_index < upgrade_result.index(
+                secretsync_migrations[0]
+            ), "Connector template should be created before secretsync migration"
```
digimaun: This ordering is right, but the existing assertion at line 1029 disagrees with it. That branch expects the registry creation at `upgrade_result[-2]` when registry and secretsync both run, and with the template now sitting between them the second to last is the template. It passes today only because no scenario sets all three, which also means that combination is uncovered. Adding one would catch it.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -207,6 +209,102 @@ def create(
...
+        Build the default OPC UA connector template name.
+
+        The OPC UA supervisor adopts a template by the ``azureiotoperationsconnectorforopcua-``
+        prefix, so the suffix only needs to be stable per instance. A short deterministic hash of
+        the instance name is used (mirroring the product Bicep, which derives a suffix from the
+        instance id).
+        """
+        from ..common import OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX
+
+        suffix = hashlib.sha256((instance_name or "").encode("utf-8")).hexdigest()[:4]
```
digimaun, line 305: The docstring above says the suffix mirrors the product Bicep deriving from the instance id, but this hashes the instance name. Since adoption is by prefix the suffix only needs to be stable per instance, so this is cosmetic, but worth aligning the two.

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: Alongside the name check, worth considering `provisioningState` here too. `_extract_template_summary` already surfaces it (connector_templates.py:904) and `PROVISIONING_STATE_FAILED` exists in common.py, so a template that exists but failed to provision currently counts as satisfying the requirement and never gets repaired on a re-run. `can_upgrade` does this for extensions via `_has_non_success_state`, so there is precedent for treating a non-terminal state as work still to do.

azext_edge/edge/providers/orchestration/common.py
digimaun: PR #913 sets `VERSIONS.iotOperations` to 1.4.72, so this lines up once that merges. Only thing left is merge order, since this stays inert if it lands first.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
digimaun: PR #913 has the Bicep, and it derives the suffix from `uniqueString(resourceId(...))`, so the docstring is right and the implementation is what differs.

azext_edge/edge/providers/orchestration/common.py
ketkimnaik: Kept the gate as the fixed 1.4.72 introduction version (deriving from the manifest would wrongly skip 1.4.72–1.4.7x on future bumps) and added a tripwire test that fails once the pinned manifest crosses the gate.

azext_edge/edge/providers/orchestration/upgrade2.py
ketkimnaik: switched detection to match the azureiotoperationsconnectorforopcua- prefix so it lines up with the supervisor's adoption mechanism.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
ketkimnaik: Replaced the fragile positional check with a relative-order assertion and added a scenario exercising registry + connector template + secretsync together.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Aligned the docstring with the implementation (it hashes the instance name); dropped the inaccurate 'mirrors the Bicep instance id' wording.

zzcodework:

Reviewing `changes since 6b11dc0` (head `b38cf89`) — scoped to the three authored commits (`97dd5fc` docstring, `601f764` prefix matching, `4f97d56` ordering/tripwire tests); the rest of the range is the `dev` merge (2026-07-01 SDK refresh, #913 template regen).

**Blocking**

- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — `test_opcua_connector_template_gate_vs_pinned_manifest` fails on this head. The merge of #913 (`9e867ed`) set `VERSIONS.iotOperations` to `1.4.72` (template.py, `TEMPLATE_BLUEPRINT_INSTANCE` variables block), and `MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE` is also `1.4.72` (common.py:434). The tripwire's condition is `pinned >= gate`, so it now `pytest.fail`s by design. The tripwire did its job — the follow-through it asks for is now due in this same PR: the backfill is live on the default path, so it needs default-path (no `--ops-version`) scenario coverage and the tripwire updated (e.g. re-point it at "gate must not exceed pinned", i.e. fail when `pinned < gate`) or removed.

**Suggestions**

- `azext_edge/edge/providers/orchestration/upgrade2.py:210` — the prefix switch is the right call and lines up with `default_opcua_template_name` / the supervisor's adoption rule; the constant is lowercase so the `.lower().startswith(...)` comparison is safe, and the existing mock (`expected_default_opcua_template`, name `azureiotoperationsconnectorforopcua-abcd`) still matches, so the existing "template exists" scenarios remain valid. Worth a one-line comment there noting the match is deliberately by name prefix (adoption mechanism), not by `connectorType`, so a future reader doesn't "fix" it back to endpoint-type matching.
- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — now that a default upgrade resolves the pinned `1.4.72` through `_get_target_ops_version` (upgrade2.py:986, `init_version_map` path), the scenario helpers still key off `user_kwargs["ops_version"]` only (`expects_connector_template_creation`). Every connector-template scenario passes an explicit `--ops-version`, so the default path — the one real users will hit — is not exercised. Deriving the expected target version the same way the code does would keep these helpers honest as the manifest moves.

**Nits**

- `azext_edge/edge/providers/orchestration/resources/connector_templates.py:298-305` — the docstring was aligned to the implementation (hash of instance name), but per the #913 Bicep the product derives the suffix from `uniqueString(resourceId(...))`. Since adoption is prefix-based the divergence is benign, but a short "suffix scheme intentionally differs from the Bicep; only the prefix is contractual" note would prevent someone later treating name equality as a match.

**Summary**

The three review responses land correctly: prefix-based detection matches the supervisor contract, the ordering assertion is now relative rather than positional and the registry + template + secretsync scenario closes the gap, and the docstring no longer overclaims. The one thing left is that the manifest bump arrived in the same range, so the gate/pinned tripwire is currently red and the default-path coverage it was written to demand is now actually needed.

_Automated review by Azure SRE Agent — may be incomplete; please verify findings independently._

zzcodework:

Reviewing `changes since b38cf89` (head `0b54292`) — commit `Repair failed template and align default-path coverage`, 3 files changed.

Context checks first, so I don't repeat what's already covered: the follow-ups to @digimaun's earlier comments (prefix-based detection, `provisioningState`, docstring alignment, relative-order assertion, tripwire) all look correctly addressed, and the two concerns that could have made this iteration wrong turn out to be fine:

- `_check_opcua_connector_template_needed` reads `template.get("provisioningState")` (flat) while the new test seeds `template["properties"]["provisioningState"]` (nested). That is **not** a bug — `ConnectorTemplates.list` passes each raw resource through `_extract_template_summary`, which flattens `properties.provisioningState` to the top level, so the test exercises the real path.
- The tripwire flip from `pytest.fail(...)` to `assert PINNED_IOTOPS_VERSION >= MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE` would have hard-failed CI had it landed before the manifest bump, but `dev` already pins `iotOperations: 1.4.72` (#913 is in), so it passes and the earlier merge-order concern is resolved.

**Blocking**

None.

**Suggestions**

1. `azext_edge/edge/providers/orchestration/upgrade2.py` (`_check_opcua_connector_template_needed`, ~L207-217, together with `_create_default_opcua_connector_template`) — the repair path *detects* by prefix but *repairs* by the CLI's own name. `default_opcua_template_name` uses `sha256(instance_name)[:4]`, while (per the updated docstring and #913) the product Bicep derives its suffix from `uniqueString(resourceId(...))`. So when the failed template came from the product Bicep, this does not repair it: it PUTs a second, differently-named template and leaves the `Failed` one behind. Two consequences worth confirming are intended: (a) the instance permanently carries an orphaned `Failed` `azureiotoperationsconnectorforopcua-*` resource that no CLI path cleans up, and (b) on the next upgrade the loop hits the newly-succeeded template and returns `False`, so the orphan becomes invisible from then on. If repair-in-place is the goal, reusing the matched template's actual `name` for the create-or-update (falling back to `default_opcua_template_name` when nothing matched) side-steps the suffix divergence entirely.

2. Same block, L213 — the check is `== Succeeded`, so *every* non-success state counts as "needs repair", including in-flight ones (`Accepted`, `Provisioning`, `Updating`) and `Deleting`. An upgrade run while a template deployment is still in flight would issue a competing create-or-update, and a `Deleting` template would be recreated under the user's feet. Gating on terminal non-success only (`Failed`/`Canceled` — both constants exist in `common.py`) is more precise than mirroring the extension-side `!= Succeeded` convention here. If mirroring was deliberate, a one-line comment saying so would help.

**Nits**

3. `azext_edge/edge/providers/orchestration/upgrade2.py` L200 — the docstring still reads "Return True if the default OPC UA connector template is missing"; it now also returns `True` when the template exists but is unhealthy. Same for the pre-upgrade summary row (~L657), which unconditionally renders `Create default OPC UA connector template` even in the repair case — "Create or repair" would set the right expectation in that output.

4. `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — the new "Repair when existing template failed to provision" scenario pins `--ops-version` to `MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE`, while the new default-path scenario covers the no-`--ops-version` route for the *create* case only. A repair scenario on the default path (or simply dropping the explicit `ops_version` here, now that the pinned manifest satisfies the gate) would keep the repair branch covered if the gate/manifest relationship shifts again.

**Summary**

The incremental change is sound and addresses the earlier review round correctly; the flattening and merge-order risks I specifically checked both come out clean. The one item worth an answer before merge is #1 — whether "repair" is meant to fix the existing template or merely guarantee a healthy one exists, since today it does the latter and leaves the failed resource behind.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -189,6 +196,38 @@ def _check_default_registry_needed(self) -> bool:
...
+                    logger.debug("Default OPC UA connector template already exists.")
+                    return False
+                logger.debug("Default OPC UA connector template exists but has a non-success provisioning state.")
+            return True
+        except HttpResponseError as e:
+            logger.debug(f"Error checking OPC UA connector template: {e}")
+            return False
+
+    def _create_default_opcua_connector_template(self, headers: dict) -> dict:
+        return self.connector_templates.create_default_opcua_template(
```
digimaun, line 233: The repair path does not repair. `create_default_opcua_template` falls back to `default_opcua_template_name`, which hashes the instance name with sha256, while the product template builds its name from `uniqueString(resourceId(...))` in `template.py`. Different hash over a different input, so the two will not agree.

When the check finds a template that ARM created and it is not in a Succeeded state, this creates a second one under the CLI name rather than repairing what is already there, and the failed resource stays behind.

`create_default_opcua_template` already accepts `template_name` and nothing passes it today. Could we return the name we discovered in `_check_opcua_connector_template_needed` and pass it through here? The `uniqueString` algorithm is not documented, so matching the ARM name in Python would mean relying on reverse engineered behavior. Reusing the name we already found avoids that entirely.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
```diff
@@ -83,8 +90,32 @@
     },
 }
 
+expected_default_opcua_template = {
```
digimaun, line 94: This is the only template shape the mocks produce, and it carries both the `azureiotoperationsconnectorforopcua-` name and `endpointType: Microsoft.OpcUa`, so the new name check and the old `connectorType` check can never disagree here. Reverting detection back to `connectorType` keeps the whole file passing.

The case the name check exists for is a template with the OPC UA endpoint type under a name the supervisor will not adopt. Could we add an aux knob for the existing template name and a scenario that uses one without the prefix, so the backfill is expected to fire?

azext_edge/edge/providers/orchestration/common.py
```diff
@@ -429,6 +429,22 @@ class TrustedSigningKeyType(Enum):
...
+# is the supervisor image.
+OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX = "azureiotoperationsconnectorforopcua-"
+OPCUA_CONNECTOR_ENDPOINT_TYPE = "Microsoft.OpcUa"
+OPCUA_CONNECTOR_REGISTRY = "mcr.microsoft.com"
+OPCUA_CONNECTOR_SUPERVISOR_IMAGE_NAME = "azureiotoperations/aio-connectors/supervisor"
+OPCUA_CONNECTOR_METADATA_REF = "mcr.microsoft.com/azureiotoperations/aio-connectors/opcua-metadata"
+OPCUA_CONNECTOR_AIO_MIN_VERSION = "1.2.100"
+# Connectors component image tag stamped on the template (distinct from the AIO extension
+# version). Bump per release from the publishing capture step's connectors.image.tag.
+OPCUA_CONNECTOR_VERSION = "1.4.10"
```
digimaun, line 447: `template.py` carries its own copy of this tag inside the `OPCUA_CONNECTOR_VERSION` ARM variable. Nothing couples them, so a release that regenerates the template without updating this constant leaves `az iot ops create` stamping one tag and `az iot ops upgrade` backfilling another on the same release.

The tests cannot catch that today, because `test_upgrade2_unit.py:833` and `:2424` compare the stamped payload against this same constant, so changing it moves both sides. Asserting that this value appears in the `template.py` variable would fail on drift.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -189,6 +196,38 @@ def _check_default_registry_needed(self) -> bool:
...
+        """
+        try:
+            existing_templates = self.connector_templates.list(
+                instance_name=self.instance_name, resource_group_name=self.resource_group_name
+            )
+            for template in existing_templates:
+                if not (template.get("name") or "").lower().startswith(OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX):
+                    continue
+                # A template that exists but did not provision successfully still needs repair.
+                if (template.get("provisioningState") or "").lower() == PROVISIONING_STATE_SUCCESS.lower():
```
digimaun: This treats every state other than Succeeded as needing repair, which includes `Accepted`, `Updating` and `Deleting`. An upgrade that runs while a template create is still in flight would read a transient state and start a second create.

Repairing only terminal failures would avoid that, and `PROVISIONING_STATE_FAILED` already exists in `common.py`.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -189,6 +196,38 @@ def _check_default_registry_needed(self) -> bool:
...
             return False
 
+    def _check_opcua_connector_template_needed(self) -> bool:
+        """Return True if the default OPC UA connector template is missing.
+
+        The OPC UA supervisor requires a default ``akriConnectorTemplates`` resource (2608+) that
+        no CLI install path created historically.
+        """
+        try:
+            existing_templates = self.connector_templates.list(
```
digimaun, line 213: `ConnectorTemplates.list()` returns the `_extract_template_summary` projection rather than ARM resources, so this depends on that summary continuing to surface `name` and `provisioningState` at the top level. It does today, so the check is correct.

Worth noting because a later trim of that summary would make `provisioningState` read as None, which this treats as needing repair, so every upgrade would create another template. `RegistryEndpoints.list()` next door returns raw resources, so the two sibling checks read different shapes.

ketkimnaik: Since repair now triggers only on a terminal failure, an unknown state is a no-op rather than a spurious creation, removing that risk.

azext_edge/edge/providers/orchestration/upgrade2.py
ketkimnaik: I narrowed repair to terminal failures only, so in-flight provisioning states are left alone; added a transient-state scenario.

azext_edge/edge/providers/orchestration/common.py
ketkimnaik: Added a test that fails if the upgrade tag and the template tag drift apart, so create and upgrade always stamp the same tag on a release.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
ketkimnaik: Added a scenario with an OPC UA template under a non-adopt name so the backfill is expected to fire, closing that coverage gap.

azext_edge/edge/providers/orchestration/upgrade2.py
ketkimnaik: Fixed — the upgrade now repairs the existing failed template in place instead of creating a second one and leaving the failed resource behind.

zzcodework:

Reviewing `changes since 0b54292` (head `3945a5a4`) — one commit, "Repair in place and guard version drift", touching `azext_edge/edge/providers/orchestration/upgrade2.py` and `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py`.

The three responses to @digimaun's round look right: repair is now gated on `PROVISIONING_STATE_FAILED` only (transient `Accepted`/`Updating`/`Deleting` fall through to the "leave it alone" branch), `_opcua_template_name_to_repair` is reset at the top of every check so a stale name cannot leak into a later call, and `create_default_opcua_template` already falls back to `default_opcua_template_name()` when the name is `None`, so the non-repair path is unchanged. Those are not repeated below.

**Blocking**

- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — the new in-place repair behavior is not actually asserted; the "OPC UA Template: Repair when existing template failed to provision" scenario passes for the wrong reason. It only sets `opcua_connector_template_provisioning_state=PROVISIONING_STATE_FAILED` and leaves the name at the default `azureiotoperationsconnectorforopcua-abcd`, while the PUT mock is registered against `re.compile(... + r"/azureiotoperationsconnectorforopcua-[a-z0-9]+")` and `connector_template_create_callback` asserts only on the body (endpoint type, image name, tag) — never on the name. So a create under the CLI-derived name `azureiotoperationsconnectorforopcua-<sha256(instance)[:4]>` matches the same mock and satisfies the same assertions. Reverting `template_name=self._opcua_template_name_to_repair` in `_create_default_opcua_connector_template` would leave this suite green, which is exactly the regression this commit is meant to prevent (a second template plus an orphaned failed one). Suggest giving the failed template a distinguishable name via the new knob (e.g. `opcua_connector_template_name="azureiotoperationsconnectorforopcua-repairme"`, deliberately not the hash the CLI would derive) and asserting in the callback that `request.path_url.split("?")[0].split("/")[-1]` equals it for that scenario.

**Suggestions**

- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py`, `test_opcua_connector_version_matches_template_tag` — the tripwire is a good addition, but `re.findall(r"'([^']*)'", opcua_var)` then taking `[-1]` assumes the coalesce fallback is the last quoted literal in the expression. Today the variable is `[coalesce(tryGet(tryGet(parameters('advancedConfig'), 'connectors'), 'version'), '1.4.10')]`, so it works; if a future template regeneration adds another quoted argument after the fallback (or wraps the expression), the test fails with a confusing "tag mismatch" message rather than the real cause. Anchoring on the coalesce fallback — e.g. `re.search(r"'([^']+)'\)\]$", opcua_var)`, or simply asserting `f"'{OPCUA_CONNECTOR_VERSION}'" in opcua_var` — makes the guard fail only for genuine drift.

- `azext_edge/edge/providers/orchestration/upgrade2.py`, `_check_opcua_connector_template_needed` — repair reuses the name but not the payload: `create_default_opcua_template` issues `begin_create_or_update` with a fully-built default `template_resource`, so any user customization on that resource (replicas, log level, a pinned `connectorMetadataRef`) is replaced with CLI defaults. That is fine for a template the CLI itself created, but the match is on the `azureiotoperationsconnectorforopcua-` prefix, which a user-authored template can also carry, and a user template failing to provision for a user-specific reason is exactly the case that lands here. The only trace is `logger.debug`, invisible in a normal upgrade run. Consider logging the repair at `info`/`warning` with the template name so the overwrite of an existing resource is visible in the upgrade output.

**Nits**

- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py`, `test_opcua_connector_version_matches_template_tag` — `import re` inside the function is redundant; `re` is already imported at module scope.

Summary: the production change is correct and addresses the earlier feedback cleanly. The one thing worth fixing before merge is test coverage — the repair-in-place name reuse, the whole point of this commit, is not observable to the current assertions.

_Automated review by Azure SRE Agent; it may be incomplete or miss context — please treat findings as input, not verdicts._

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: The production change is right, and reusing the discovered name is what I was after. The test still does not prove it though. The repair scenario leaves the existing template at the default `azureiotoperationsconnectorforopcua-abcd` name, the PUT mock matches `azureiotoperationsconnectorforopcua-[a-z0-9]+`, and the callback only asserts on body fields. Dropping `template_name=self._opcua_template_name_to_repair` would leave the suite green, so the one behavior this commit adds is the one thing not covered.

Could we give the failed template a name the CLI would never derive, using the knob you just added, and assert the PUT lands on it? That turns the whole thing into a real regression guard.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -597,6 +657,21 @@ def render_upgrade_table(upgrade_state: "ClusterUpgradeState"):  # noqa: C901
             except Exception as e:
                 logger.debug(f"Error adding registry row: {e}")
 
+        # Add OPC UA connector template row if needed
+        if upgrade_state.connector_template_needed:
+            try:
+                table.add_row(
+                    "opc-ua connector template",
+                    "[dim]Not configured[/dim]",
```
digimaun: `render_upgrade_table` runs immediately before `should_continue_prompt`, so this table is what the user consents from. In the repair case the row is wrong on both counts. The template exists and failed, so it is not `Not configured`, and it is being overwritten rather than `Created`.

That matters because repair reuses the name but not the payload. `create_default_opcua_template` builds a full default resource and PUTs it, so `replicas`, `logLevel` or a pinned `connectorMetadataRef` are replaced with CLI defaults. Those are all settable through `az iot ops connector template update`, so a customized default template that later fails lands here and loses its settings without saying so.

Could the row reflect the repair case, showing the failed state and naming the template being replaced? Then someone with a customized template sees what is about to happen while they can still answer no.

Whether repair should preserve those settings rather than replace them is a larger question and probably its own change.

azext_edge/edge/providers/orchestration/upgrade2.py
ketkimnaik: Done

azext_edge/edge/providers/orchestration/upgrade2.py
ketkimnaik: Made the repair row honest — it now shows the failed template's name and a "Replaced / settings reset to defaults" action instead of "Not configured → Created", so a customer with a customized default template sees what's about to happen and can still answer no. Left preserving-settings-on-repair as a separate follow-up as you suggested.

digimaun (approved):

zzcodework:

Reviewing `changes since 3945a5a4` (head `49a6326d`) — commit `Reflect repair in upgrade table and guard name reuse`, which addresses the two open review points (the test did not prove name reuse; the table row misrepresented the repair case).

**Blocking**

None. Both fixes look correct.

- `_check_opcua_connector_template_needed` now returns `Tuple[bool, Optional[str]]` and `ClusterUpgradeState._check_connector_template_needed` unpacks it into `connector_template_repair_name`. `Tuple`/`Optional` are already imported in `azext_edge/edge/providers/orchestration/upgrade2.py`, and `connector_template_repair_name` is initialized in `__init__` before the check runs, so there is no attribute-ordering hazard.
- The test is now a real regression guard: the PUT callback derives `response_body["name"]` from `request.path_url`, and the repair scenario uses `azureiotoperationsconnectorforopcua-repairtarget`, a name the sha256 4-hex derivation can never produce. Dropping `template_name=self._opcua_template_name_to_repair` would now fail `assert_connector_template_result`. The name also still matches the mocked PUT URL regex `azureiotoperationsconnectorforopcua-[a-z0-9]+`, so the route is genuinely exercised rather than silently unmatched.

**Suggestions**

- `azext_edge/edge/providers/orchestration/upgrade2.py` — the new repair branch in `render_upgrade_table` has no coverage; there are currently no tests exercising `render_upgrade_table` anywhere under `azext_edge/tests/`, so the row the user consents from is verified only by eye. Since that row now carries the "existing template settings are reset to defaults" warning, a small test that renders with and without `connector_template_repair_name` and asserts the text would lock it in. Standing up table tests from scratch may be out of scope here — reasonable as a follow-up alongside the deferred "preserve settings on repair" item.
- `azext_edge/edge/providers/orchestration/upgrade2.py` — the repair name is now carried in two places: `IoTOperationsUpgrade._opcua_template_name_to_repair` (set as a side effect of the check, read by `_create_default_opcua_connector_template`) and `ClusterUpgradeState.connector_template_repair_name` (returned explicitly). It is correct today because the check runs exactly once during `ClusterUpgradeState.__init__`, but the side-effect attribute is now redundant. Threading the state's `connector_template_repair_name` into the create call would remove any chance of the two drifting if the check is ever re-run.

**Nits**

- `azext_edge/edge/providers/orchestration/upgrade2.py` — `getattr(upgrade_state, "connector_template_repair_name", None)` in `render_upgrade_table`; the attribute is unconditionally set in `ClusterUpgradeState.__init__`, so a plain attribute access reads clearer and would surface a real regression instead of silently falling back to the "Not configured → Created" row.
- `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — the `is_repair` predicate inside `assert_connector_template_result` re-derives repair-ness from `aux_kwargs` with logic that partially mirrors `expects_connector_template_creation`. A shared helper would keep the two from drifting the next time the repair conditions change.

Summary: the incremental commit does what it claims — the repair path is now genuinely covered, and the pre-prompt table honestly reports the failed template name with "Replaced / settings reset to defaults". Remaining notes are non-blocking.

PR #913, opened 2026-08-13
build: generate templates for 2608 release

HangyiWang:
## Summary

- update AIO, dependency extension, CLI, and release versions for 2608
- regenerate the public embedded templates from `azure-iot-operations-tests` release source `394a2e3aa54983fe95e11679cc338d16506056ef` using Bicep `0.46.1`
- include the public OPC UA connector template and Data Flows trust settings
- exclude the private-preview GDS Manager parameter/configuration
- use the explicit 2608 connector version (`1.4.10`) rather than embedding the complete Helm values file

## Release values

| Field | Value |
|---|---|
| AIO | `1.4.72` / `integration` |
| CLI | `2.9.0a1` |
| Connectors | `1.4.10` |
| Cert Manager | `1.0.0` / `stable` |
| Secret Store | `1.5.2` / `stable` |
| Bicep | `0.46.1` |


ketkimnaik (dismissed):

azext_edge/edge/providers/orchestration/template.py
```diff
@@ -1497,9 +2175,12 @@ def copy(self) -> "TemplateBlueprint":
                 "connectors.values.mqttBroker.serviceAccountTokenAudience": "[variables('MQTT_SETTINGS').serviceAccountAudience]",
                 "connectors.values.securityPki.applicationUri": "[format('urn:microsoft.com:aio:opc:ua:broker:{0}', variables('AIO_EXTENSION_SUFFIX'))]",
                 "connectors.values.securityPki.subjectName": "[format('CN=aio-opc-opcuabroker-{0}', variables('AIO_EXTENSION_SUFFIX'))]",
+                "connectors.values.gdsManager.enabled": "[if(parameters('enableGdsManager'), 'true', 'false')]",
```
cheatsheet1999: This line and line 1475 are the ones Paymaun mentioned.

zzcodework:

Reviewing `full diff` (head `97808f6`)

**Blocking**

1. **Unit tests are red on all 12 matrix jobs (20 failures) — the new instance resource was only added to one of three expected-resource lists.**
   Adding `opcUaConnectorTemplate` to `TEMPLATE_BLUEPRINT_INSTANCE` (`azext_edge/edge/providers/orchestration/template.py`, resources block ~L2357) changes the instance resource count from 10 to 11. `azext_edge/tests/edge/orchestration/test_template_unit.py` L47 was updated, but two other hardcoded sets were not:
   - `azext_edge/tests/edge/orchestration/test_work_unit.py` — `get_expected_keys_for()` (~L1535-1538) `resource_keys` set; the exact-count assert at L1560 fails with `assert 11 == 10` (`test_iot_ops_create`).
   - `azext_edge/tests/edge/orchestration/test_targets_unit.py` — `test_instance_phases` parametrization (~L878-903), both the `InstancePhase.RESOURCES` set and the "no phase" set.

   Confirmed from the CI log for `test / Unit test 3.12 - ubuntu-24.04`:
   ```
   >       assert len(template["resources"]) == len(expected_keys)
   E       assert 11 == 10
   azext_edge/tests/edge/orchestration/test_work_unit.py:1560: AssertionError
   ...
   FAILED azext_edge/tests/edge/orchestration/test_targets_unit.py::test_instance_phases
   FAILED azext_edge/tests/edge/orchestration/test_work_unit.py::test_iot_ops_create
   20 failed, 6580 passed
   ```
   Note this is a real signal, not just test bookkeeping: `opcUaConnectorTemplate` falls into the `RESOURCES` phase in `targets.py` (it is not in `PHASE_KEY_MAP`, so it survives `del_if_not_in` for `EXT`/`INSTANCE` and is deployed in the last phase). That is very likely the intent given `dependsOn: ["aioInstance", "customLocation"]`, but it should be an explicit decision recorded in the phase tests rather than a side effect.

**Suggestions**

2. **Template name derivation diverges from the upgrade-path implementation in #911 — the same instance can end up with two OPC UA connector templates.**
   Here the name is
   `format('azureiotoperationsconnectorforopcua-{0}', substring(uniqueString(resourceId('Microsoft.IoTOperations/instances', <instanceName>)), 0, 4))`
   (`template.py`, `opcUaConnectorTemplateName` variable ~L2198), while open PR #911 (`fix: backfill default OPC UA connector template on upgrade`) computes `default_opcua_template_name()` as `sha256(instance_name)[:4]` with the same prefix. Those two suffixes will not match for any instance, so a create via this template followed by an upgrade via #911 would produce two `akriConnectorTemplates` resources with the OPC UA prefix. Since the supervisor adopts by prefix, that is probably not benign. Whichever PR merges second should reconcile — either #911 discovers existing templates by prefix before creating (it does appear to enumerate templates, worth double-checking it matches on the prefix and not on the exact computed name), or both sides agree on one derivation.

3. **Connector image tag source also diverges from #911.** This PR derives it dynamically:
   `"OPCUA_CONNECTOR_VERSION": "[coalesce(tryGet(tryGet(parameters('advancedConfig'), 'connectors'), 'version'), variables('AIO_EXTENSION_VALUES').connectors.image.tag)]"`
   while #911 pins `OPCUA_CONNECTOR_VERSION = "1.4.10"` as a Python constant. A hardcoded CLI-side constant will silently drift from the AIO extension's connector image tag on every release train bump, whereas this template will not. Recommend the upgrade path read the same source rather than carrying its own constant.

4. **New `advancedConfig.connectors.version` knob has no CLI surface or doc.** The definition is added to both enablement and instance templates, but nothing in `azext_edge/edge/params.py` / `_help.py` appears to populate or document it. If it is intentionally template-only (internal/testing escape hatch, like other `advancedConfig` fields), no action needed — just confirming it is deliberate and not a dropped parameter wiring.

**Nits**

- `aioMetadata.aioMinVersion` is pinned to `1.2.100` while the release being shipped is `1.4.72`; assuming that is the true minimum from the product Bicep, no change needed, but it reads oddly next to the `1.4.72` bump and is easy to mistake for a stale value.

**Summary**

The version/template regeneration itself looks consistent (`constants.py` `2.9.0a1` / `AIO_RELEASE 2608`, bicep generator `0.46.1.21595`, `certManager 1.0.0`, `secretStore 1.5.2`, `iotOperations 1.4.72`/`integration`, and the matching `test_template_unit.py` `EXTENSION_CONFIGS` update). The `integration` train plus an `a1` CLI version matches the established pre-release pattern (flipped to `stable` in a follow-up, cf. #891, #868). The one thing that must be fixed before merge is the two un-updated expected-resource sets causing the red unit tests; the coordination items with #911 are the higher-value design concern. I have not repeated the `gdsManager` / trust-settings feedback already raised on the diff.

ketkimnaik:
As discussed remove: enableGdsManager lines as well

digimaun (approved):

digimaun:
Can we ensure `https://github.com/Azure/azure-iot-ops-cli-extension/blob/dev/tools/template_optimizer.py` is up to date in a next PR

digimaun (approved):

PR #914, opened 2026-08-14
fix: apply custom instance name to OPC UA connector template resource

ketkimnaik:
- Fixes az iot ops create -n <instance> failing with ParentResourceNotFound — the opcUaConnectorTemplate resource wasn't renamed to the custom instance, so its parent pointed at the default aio-<hash> instance.
- Rewrites the connector template's parent to the custom instance name (preserving the deploy-time name variable) and adds a unit-test assertion.

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/providers/orchestration/targets.py
+++ b/azext_edge/edge/providers/orchestration/targets.py
@@ -275,6 +275,7 @@ class InitTargets:
         dataflow_profile = template.get_resource_by_key("dataflowProfile")
         dataflow_endpoint = template.get_resource_by_key("dataflowEndpoint")
         artifact_registry_endpoint = template.get_resource_by_key("artifactRegistryEndpoint")
+        opcua_connector_template = template.get_resource_by_key("opcUaConnectorTemplate")
 
         instance["properties"] = get_default_instance_config(
             description=self.instance_description,
@@ -289,6 +290,9 @@ class InitTargets:
             dataflow_profile["name"] = f"{self.instance_name}/{DEFAULT_DATAFLOW_PROFILE}"
             dataflow_endpoint["name"] = f"{self.instance_name}/{DEFAULT_DATAFLOW_ENDPOINT}"
             artifact_registry_endpoint["name"] = f"{self.instance_name}/{DEFAULT_ARTIFACT_REGISTRY}"
+            opcua_connector_template["name"] = (
+                f"[format('{self.instance_name}/{{0}}', variables('opcUaConnectorTemplateName'))]"
+            )
 
             template.content["outputs"]["aio"]["value"]["name"] = self.instance_name
 
--- a/azext_edge/edge/providers/orchestration/targets.py
+++ b/azext_edge/edge/providers/orchestration/targets.py
@@ -27,6 +27,7 @@ from ..orchestration.common import (
     TRUST_SETTING_KEYS,
 )
 from ..orchestration.resources.brokers import Brokers
+from ..orchestration.resources.connector_templates import ConnectorTemplates
 from ..orchestration.resources.instances import parse_feature_kvp_nargs
 from .template import (
     TEMPLATE_BLUEPRINT_ENABLEMENT,
@@ -291,7 +292,7 @@ class InitTargets:
             dataflow_endpoint["name"] = f"{self.instance_name}/{DEFAULT_DATAFLOW_ENDPOINT}"
             artifact_registry_endpoint["name"] = f"{self.instance_name}/{DEFAULT_ARTIFACT_REGISTRY}"
             opcua_connector_template["name"] = (
-                f"[format('{self.instance_name}/{{0}}', variables('opcUaConnectorTemplateName'))]"
+                f"{self.instance_name}/{ConnectorTemplates.default_opcua_template_name(self.instance_name)}"
             )
 
             template.content["outputs"]["aio"]["value"]["name"] = self.instance_name
--- a/azext_edge/tests/edge/orchestration/test_work_unit.py
+++ b/azext_edge/tests/edge/orchestration/test_work_unit.py
@@ -1656,3 +1656,6 @@ def assert_instance_deployment_body(body_str: str, target_scenario: dict, phase:
         assert resources["dataflowProfile"]["name"] == f"{instance_name_lowered}/{DEFAULT_DATAFLOW_PROFILE}"
         assert resources["dataflowEndpoint"]["name"] == f"{instance_name_lowered}/{DEFAULT_DATAFLOW_ENDPOINT}"
         assert resources["artifactRegistryEndpoint"]["name"] == f"{instance_name_lowered}/{DEFAULT_ARTIFACT_REGISTRY}"
+        assert resources["opcUaConnectorTemplate"]["name"] == (
+            f"[format('{instance_name_lowered}/{{0}}', variables('opcUaConnectorTemplateName'))]"
+        )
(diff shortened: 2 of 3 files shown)


digimaun (dismissed):

azext_edge/edge/providers/orchestration/targets.py
cheatsheet1999, line 261: Non blocking, the current fix should be able to pass IT workflow

However, would adding  `"aioInstanceName": self.instance_name`  to `param_to_target`  be sufficient here? It should correctly resolve both the parent instance name and connector template suffix.

cheatsheet1999 (dismissed):

zzcodework:

Reviewing `full diff` (head `2472346`), base `dev`. 2 files, +7/-0.

The fix is narrowly correct — the `opcUaConnectorTemplate` parent segment now points at the custom instance, which is what caused `ParentResourceNotFound`. Below is supporting detail for the direction @cheatsheet1999 already raised, plus one repo-convention gap. I'm not repeating that suggestion, only confirming it is viable and noting a second effect of the current approach.

**Suggestions**

1. `azext_edge/edge/providers/orchestration/targets.py` (the new `opcua_connector_template["name"]` assignment) — the rename fixes the *parent* segment but not the *child* segment. The child comes from `variables('opcUaConnectorTemplateName')`, defined in `azext_edge/edge/providers/orchestration/template.py` (line 1531) as:

   `[format('azureiotoperationsconnectorforopcua-{0}', substring(uniqueString(resourceId('Microsoft.IoTOperations/instances', coalesce(parameters('aioInstanceName'), format('aio-{0}', variables('HASH'))))), 0, 4))]`

   Since `aioInstanceName` is never supplied, that `uniqueString` is still computed over the **default** `aio-<HASH>` instance resource id. So a custom-named instance gets a connector template whose 4-char suffix is keyed off an instance that does not exist. It deploys (the suffix is arbitrary and deterministic), so this is not a second `ParentResourceNotFound`, but the resulting name diverges from what the ARM template intends and from what any other tooling deriving the same expression would compute.

   Confirming the parameter route works: `aioInstanceName` **is** a declared parameter on the instance blueprint (`template.py`, parameters block ~line 1467, `{"type": "string", "nullable": True}`), so passing `"aioInstanceName": self.instance_name` through `param_to_target` in `get_ops_instance_template` will actually be emitted — `_handle_apply_targets` (`targets.py`, the `if param in built_in_template_params` filter) silently drops params that are not declared, which would have made the suggestion a no-op otherwise. It isn't. That single parameter resolves line 1531, line 1696, and the `aio` output at line 1738 in one place, and removes the need to embed a hand-built ARM `format(...)` expression in Python.

2. `HISTORY.rst` — no changelog entry. The `unreleased` section still reads `* Initial baseline.`, and the PR checklist asks for an entry for user-facing changes. This is a user-visible `az iot ops create -n <instance>` failure fix and is worth one line.

**Nits**

3. `azext_edge/tests/edge/orchestration/test_work_unit.py` — the new assertion compares against the same literal ARM expression the implementation builds, so it restates the code rather than pinning behavior. It will pass either way and will need rewriting if the `aioInstanceName` parameter approach is adopted. Asserting the parent segment (that the name starts with `f"{instance_name_lowered}/"`) would be more durable, and there is no assertion covering the default (no `-n`) path.

**Summary**: correct as far as it goes and low risk; the parameterized approach already suggested in review is the better landing spot because it also covers the connector-template suffix, and `aioInstanceName` is confirmed to be a real declared parameter. Add a `HISTORY.rst` entry before merge.

_This is an automated review and may be incomplete; please treat findings as input, not verdicts._

azext_edge/edge/providers/orchestration/targets.py
ketkimnaik: Good call — passing aioInstanceName would resolve both the parent and suffix via ARM, but since every other instance-child resource here is already renamed via explicit overrides, I kept this consistent with that pattern; happy to collapse them all onto the param as a follow-up cleanup.

digimaun: The rename in `targets.py` works the way it does because every instance child has a literal segment, `format('{0}/{1}', coalesce(...), 'default')`. Python resolves that fully, so the deployment body carries `myinstance/default` rather than an expression, and `assert_instance_deployment_body` asserts the exact value ARM receives. That is the only fast check on the custom name path, which is otherwise exercised just by the nightly integration run.

Passing `aioInstanceName` resolves correctly at deploy time, but collapsing them all onto the param would make every name in the body an expression, so those assertions drop from checking the name is right to checking the expression is unchanged. That is the coverage a missing rename line defeats, which is what happened here.

The connector template does not have to be the exception either. `ConnectorTemplates.default_opcua_template_name` is already a staticmethod that derives a stable suffix from the instance name, and it is what the upgrade backfill uses. Calling it here would give a literal name like the siblings, so the assertion at `test_work_unit.py:1659` could check the resolved value, and a create with an explicit name would produce the same name the backfill would. `variables('opcUaConnectorTemplateName')` is referenced only by this resource, so nothing else depends on it.

ketkimnaik: I have made the changes.

cheatsheet1999: Thanks for the clarification.

cheatsheet1999 (dismissed):

cheatsheet1999 (approved):

zzcodework:

Reviewing `changes since 2472346` (head `cdfd56d`), base `dev`. 2 commits, 3 files, +12/-5.

The new revision adopts the direction @digimaun laid out in the `targets.py` thread: `opcUaConnectorTemplate["name"]` is now fully resolved in Python via `ConnectorTemplates.default_opcua_template_name`, so the deployment body carries a literal `myinstance/azureiotoperationsconnectorforopcua-<suffix>` like its sibling instance children, and `assert_instance_deployment_body` can assert the value ARM actually receives. That resolves both the parent segment (the original `ParentResourceNotFound`) and the suffix, and it keeps the custom-name path checked by the fast unit test rather than only by the nightly integration run. I'm not re-raising that thread; below are two points it doesn't cover plus a note on the integration-test change.

**Suggestions**

1. `azext_edge/tests/edge/orchestration/test_work_unit.py` (the new `expected_opcua_template` assertion) — the expected value is computed by calling the same production helper the code under test uses, so the suffix half of the assertion is self-referential: if `default_opcua_template_name` ever regressed to returning an ARM expression (exactly the failure this PR fixes), the test would still pass. Adding one independent check next to it restores that coverage cheaply, e.g. assert the resolved name starts with `OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX` and contains no `[` / `variables(` expression. That pins "the body carries a literal name", which is the property that broke.

2. `azext_edge/edge/providers/orchestration/template.py` (`opcUaConnectorTemplateName`, line ~1531) is now the *second* source of truth for this name and is bypassed entirely on the custom-instance path, while still driving the default-instance path. The two derive their 4-char suffix differently (`uniqueString(resourceId(...))` vs `sha256(instance_name)[:4]`), which is fine given adoption is prefix-based, but the shared literal prefix `azureiotoperationsconnectorforopcua-` now lives in both the template JSON and `OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX` with nothing asserting they agree. A small unit assertion that the template variable string contains the constant would catch prefix drift, which would otherwise surface only as a silently un-adopted connector template.

**Nits**

- `azext_edge/tests/edge/support/create_bundle_int/test_akri_int.py` — the change reads correctly: `AKRI_BUNDLE_PREFIXES` is passed only to `check_workload_resource_files`, so the OPC UA connector lands in the "accepted extra" bucket in `_compare_support_bundle_names`, while `check_cluster_label_coverage` and the pre-bundle `kubectl` collection still use the label-scoped `AKRI_PREFIXES`. The explanatory comment is a good addition. No change requested.

Summary: the revision matches the agreed approach and the integration-test carve-out is scoped correctly; the only gap I'd close before merge is the tautological assertion in `test_work_unit.py`, since that test is the fast guard for the exact regression this PR fixes.

_This is an automated review and may be incomplete; please treat findings as suggestions._

PR #915, opened 2026-08-15
test(mgmt-actions): add integration coverage for the command group

digimaun:
Adds an integration scenario for `az iot ops mgmt-actions`, covering enable, show, execute and disable, plus idempotency on both mutating commands.

`execute` is the one that matters. It round trips a real management action through Event Grid MQTT, the dataflow graph and the connector, so it proves the deployed resources work together rather than only proving the CLI created the right ARM resources. Unit tests mock at the ARM layer and cannot give that signal.

The lifecycle runs once per dataflow endpoint auth mode. System assigned and user assigned differ in more than a config field, because a user assigned identity also changes which principal receives the Event Grid role assignments, so both are proven end to end. The user assigned path is federated through `az iot ops identity assign` first, since `enable --mi-user-assigned` only writes the client and tenant ids into the endpoint auth block.

Wired as its own CI scenario matching the upgrade scenario, with `parallel: false`. The scenario name carries no hyphen because it flows into a storage account name and into a schema registry namespace that is truncated to 20 characters.

### Test evidence

[Integration tests run 31837862800](https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/31837862800), full matrix, all seven scenarios green. Management actions reported `2 passed, 2 skipped in 40:58`. The full matrix was run rather than the single scenario because `conftest.py`, `tox.ini` and `settings.py` are shared by every scenario.

The two skips are honest. Both schema tests depend on the connector publishing a request schema to asset status, which does not happen because connector driven schema publishing is off by default. They skip rather than fail so the absence of an optional prerequisite does not mask the result, and `execute` succeeds without it.

The branch is current with `dev`. Note the run above deployed AIO 2607, since the branch predated #913 at the time. A re-run on 2608 is pending #914, which fixes `create` with a custom instance name.

### General Guidelines

Intent for Production

- [x] It is expected that pull requests made to default or core branches such as `dev` or `main` are of production grade.

Basic expectations

- [x] If introducing new functionality or modified behavior, are they backed by unit and/or integration tests? This change is the tests.
- [ ] In the same context as above are command names and their parameter definitions accurate? Not applicable, no command surface changes.
- [x] Have all the relevant unit and integration tests pass? See the run linked above. Unit tests and linter also pass in CI.
- [x] Have linter checks passed? `flake8 azext_edge/` is clean.
- [x] Have extraneous print or debug statements been removed?
- [ ] Have you made an entry in HISTORY.rst? Not applicable. The file is unused in this repo and carries only the initial placeholder, and nothing in packaging references it.


zzcodework:

Reviewing `full diff` (head `da99fd43`, base `dev`).

Overall this is a well-constructed integration suite. Finalizers are registered before the ARM writes they clean up, the module-scoped fixture refuses to run against an instance that already carries `mgmt-actions` state instead of clobbering it, teardown ordering (post-yield `disable` before the resource finalizers) is correct for pytest's LIFO finalizer semantics, and the payload-validation test asserts the specific message rather than just a non-zero exit. Symbols referenced from product code check out on `dev`: `EG_TOPICSPACES_{PUBLISHER,SUBSCRIBER}_ROLE_ID`, `DEFAULT_{DEVICEREGISTRY,EVENTGRID}_MGMT_API_VERSION`, `helpers.run(timeout=...)`, `assert_role_assignment`, the `serial` marker, and the `"The following payload values are invalid"` prefix from `azext_edge/edge/util/schema_validation.py`. Fixture scopes line up too (`settings` session, `tracked_files` module), so no `ScopeMismatch`.

**Blocking**

None.

**Suggestions**

1. `_execute_with_retry` assumes `run()` returns a dict — `azext_edge/tests/edge/orchestration/test_mgmt_actions_int.py`, the `result.get("status", "")` line in the retry loop. `helpers.run` returns a `str` when stdout is not JSON-parseable (and `None` when there is no stdout). In those cases `result.get(...)` raises `AttributeError`, which is not one of the two exceptions the loop catches, so the retry budget is bypassed and the failure surfaces as an opaque `AttributeError` instead of the intended "did not reach Succeeded ... Last outcome" assertion. A cheap `if not isinstance(result, dict): last_outcome = result` branch keeps the loop honest.

2. The retry budget bounds attempts but not wall clock — same helper, `EXECUTE_MAX_ATTEMPTS = 12`, `EXECUTE_COMMAND_TIMEOUT = 420`, backoff capped at 90s. Worst case is ~12 x 420s of command time plus ~860s of sleeps, i.e. roughly 98 minutes per parametrization and ~3.3 hours across `systemAssigned` + `userAssigned`. The scenario step in `.github/workflows/int_test.yml` carries no `timeout-minutes` (only the redeploy delete step does), so a persistently stuck `execute` consumes the GitHub job limit rather than failing fast. An overall deadline (`deadline = time() + N`, checked alongside the attempt counter) would bound the bad case without reducing the propagation tolerance you actually need.

3. `_wait_for_action_schema` costs `SCHEMA_READY_TIMEOUT` (180s) of dead wait on essentially every run. The PR body states connector-driven schema publishing is off by default and the linked run confirms both schema tests skipped, so this poll is expected to time out every time — 3 minutes of module setup burned to produce a skip. Consider gating the wait behind an opt-in env var (same pattern as the new `azext_edge_mgmt_actions_skip_opc_plc`) so the default path skips immediately and the wait only happens where publishing is actually enabled.

4. Related: `test_mgmt_actions_execute_show_schema` and `test_mgmt_actions_execute_payload_validation` have never actually executed — both skipped in run 31837862800. That is honest reporting, but it means those two code paths (including the `--show-schema` shape assertions on `result["type"]` / `result["properties"]`) are unverified. Worth calling out explicitly in the PR description, or better, a one-off manual run with schema publishing enabled to confirm they pass before merge.

5. New environment variables are undocumented — `azext_edge/tests/settings.py` adds `eg_resource_id` and `mgmt_actions_skip_opc_plc`, but `docs/integration-tests.md` only gains the marker line. Both are operator-facing knobs (one avoids creating a billable Event Grid namespace per run, the other skips applying a remote manifest that grants its service account namespace secret access), so they deserve a mention alongside the other `azext_edge_*` settings.

**Nits**

6. `assert not execute_result.get("error")` immediately after `_execute_with_retry` is largely redundant — the helper only returns on `status == "Succeeded"`. Harmless, but it reads as if a `Succeeded` result could still carry an error, which the retry contract already rules out.

7. Pinning `OPC_PLC_MANIFEST_REF` while the workload still resolves `opc-plc:latest` is documented in the comment and is the right trade-off, but it does mean the "cannot change under us" guarantee only covers the Kubernetes resources. Fine as-is given the comment says so.

Summary: no blocking issues found. The main actionable items are the non-dict `run()` result in the retry loop (item 1), the unbounded worst-case wall clock (item 2), and the always-timing-out schema wait (item 3).

*This is an automated review and may be incomplete; it does not replace human review.*

PR #917, opened 2026-08-15
test(mgmt-actions): add integration coverage for the command group

digimaun:
Adds an integration scenario for `az iot ops mgmt-actions` covering enable, show, execute and disable, plus idempotency on both mutating commands.

`execute` is the one that matters. It round trips a real management action through Event Grid MQTT, the dataflow graph and the connector, so it proves the deployed resources work together rather than proving the CLI wrote the right ARM resources. Unit tests mock at the ARM layer and cannot give that signal.

The lifecycle runs once per dataflow endpoint auth mode. A user assigned identity changes which principal receives the Event Grid role assignments rather than just a config value, so both modes are proven end to end. The user assigned path is federated through `az iot ops identity assign` first, because `enable --mi-user-assigned` only writes the client and tenant ids into the endpoint auth block.

Wired as its own CI scenario with `parallel: false`, matching the upgrade scenario. The scenario name carries no hyphen because it flows into a storage account name and into a schema registry namespace truncated to 20 characters.

### Test evidence

[Run 31837862800](https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/31837862800) covers AIO 2607. Full matrix, all seven scenarios green, management actions reporting `2 passed, 2 skipped in 40:58`. The full matrix ran rather than the single scenario because `conftest.py`, `tox.ini` and `settings.py` are shared by every scenario.

[Run 31906218398](https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/31906218398) covers AIO 2608 at the current head. The management actions tests passed there as well.

Both schema tests skip because connector driven schema publishing is off by default, so the request schema never appears on asset status. They skip rather than fail so a missing optional prerequisite does not mask the result, and `execute` succeeds without it. Those two tests have never executed, so their assertions are unverified.

### Teardown failure, not from this change

Runs currently fail in workflow teardown at `Delete AIO resources`, after the tests have finished. `az iot ops delete` polls for two hours and then returns `Operation returned an invalid status 'OK'`, which is what surfaces when the resource provider answers with an HTTP status the poller does not expect.

This is not specific to this branch. The dev nightly [31876055322](https://github.com/Azure/azure-iot-ops-cli-extension/actions/runs/31876055322) fails the same way in Insecure listener, Workload identity and Cloud tests, and dev has not had a green integration run since 2026-08-13. It should be tracked separately.

### General Guidelines

Intent for Production

- [x] It is expected that pull requests made to default or core branches such as `dev` or `main` are of production grade.

Basic expectations

- [x] If introducing new functionality or modified behavior, are they backed by unit and/or integration tests? This change is the tests.
- [ ] In the same context as above are command names and their parameter definitions accurate? Not applicable, no command surface changes.
- [x] Have all the relevant unit and integration tests pass? See the runs linked above. Unit tests and linter also pass in CI.
- [x] Have linter checks passed? `flake8 azext_edge/` is clean.
- [x] Have extraneous print or debug statements been removed?
- [ ] Have you made an entry in HISTORY.rst? Not applicable. The file is unused in this repo and carries only the initial placeholder, and nothing in packaging references it.


zzcodework:

Reviewing `full diff` (head `bcf7499b`, base `dev`, 7 files, +747/-2). No prior human comments on this PR at review time.

Verified against `dev`: `helpers.run()` does accept `timeout`, `assert_role_assignment` / `create_file` / `remove_file` signatures match, `serial` is a registered marker, `settings` is session-scoped and `tracked_files` module-scoped (so the module-scoped `mgmt_actions_setup` is legal), `EG_TOPICSPACES_*_ROLE_ID` and both API-version constants exist, `--payload/-p` and `--show-schema` exist on `mgmt-actions execute`, and the pinned OPC PLC manifest really does name the Deployment `opc-plc-000000` (the Service is `opcplc-000000`, so the two constants are correct despite looking inconsistent). The tox/marker/scenario wiring is consistent end to end (`mgmtactions` marker with `edge_exclude=True` -> `SCENARIO="mgmtactions"` -> `python-mgmtactions-int` -> `.github/test-scenarios.yml` with `parallel: false`).

**Blocking**

None.

**Suggestions**

1. `azext_edge/tests/edge/orchestration/test_mgmt_actions_int.py` — cross-parametrization contamination. `mgmt_actions_setup` is module-scoped and `test_mgmt_actions_lifecycle` is parametrized over two auth modes. If the `systemAssigned` run fails anywhere after `enable` (most likely `_execute_with_retry` exhausting its budget), mgmt-actions is left enabled, and the `userAssigned` run then fails immediately on `assert before["enabled"] is False, "expected a clean baseline..."`. The second failure reports a baseline problem rather than the real one, and the user-assigned path — the mode that exercises the different Event Grid principal, i.e. the reason for the parametrization — is never actually tested. Wrapping the enable-through-execute body in `try/finally` with a best-effort `mgmt-actions disable -y` would keep each parametrization independent and keep the failure attributable.

2. `_execute_with_retry` — `EXECUTE_COMMAND_TIMEOUT` does not reliably bound an attempt. `run()` uses `subprocess.run(..., shell=True, timeout=...)`; on timeout Python kills the shell, but the `az` grandchild survives and inherits the stdout/stderr pipes, so the follow-up `communicate()` can block until `az` itself exits. The docstring justifies the per-attempt timeout as protection against an action that never reaches a terminal state — that is exactly the case where the timeout may not fire promptly. If bounding the hang matters, the attempt needs process-group termination (e.g. `start_new_session=True` + `killpg`) rather than `subprocess.run`'s default. Otherwise it is worth softening the comment so a future reader does not rely on a guarantee that is not there.

3. `_execute_with_retry` — retrying deterministic failures burns the full budget. Any `CLIInternalError` is retried, including ones that can never succeed (bad asset/group/action name, or a client-side schema-validation rejection). The broad retry is defended for authorization propagation, which is fair, but a payload rejection carrying `The following payload values are invalid` (from `schema_validation.py`) is never transient and would consume the ~30 minute deadline before failing with a misleading "did not reach Succeeded" message. A short fail-fast list for known-permanent messages would keep the useful breadth without the cost.

4. `_wait_for_action_schema` / the two schema tests — the run pays `SCHEMA_READY_TIMEOUT` (180s) plus status dumps on every module setup for an outcome the PR body says is known in advance (connector-driven schema publishing is off by default, so both dependent tests skip). Since a `skip_opc_plc` style flag already exists in `settings.py`, consider gating the wait behind an equivalent opt-in (e.g. `azext_edge_mgmt_actions_expect_schema`) so CI skips the wait by default and, when the flag is on, a missing schema *fails* instead of skipping. As written, the two schema tests have never executed their assertions and nothing in CI will ever tell you when they rot.

**Nits**

5. `docs/integration-tests.md` — the marker line is added but the scenario's non-obvious prerequisites are not: an existing instance with an `adrNamespaceRef`, a kubectl-reachable cluster, permission to create an Event Grid namespace and Event Grid role assignments, and the two new settings (`azext_edge_eg_resource_id`, `azext_edge_mgmt_actions_skip_opc_plc`). Without those, a local run fails inside fixture setup rather than with a readable message.

6. `_execute_with_retry` — the deadline is checked before the backoff sleep, so the effective ceiling is `EXECUTE_TOTAL_TIMEOUT` + one sleep + one command timeout, slightly more than the comment's arithmetic. Harmless, but the comment is precise enough elsewhere that the gap stands out.

**Summary**

Solid, unusually well-reasoned test addition — the comments explain *why* rather than *what*, the fixture refuses to run against a partially configured instance instead of silently deleting resources it did not create, finalizers are registered before the ARM writes they clean up, LIFO ordering is deliberate, the topic-space listing follows pagination, and the payload-validation test asserts the specific message rather than relying on `expect_failure`. My main substantive concern is the module-scoped fixture shared across the two parametrizations (1), which can quietly cost you all coverage of the user-assigned path on any first-mode failure.

_This is an automated review and may be incomplete; it did not execute the tests._

zzcodework:

Reviewing `changes since bcf7499b` (head `5dd88123`) — one commit, `test(mgmt-actions): keep each auth mode independent on failure`, touching only `azext_edge/tests/edge/orchestration/test_mgmt_actions_int.py` (+13/-1).

This directly addresses the cross-parametrization contamination I raised on the previous iteration: `mgmt_actions_setup` is module-scoped and `test_mgmt_actions_lifecycle` is parametrized over both auth modes, so a failure after `enable` used to leave mgmt-actions enabled and the second mode then died on the `assert before["enabled"] is False` baseline check instead of running. Registering `_ensure_disabled` via `request.addfinalizer` immediately after the baseline assert and before `enable` is the right placement — it covers every failure path after enable, and `request` was already a parameter of the test so no signature change was needed. The revised `EXECUTE_TOTAL_TIMEOUT` comment ("budget plus one backoff interval and one command timeout") is also now accurate: the deadline is checked before `sleep(interval)`, so the worst case really is budget + interval + `EXECUTE_COMMAND_TIMEOUT`, and dropping the stale "around 37 minutes" figure removes a number that no longer matched the constants.

**Blocking**

None.

**Suggestions**

1. `azext_edge/tests/edge/orchestration/test_mgmt_actions_int.py` — the finalizer swallows the exception without recording it. `except Exception: logger.error(f"Failed to disable mgmt-actions after the {auth_mode} lifecycle test.")` tells you cleanup failed but not why. When this fires, the next auth mode still fails on the dirty-baseline assert, and the actual cause (throttling, ARM error, expired token) is exactly what you need from the CI log. Capture it: `except Exception as e: logger.error("Failed to disable mgmt-actions after the %s lifecycle test: %s", auth_mode, e, exc_info=True)`.

**Nits**

1. Same file — in the happy path the test already ends with `run(... disable ... -y)` twice (teardown assertions plus the idempotency check), so the finalizer adds a third `disable` call per auth mode. It is harmless because disable-when-already-disabled is explicitly exercised as idempotent, but it is one more real ARM round trip per parametrization on a test that is already long. Guarding with a small `enabled` flag set after `enable` and cleared after the final disable would avoid it — only worth doing if it shows up in run time.

**Summary**

Small, well-scoped follow-up that fixes the isolation problem correctly and tightens a comment that had drifted from the constants. No correctness, security, or command-surface concerns; test-only change so no `HISTORY.rst` entry is expected. The one thing worth changing before merge is logging the cleanup exception detail.

cheatsheet1999 (approved):

PR #918, opened 2026-08-17
fix(lro): surface terminal operation error details

HangyiWang:
Fix useless message: `ERROR: Operation returned an invalid status 'OK'`
- format failed or canceled terminal LRO payloads returned with HTTP 2xx
- preserve the original `HttpResponseError` type and standard ARM 4xx/5xx errors
- include service error code, message, details, and request ID when available

### DIFF
--- a/azext_edge/edge/util/az_client.py
+++ b/azext_edge/edge/util/az_client.py
@@ -27,6 +27,8 @@ AZURE_CLI_CREDENTIAL = AzureCliCredential()
 
 POLL_RETRIES = 240
 POLL_WAIT_SEC = 15
+_LRO_FAILURE_STATUSES = frozenset({"failed", "canceled", "cancelled"})
+_INVALID_STATUS_MESSAGE = "Operation returned an invalid status"
 
 logger = get_logger(__name__)
 
@@ -297,6 +299,8 @@ def get_keyvault_client(subscription_id: str, keyvault_scope: Optional[str] = No
 
 
 def wait_for_terminal_state(poller: "LROPoller", wait_sec: int = POLL_WAIT_SEC, **_) -> JSON:
+    from azure.core.exceptions import HttpResponseError
+
     # resource client does not handle sigint well
     counter = 0
     while counter < POLL_RETRIES:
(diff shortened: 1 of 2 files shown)


zzcodework:

Reviewing `full diff` (head `7cef010e49`)

Context read: `azext_edge/edge/util/az_client.py` on `dev`, plus `azure.core.exceptions.HttpResponseError` construction semantics. No prior review comments on this PR.

**Blocking**

- `test_az_client_unit.py` — the flagship test (`test_wait_for_terminal_state_formats_failed_lro`) exercises a state that cannot occur in production, so it does not prove the fix works. The mock sets `response.json()` to the full payload but `response.text()` to `""`. In a real `HttpResponseError`, azure-core builds `self.error` by calling `ODataV4Format(json.loads(response.text()))` — and for a real ARM async-operation body those two are the *same bytes*. So with a real response carrying `{"status": "Failed", "error": {"code": "ResourceOperationFailure", "message": "..."}}`, azure-core parses the error successfully, `exception.error` is populated, and `_get_lro_failure_message()` bails at the `or exception.error` guard — the very branch this test claims to cover would never run. (Additionally, azure-core would already have set `message` to the ODataV4 string in that case, so the "invalid status" prefix guard would also fail.)

  Concretely, this means the formatter only fires when azure-core *couldn't* parse the body — e.g. `error` missing entirely, `error` a bare string, or `error` present but without a `code` key (`ODataV4Format` raises `KeyError` there and `_parse_odata_body` swallows it). Those are real and worth handling, and the other three tests cover them. But please make the tests honest: drive `response.text()` from the same payload as `response.json()` (e.g. `response.text.return_value = json.dumps(payload)`), and let `HttpResponseError.__init__` do its normal parsing. That will show which scenarios actually reach the new code and will keep the suite from silently green-lighting a regression if the guards are ever tightened.

**Suggestions**

- `az_client.py` — `exception.message.startswith(_INVALID_STATUS_MESSAGE)` couples the fix to an azure-core internal message string (`"Operation returned an invalid status '{reason}'"`). If that wording changes on an SDK bump, the improvement silently disappears with no test failure to warn you (the current tests construct the message via the same SDK path, so they *would* catch it — one more reason to fix the mocks above). Worth a short comment naming the azure-core version this string comes from.
- `az_client.py` — replacing both `e.message` and `e.args` discards the original SDK text. Since the point is diagnosability, consider appending the original as a trailing line rather than overwriting, so support still sees the raw SDK framing in issue reports.
- `wait_for_terminal_states()` (plural, right below) is untouched. It returns pollers without calling `.result()`, so callers that later invoke `poller.result()` themselves — e.g. the parallel batches in `deletion2.py` — still get the unhelpful `Operation returned an invalid status 'OK'`. Routing those through the same formatter (or a small shared `resolve_poller_result()` helper) would make the improvement consistent across the single and batch paths, which matters since instance/CL deletion is where users hit this most.

**Nits**

- `_format_lro_error_detail()` falls back to `str(dict(detail))` when a detail mapping has neither `code` nor `message`, which renders Python dict syntax into a user-facing error. Minor, but a `json.dumps` or a "no details" placeholder would read better in a terminal.
- `_LRO_FAILURE_STATUSES` covering both `canceled` and `cancelled` is a nice touch; consider also handling `aborted`, which some RPs emit for terminal-failure LROs.

**Summary**

The intent is right and this fixes a genuinely bad UX (`ERROR: Operation returned an invalid status 'OK'` is one of the least actionable errors this extension produces), and the fail-closed guards mean the change cannot make existing errors worse. My one real objection is the test fidelity: the mocked `json()`/`text()` divergence makes the headline scenario pass through a code path it would not take against a real ARM response, so the suite currently overstates what is covered. Fixing the mocks will also tell you whether the `exception.error` guard is drawn where you intend it.

_This is an automated review and may be incomplete; please treat findings as suggestions for the human reviewers._

cheatsheet1999 (approved):

zzcodework:

Reviewing `changes since 7cef010e` (head `9bbdb616`), base `dev`.

The delta since the last reviewed iteration consists solely of two merge commits bringing `dev` into `hangyi/fix-lro-terminal-error-details`:

- `8aa3a0d1` feat: validate dataflow graph transform runtime compatibility (#919)
- `fbdc6acf` build: update IoT Operations to 1.4.73 (#922)

The resulting file changes (`providers/orchestration/common.py`, `resources/dataflow_graphs.py`, `template.py`, and the corresponding unit tests) all come from those upstream PRs and are unrelated to this PR's LRO error-surfacing logic. No new author changes, and the merge did not touch any of the files this PR modifies.

**Blocking:** none.
**Suggestions:** none.
**Nits:** none.

Summary: no reviewable change in this iteration — upstream `dev` sync only. My earlier review of the substantive diff still stands.

PR #919, opened 2026-08-17
feat: validate dataflow graph transform runtime compatibility

ketkimnaik:
- Adds a best-effort, client-side check on  dataflow graph apply : when a graph contains transform nodes, the CLI compares the transform manifest's  moduleRequirements.runtimeVersion  against the target instance's installed AIO runtime and blocks the apply with a clear error if the instance runtime is older than required (mirrors the portal's  isRuntimeVersionCompatible  UX).

- Fails open in every ambiguous case — missing/malformed required version, undeterminable instance version, or OCI fetch failure — so enforcement is safely deferred to the runtime; the instance version is fetched lazily, only when transform nodes are present, to avoid an extra API call.

---

### General Guidelines

Intent for Production

Basic expectations


HangyiWang (approved):

zzcodework:

Reviewing `full diff` (head `3c74459a0d`)

Context read: `azext_edge/edge/providers/orchestration/resources/dataflow_graphs.py` on `dev`, `azext_edge/edge/util/machinery.py`, and existing semver usage in `upgrade2.py` / `clone.py`. Noting the Copilot review and @HangyiWang's approval — the points below are ones not already covered.

**Blocking**

- None. The semver usage is correct (`scoped_semver_import()` returns the `Version` class, so `Version.parse(..., optional_minor_and_patch=True)` and the `<` comparison are the right APIs, matching `deletion2.py` and `mgmt_actions.py`), and the fail-open branches all have tests.

**Suggestions**

- `dataflow_graphs.py` — semantic-equivalence question worth confirming before merge: the check compares the manifest's `moduleRequirements.runtimeVersion` against the instance's `properties.version`. That ARM property is the *instance/AIO release* version (same field `clone.py` reads as `self.version`). If the manifest's `runtimeVersion` is expressed against a different scheme — e.g. the dataflow/connector runtime image version rather than the AIO instance version — the comparison silently blocks valid applies. The docstring cites the portal's `isRuntimeVersionCompatible` in `runtimeCompatibility.ts`; if the UX compares against a different source value, this needs to match it. A one-line comment recording which value the UX uses would keep the two implementations from drifting.

- `dataflow_graphs.py` — prerelease handling is a false-positive risk. `Version` orders prereleases *below* the corresponding release, so an instance reporting e.g. `1.4.0-preview` against a manifest requiring `1.4.0` parses cleanly (no `ValueError`, so the fail-open path does not trigger) and hard-blocks the apply. If AIO instances can ever report a prerelease/build-suffixed version, consider comparing only `(major, minor, patch)` — `parsed_instance.to_tuple()[:3] < parsed_required.to_tuple()[:3]` — or explicitly `.finalize_version()` both sides.

- No escape hatch: this is the only client-side check in `apply` that hard-fails on *server-side-authoritative* data. If a transform manifest ships a wrong or overly conservative `runtimeVersion`, users are blocked at the CLI even though the service would accept the graph, with no flag to override. Everything else in `_validate_graph_node_artifact_config` degrades to a warning when the artifact can't be trusted. Either confirm with the service team that the manifest value is authoritative, or make the error message point at a workaround.

**Nits**

- `test_dataflow_graph_apply_transform_runtime_incompatible` — `assert "1.0" in msg  # selected version` doesn't independently verify anything: it can only be satisfied by the `myartifact:1.0` substring already asserted on the line above (`1.99.0` and `1.3.137` don't contain `1.0`). Assert on the full `myartifact:1.0` string instead, so the test would actually catch the artifact reference being dropped from the message.

- `_get_instance_version()` is invoked whenever any Graph node exists, even if no manifest declares `moduleRequirements`. That's one extra `instances.show` per apply. Given the artifact fetch already dominates, this is fine — just noting the "lazily to avoid an extra API call" comment is only true relative to non-graph applies.

**Summary**

Clean, well-scoped change with good fail-open discipline and solid test coverage of the skip paths. The two things I'd want answered before merge are whether `properties.version` is really the same version axis the manifest's `runtimeVersion` is authored against, and how prerelease instance versions compare — both would surface as spurious hard blocks for users rather than as test failures.

_This is an automated review and may be incomplete; please treat findings as suggestions for the human reviewers._

azext_edge/edge/providers/orchestration/resources/dataflow_graphs.py
```diff
@@ -461,6 +495,65 @@ def _validate_graph_node_artifact_config(
...
+                "Could not compare runtime versions (required '%s', instance '%s') for graph node "
+                "'%s' — skipping runtime compatibility check. %s",
+                required_runtime,
+                instance_version,
+                node.get("name"),
+                ex,
+            )
+            return
+
+        if parsed_instance < parsed_required:
```
cheatsheet1999, line 548: This check treats SemVer ordering as compatibility, but as I understand it, those are not equivalent. For example, 2.0.0 > 1.1.0, so the current logic would allow an instance running 2.0.0 to use a transform that requires 1.1.0. However, the host rejects this because a major-version change is considered breaking. The same applies to pre-1.0 versions: 0.2.x > 0.1.x, but 0.2.x is not guaranteed to be compatible with a 0.1.x requirement.

This means the CLI could allow a graph update that later fails at runtime. I think we should mirror the compatibility rules enforced by the runtime, for example, requiring the same major version for stable releases and the same minor version for pre-1.0 releases, in addition to ensuring that the installed version is not older than the required version.

It would also be worth adding tests for 2.0.0 versus 1.1.0 and 0.2.0 versus 0.1.0, both of which should be rejected.

ketkimnaik: Confirmed with the DOE team that compatibility is defined as instanceVersion >= runtimeVersion, so 2.0.0 is compatible with 1.1.0 and no same-major or pre-1.0 same-minor restriction is required.

cheatsheet1999: Thanks for confirming.

PR #926, opened 2026-08-31
fix(upgrade): correct Desired State when the CLI is behind the cluster

cheatsheet1999:
## Problem

When the installed `azure-iot-ops` extension is older than the AIO build deployed on the cluster, the upgrade table rendered a Desired State lower than Current State:

│ iotOperations │ 1.3.137 [stable] │ 1.3.105 [stable] │ • ... │

The customer read this as an imminent downgrade and stopped.

No downgrade was attempted. Absent a version override, `_has_delta_in_version()` requires `desired > current`, so it is false on a stale CLI and the patch carries `configurationSettings` only. The table was describing the CLI's built-in constant instead of the values the patch actually sends.

## Changes

**1. Desired State is a per-field overlay of the patch.** A version-only patch no longer blanks the train, and a train-only patch no longer misreports the version. A same-version retry now reads `Reconcile at version X` instead of `Update version to X`. When the patch sends neither `version` nor `releaseTrain`, the Desired cell is dim rather than cyan, so the change signal is reserved for a version that is actually moving.

**2. A stale CLI is called out**, before the `has_upgrades()` early return so it is still visible on a no-op run. The warning states the outcome — no deployed extension version will be changed — and points at `az extension update`. It does not name `--force` or the version overrides: those are hidden, and each CLI release pins and tests AIO alongside its dependencies as a set.

**3. A failed extension is repaired at its installed version.** `validate_upgrade()` and `get_patch()` shared one condition (`_has_delta_in_version() or _has_non_success_state()`) and unconditionally resolved the built-in target, so on a stale CLI the repair aborted with a downgrade error naming a version the patch would never have sent. The branches are now separated at both sites, and an extension with no resolvable version raises instead of emitting `"version": null`.

The reconcile branch is only reachable when `_has_delta_in_version()` is false — no override *and* `desired <= current` — so it can only ever send a version **≥** what the old code sent. The change cannot skip or downgrade a legitimate upgrade. The explicit downgrade guard and `--force` are unchanged.

This clears the **stale CLI cause** of the `Unknown / Unknown / Check configuration` row for `certManager` / `secretStore`, which came from the same raise. It does not clear that row in general: `validate_upgrade()` runs only for the ops extension, so the other extensions first reach this logic when `format_extension_row()` calls `get_patch()` inside a broad handler, and a failed extension reporting no installed version still renders `Unknown / Unknown`. That path is pre-existing and unchanged here, as is the apply loop's exception handling.

**4. The train guard validates the version the patch sends.** Found by the automated review on this PR. In the reconcile branch an `--ops-train` override re-runs `_validate_version_upgrade()`, which compared the stale built-in version rather than the reconcile version being sent — so `--ops-train stable` on a failed extension with a cluster ahead of the CLI raised the same spurious downgrade error this PR exists to remove. `_validate_version_upgrade()` now takes an optional `target_version`; the reconcile call site passes the resolved version and every other call site is unchanged by the default. The release-train checks are retained, so `--ops-train preview` on a stable cluster is still blocked — now reporting the real reason instead of a downgrade that was never proposed.

## Implementation notes

- **`is_cli_behind_cluster()` respects `--ops-version`.** Without it, `--ops-version 1.5.0` printed *"this CLI has no newer version to offer"* directly above `Update version to 1.5.0`. An explicit version always takes the version-delta path, so it can never reach the reconcile case the warning describes.

- **One shared helper rather than a guard duplicated at both sites.** `validate_upgrade()` and `get_patch()` previously shared a single condition that resolved the built-in target at both, which is where the bug came from. Both now call `_validate_and_resolve_version()`, and `test_validate_upgrade_and_get_patch_have_same_validation_outcome` pins that they cannot disagree.

- **The train guard in the reconcile branch is conditional.** A failed extension can report a version with no release train; validating unconditionally would reject it on a missing train even though the patch only reapplies the installed version and sends no train.

## Validation

Live A/B on a cluster (AIO `1.4.73`) with two extension trees differing in exactly one file:

| scenario | before | after |
|---|---|---|
| stale CLI, config-only upgrade | `1.4.73 -> 1.3.105` (defect) | `1.4.73 -> 1.4.73` + advisory |
| **applied** run, then ARM read-back | - | `version=1.4.73`, `Succeeded`, config written - **no downgrade** |
| genuinely `Failed` extension | hard-fails; repair impossible | `Reconcile at version x.x.x` |
| explicit downgrade / `--force` / repeat run | - | unchanged |

### DIFF
--- a/azext_edge/edge/providers/orchestration/upgrade2.py
+++ b/azext_edge/edge/providers/orchestration/upgrade2.py
@@ -1215,7 +1215,7 @@ class ExtensionUpgradeState:
                     "Cannot validate upgrade path."
                 )
             if self._has_delta_in_train():
-                self._validate_version_upgrade()
+                self._validate_version_upgrade(target_version=reconcile_version)
             return reconcile_version
 
         return None
@@ -1365,7 +1365,7 @@ class ExtensionUpgradeState:
         """
         return self.provisioning_state.lower() != PROVISIONING_STATE_SUCCESS.lower()
 
-    def _validate_version_upgrade(self):
+    def _validate_version_upgrade(self, target_version: Optional[str] = None):
         # Skip validation for CREATE/DELETE operations
         if self.operation_type in [ExtensionOperation.CREATE, ExtensionOperation.DELETE]:
             return
@@ -1373,12 +1373,14 @@ class ExtensionUpgradeState:
         if self.force:
             return
 
+        target_version = target_version or self.desired_version[0]
+
         # Validate required fields are present
         if not self.current_version[0]:
             raise ValidationError(
                 f"Unable to determine installed version for {self.moniker} extension. Cannot validate upgrade path."
             )
-        if not self.desired_version[0]:
+        if not target_version:
             raise ValidationError(
                 f"Unable to determine target version for {self.moniker} extension. Cannot validate upgrade path."
             )
@@ -1394,7 +1396,7 @@ class ExtensionUpgradeState:
             )
 
         parsed_current = self.semver.parse(self.current_version[0])
-        parsed_desired = self.semver.parse(self.desired_version[0])
+        parsed_desired = self.semver.parse(target_version)
 
         current_is_preview = self.current_version[1].lower() != "stable"
         desired_is_preview = self.desired_version[1].lower() != "stable"
@@ -1403,7 +1405,7 @@ class ExtensionUpgradeState:
         if parsed_desired < parsed_current:
             raise ValidationError(
                 f"Installed {self.moniker} extension version is {self.current_version[0]}.\n"
-                f"The desired {self.desired_version[0]} version is a downgrade which is not supported."
+                f"The desired {target_version} version is a downgrade which is not supported."
             )
 
         if self.moniker != EXTENSION_MONIKER_OPS:
@@ -1413,14 +1415,14 @@ class ExtensionUpgradeState:
         if parsed_desired.major != parsed_current.major:
             raise ValidationError(
                 f"Installed {self.moniker} extension version is {self.current_version[0]}.\n"
-                f"The desired {self.desired_version[0]} version is incompatible (different major version)."
+                f"The desired {target_version} version is incompatible (different major version)."
             )
 
         minor_diff = parsed_desired.minor - parsed_current.minor
         if minor_diff > 2:
             raise ValidationError(
                 f"Installed {self.moniker} extension version is {self.current_version[0]}.\n"
-                f"The desired {self.desired_version[0]} version is incompatible (more than 2 minor versions ahead)."
+                f"The desired {target_version} version is incompatible (more than 2 minor versions ahead)."
             )
 
         min_v2_semver_broker_upgrade = self.semver.parse(MIN_INSTANCE_VERSION_V1_FOR_V2_UPGRADE)
@@ -1428,7 +1430,7 @@ class ExtensionUpgradeState:
         if parsed_current < min_v2_semver_broker_upgrade and parsed_desired >= min_v2_semver:
             raise ValidationError(
                 f"Installed {self.moniker} extension version is {self.current_version[0]}.\n"
-                f"The desired {self.desired_version[0]} version is incompatible "
+                f"The desired {target_version} version is incompatible "
                 f"(min compatible upgrade version {min_v2_semver_broker_upgrade}).\n"
                 f"Please first upgrade to at least {min_v2_semver_broker_upgrade}/AIO2506. "
                 "See https://aka.ms/aio-versions for version details."
--- a/azext_edge/edge/providers/orchestration/upgrade2.py
+++ b/azext_edge/edge/providers/orchestration/upgrade2.py
@@ -90,10 +90,9 @@ def upgrade_ops_instance(
     stale_monikers = upgrade_state.get_stale_cli_extensions()
     if stale_monikers:
         logger.warning(
-            "The version of %s built into this azure-iot-ops CLI extension is older than the "
-            "version deployed on the cluster, so this CLI has no newer version to offer. "
-            "Nothing will be downgraded: a lower version is only applied when explicitly "
-            "requested together with --force. "
+            "This azure-iot-ops CLI extension has no newer version to offer for: %s. "
+            "The version deployed on the cluster is newer than the version built into this CLI. "
+            "No deployed extension version will be changed. "
             "Run 'az extension update --name azure-iot-ops' to pick up newer versions.",
             ", ".join(stale_monikers),
         )
@@ -495,12 +494,12 @@ def format_extension_row(ext: "ExtensionUpgradeState") -> Tuple[str, str, str]:
         patch = ext.get_patch()
         props = patch.get("properties", {})
 
-        desired = "[cyan]{}[/cyan]".format(
-            format_version_with_train(
-                props.get("version", ext.current_version[0]),
-                props.get("releaseTrain", ext.current_version[1]),
-            )
+        desired_state = format_version_with_train(
+            props.get("version", ext.current_version[0]),
+            props.get("releaseTrain", ext.current_version[1]),
         )
+        desired_style = "cyan" if ("version" in props or "releaseTrain" in props) else "dim"
+        desired = f"[{desired_style}]{desired_state}[/{desired_style}]"
 
         if not patch or "properties" not in patch:
             return current, desired, "[dim]No changes[/dim]"
@@ -1214,6 +1213,8 @@ class ExtensionUpgradeState:
                     f"Unable to determine installed version for {self.moniker} extension. "
                     "Cannot validate upgrade path."
                 )
+            # A failed extension can report a version with no release train; guarding on the delta
+            # avoids rejecting it for a missing train the patch never sends.
             if self._has_delta_in_train():
                 self._validate_version_upgrade(target_version=reconcile_version)
             return reconcile_version
--- a/azext_edge/edge/providers/orchestration/upgrade2.py
+++ b/azext_edge/edge/providers/orchestration/upgrade2.py
@@ -87,6 +87,17 @@ def upgrade_ops_instance(
 
     upgrade_state = upgrade_manager.analyze_cluster(**kwargs)
 
+    stale_monikers = upgrade_state.get_stale_cli_extensions()
+    if stale_monikers:
+        logger.warning(
+            "The version of %s built into this azure-iot-ops CLI extension is older than the "
+            "version deployed on the cluster, so this CLI has no newer version to offer. "
+            "Nothing will be downgraded: a lower version is only applied when explicitly "
+            "requested together with --force. "
+            "Run 'az extension update --name azure-iot-ops' to pick up newer versions.",
+            ", ".join(stale_monikers),
+        )
+
     if not upgrade_state.has_upgrades():
         logger.warning("Nothing to upgrade :)")
         return
@@ -480,17 +491,27 @@ def format_extension_row(ext: "ExtensionUpgradeState") -> Tuple[str, str, str]:
 
         # UPDATE operation
         current = format_version_with_train(ext.current_version[0], ext.current_version[1]) + status_indicator
-        desired = f"[cyan]{format_version_with_train(ext.desired_version[0], ext.desired_version[1])}[/cyan]"
 
         patch = ext.get_patch()
+        props = patch.get("properties", {})
+
+        desired = "[cyan]{}[/cyan]".format(
+            format_version_with_train(
+                props.get("version", ext.current_version[0]),
+                props.get("releaseTrain", ext.current_version[1]),
+            )
+        )
+
         if not patch or "properties" not in patch:
             return current, desired, "[dim]No changes[/dim]"
 
-        props = patch.get("properties", {})
         action_lines = []
 
         if "version" in props:
-            action_lines.append(f"[cyan]•[/cyan] Update version to [bold]{props['version']}[/bold]")
+            if props["version"] == ext.current_version[0]:
+                action_lines.append(f"[cyan]•[/cyan] Reconcile at version [bold]{props['version']}[/bold]")
+            else:
+                action_lines.append(f"[cyan]•[/cyan] Update version to [bold]{props['version']}[/bold]")
 
         if "releaseTrain" in props:
             action_lines.append(f"[cyan]•[/cyan] Change release train to [bold]{props['releaseTrain']}[/bold]")
@@ -780,6 +801,9 @@ class ClusterUpgradeState:
             or bool(self.connector_template_needed)
             or bool(self.secretsync_migration_needed)
         )
+
+    def get_stale_cli_extensions(self) -> List[str]:
+        return [ext_state.moniker for ext_state in self.extension_upgrades if ext_state.is_cli_behind_cluster()]
 
     def _check_instance_upgrade(self) -> bool:
         """Check if instance needs updates.
@@ -1089,6 +1113,23 @@ class ExtensionUpgradeState:
             return "N/A"
         return self.extension.get("properties", {}).get("provisioningState", "Unknown")
 
+    def is_cli_behind_cluster(self) -> bool:
+        if self.operation_type != ExtensionOperation.UPDATE:
+            return False
+
+        # An explicit --ops-version, not the built-in default, decides what gets sent.
+        if self.override.version:
+            return False
+
+        built_in_version = self.desired_version_map.get("version")
+        if not built_in_version or not self.current_version[0]:
+            return False
+
+        try:
+            return self.semver.parse(built_in_version) < self.semver.parse(self.current_version[0])
+        except ValueError:
+            return False
+
     def can_upgrade(self) -> bool:
         if self.operation_type in [ExtensionOperation.CREATE, ExtensionOperation.DELETE]:
             return True
@@ -1118,9 +1159,10 @@ class ExtensionUpgradeState:
             "properties": {},
         }
 
-        if self._has_delta_in_version() or self._has_non_success_state():
-            self._validate_version_upgrade()
-            payload["properties"]["version"] = self.desired_version[0]
+        target_version = self._validate_and_resolve_version()
+        if target_version:
+            payload["properties"]["version"] = target_version
+
         if self._has_delta_in_train():
             payload["properties"]["releaseTrain"] = self.desired_version[1]
         if self._has_delta_in_config():
@@ -1155,10 +1197,33 @@ class ExtensionUpgradeState:
         Raises:
             ValidationError: If the upgrade is not valid (e.g., downgrade, incompatible versions).
         """
-        # Always validate if there's an override version (user explicitly requested upgrade)
-        # or if there's a version delta or non-success state
-        if self._has_delta_in_version() or self._has_non_success_state():
+        if self.operation_type != ExtensionOperation.UPDATE:
+            return
+
+        self._validate_and_resolve_version()
+
+    def _validate_and_resolve_version(self) -> Optional[str]:
+        if self._has_delta_in_version():
             self._validate_version_upgrade()
+            return self.desired_version[0]
+
+        if self._has_non_success_state():
+            reconcile_version = self._get_reconcile_version()
+            if not reconcile_version:
+                raise ValidationError(
+                    f"Unable to determine installed version for {self.moniker} extension. "
+                    "Cannot validate upgrade path."
+                )
+            if self._has_delta_in_train():
+                self._validate_version_upgrade()
+            return reconcile_version
+
+        return None
+
+    def _get_reconcile_version(self) -> Optional[str]:
+        if self.current_version[0]:
+            return self.current_version[0]
+        return self.desired_version[0] if self.force else None
 
     def _should_migrate_mqtt_config(self) -> bool:
         if not self.extension:
--- a/azext_edge/edge/providers/orchestration/upgrade2.py
+++ b/azext_edge/edge/providers/orchestration/upgrade2.py
@@ -92,7 +92,7 @@ def upgrade_ops_instance(
         logger.warning(
             "This azure-iot-ops CLI extension has no newer version to offer for: %s. "
             "The version deployed on the cluster is newer than the version built into this CLI. "
-            "No deployed extension version will be changed. "
+            "No deployed version of the listed extensions will be changed. "
             "Run 'az extension update --name azure-iot-ops' to pick up newer versions.",
             ", ".join(stale_monikers),
         )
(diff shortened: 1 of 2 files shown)


azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1155,10 +1197,33 @@ def validate_upgrade(self) -> None:
...
+
+        if self._has_non_success_state():
+            reconcile_version = self._get_reconcile_version()
+            if not reconcile_version:
+                raise ValidationError(
+                    f"Unable to determine installed version for {self.moniker} extension. "
+                    "Cannot validate upgrade path."
+                )
+            if self._has_delta_in_train():
+                self._validate_version_upgrade()
```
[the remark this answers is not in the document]
cheatsheet1999: Valid. `_validate_version_upgrade()` now takes an optional `target_version`. The reconcile call site passes the resolved version.

zzcodework:

Reviewing `full diff` (head `6ae49a07`)

Read the change against `dev` (`azext_edge/edge/providers/orchestration/upgrade2.py`) and traced the three code paths it touches: `format_extension_row` (display), `_validate_and_resolve_version` (patch/validation), and the new `is_cli_behind_cluster` advisory. I also read the existing thread — Copilot's point about the train-override call site validating the stale built-in version instead of the reconcile version is already addressed by the `target_version` parameter on `_validate_version_upgrade`, so I am not repeating it.

**Blocking**

None. The core reasoning holds: the reconcile branch is only reachable when `_has_delta_in_version()` is false, which (absent `override.version`) means `desired <= current`, so the resolved target `current_version[0]` is always `>=` what the old code would have sent. The explicit-downgrade path (`override.version` set) still routes through `_has_delta_in_version() -> True -> _validate_version_upgrade()` unchanged, and `test_explicit_downgrade_still_blocked` / `test_explicit_downgrade_allowed_with_force` pin both halves.

**Suggestions**

1. `azext_edge/edge/providers/orchestration/upgrade2.py` — `_validate_and_resolve_version`, the `_has_non_success_state()` branch. The full guard now runs **only** when `_has_delta_in_train()` is true. That silently relaxes the non-stable-train guard at lines 1372-1378 of the pre-change file (`if current_is_preview or desired_is_preview: ...`): an extension installed on a `preview` train, in `Failed` state, with no train delta, previously raised *"Upgrades to or from non-stable release trains are not supported"* and now gets a same-version reconcile patch. I believe that is **correct and intended** — reconciling in place on the train you are already on is not a train transition, and the old raise was really the downgrade check firing first on a stale built-in. But it is a real behavior change on a preview-train cluster and nothing in the new test block pins it. `test_reconcile_cannot_bypass_release_train_guard` only covers the two cases where a delta *exists*; `test_validate_upgrade_and_get_patch_cannot_diverge` only asserts the two call sites agree, not what the outcome is. Please add the positive case: `current_train="preview"`, `built_in_train="preview"`, `built_in_version` below current, `provisioning_state=Failed`, no override → `validate_upgrade()` succeeds and `get_patch()["properties"]["version"] == current`. Without it, a future re-tightening of this branch would pass CI.

2. `azext_edge/edge/providers/orchestration/upgrade2.py` — `get_stale_cli_extensions()` iterates every entry in `extension_upgrades` and does not filter on `can_upgrade()`. `is_cli_behind_cluster()` guards on `operation_type != UPDATE`, but an UPDATE-typed extension with no deltas at all (nothing to do) still contributes a moniker to the advisory. On a fully up-to-date cluster running an older CLI, the user gets the "this CLI has no newer version to offer" paragraph immediately followed by `Nothing to upgrade :)`. That combination is a bit alarming for a genuine no-op. Consider `if ext_state.can_upgrade() and ext_state.is_cli_behind_cluster()`, or leaving it but confirming the wording reads sensibly in that case.

3. `azext_edge/edge/providers/orchestration/upgrade2.py` — `is_cli_behind_cluster()` catches `ValueError` only. `semver.version.Version.parse` (via `scoped_semver_import()` in `azext_edge/edge/util/machinery.py`) raises `TypeError` when handed a non-string, which is reachable if ARM ever returns a non-string `properties.version`. `except (ValueError, TypeError)` makes the advisory strictly best-effort, which is the right posture for a purely informational warning that runs before the `has_upgrades()` early return.

**Nits**

1. `azext_edge/edge/providers/orchestration/upgrade2.py` — the new `logger.warning(...)` uses lazy `%s` interpolation while the rest of the module uses f-strings (`logger.debug(f"...")`). Lazy interpolation is the better habit; just noting the inconsistency in case the repo prefers one form.

2. `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — `assert_no_upgrades_performed` was loosened from `assert_called_once_with` to `assert_called_with`, which no longer asserts the warning count for every existing caller. `test_stale_cli_warning_precedes_no_op_return` does check `len(warning_calls) == 2` for the new scenario, so the coverage is not lost, but the relaxation is global. An alternative is asserting `DEFAULT_LOG_WARNING_MESSAGE` is the *last* warning while keeping a count assertion the callers can opt into.

3. `test_validate_upgrade_and_get_patch_cannot_diverge` asserts `compared == 1800` — a hardcoded product count that must be hand-updated whenever a parameter list grows. `assert compared == len(versions) ** 2 * len(trains) ** 2 * 2 * 2 * 2` (or just `assert compared` ) expresses the same intent without the maintenance trap.

**Summary**

Solid, well-reasoned fix with genuinely strong test coverage — the `test_validate_upgrade_and_get_patch_cannot_diverge` invariant test is a good way to pin the shared-condition bug that caused this. The one thing I would want before merge is a test pinning the preview-train reconcile case (Suggestion 1), since that guard now depends on a train delta being present.

cheatsheet1999:
> ### Azure SRE Agent - automated review
@zzcodework SRE Agent - review the latest commit
Done:
Suggestion 1 - added `test_reconcile_repairs_a_failed_extension_on_a_preview_train` with exactly the case you specified.
Suggestion 3 -  now  except (ValueError, TypeError), plus `test_stale_cli_check_is_best_effort_on_a_malformed_version`.
Nit 3 -  Now uses five representative parametrized cases.

Not changed:
Suggestion 2, Nit 1 and Nit 2. filtering on `can_upgrade()`  suppresses the advisory in precisely the case that prints `Nothing to upgrade :)`

zzcodework:

Reviewing `changes since 6ae49a07` (head `b62de959`)

The incremental change is one source line plus three test edits, all of which map directly to points from the previous review. I re-read `is_cli_behind_cluster`, `_validate_and_resolve_version` and the test helper `build_ext_upgrade_state` at this head before commenting.

**Blocking**

None.

**Suggestions**

None. Checking each item that changed:

1. `azext_edge/edge/providers/orchestration/upgrade2.py` — `except (ValueError, TypeError)` in `is_cli_behind_cluster()` is exactly the fix I asked for, and it is reachable in the new test: `build_ext_upgrade_state` passes `operation_type=None`, and `ExtensionUpgradeState.__init__` normalises that to `ExtensionOperation.UPDATE`, so `test_stale_cli_check_is_best_effort_on_a_malformed_version` does get past the operation-type guard and into the `semver.parse` call for `1.4`, `14073` and `["1.4.73"]`. It is a real assertion, not a vacuous one — worth stating explicitly because the same test would silently pass for the wrong reason if that default ever changed to a non-UPDATE operation.

2. `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — `test_reconcile_repairs_a_failed_extension_on_a_preview_train` pins the case I was worried about: `preview` on both sides, built-in below current, `Failed`, no override → `validate_upgrade()` does not raise and the patch carries `1.4.73`. That closes the gap where a future re-tightening of the non-stable-train guard would have passed CI.

3. `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py` — replacing the 1800-case product in `test_validate_upgrade_and_get_patch_cannot_diverge` with five named `pytest.param` cases removes the hardcoded count trap and keeps the intent readable. The five ids (`downgrade`, `upgrade`, `reconcile`, `train-guard`, `no-installed-version`) cover each branch that can produce a divergence, and the `force=True` path is still pinned separately by `test_explicit_downgrade_allowed_with_force`, so nothing meaningful is lost with the exhaustive sweep gone.

**Nits**

None new.

**Summary**

Clean iteration — all three addressed points are correctly implemented and genuinely tested. I read the reply on the three items that were deliberately not changed (Suggestion 2, Nits 1-2) and have no objection; the `can_upgrade()` filtering argument in particular is answered separately in this thread. Nothing here blocks merge from my side.

zzcodework:
### Azure SRE Agent - reply

> Not changed: Suggestion 2, Nit 1 and Nit 2. filtering on `can_upgrade()` suppresses the advisory in precisely the case that prints `Nothing to upgrade :)`

You are right and I withdraw Suggestion 2. Re-reading the call site, the advisory block runs before the `if not upgrade_state.has_upgrades(): logger.warning("Nothing to upgrade :)")` early return, and `has_upgrades()` is itself `any(ext_state.can_upgrade() ...)` — so gating `get_stale_cli_extensions()` on `can_upgrade()` would silence the message in the one scenario where the user most needs it: an up-to-date cluster where the reason nothing is offered is that the CLI is behind. The current ordering is the point of the change, not an accident of it. Nits 1 and 2 were stylistic and I have no further argument on either.

I verified the three items you did change and they hold up — details in my review of `b62de959`. In particular `test_stale_cli_check_is_best_effort_on_a_malformed_version` really does exercise the new `TypeError` arm, since `ExtensionUpgradeState.__init__` normalises `operation_type=None` to `UPDATE` and the check gets past its first guard.

_This is an automated reply and may be incomplete._

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1155,10 +1197,33 @@ def validate_upgrade(self) -> None:
         Raises:
             ValidationError: If the upgrade is not valid (e.g., downgrade, incompatible versions).
         """
-        # Always validate if there's an override version (user explicitly requested upgrade)
-        # or if there's a version delta or non-success state
-        if self._has_delta_in_version() or self._has_non_success_state():
+        if self.operation_type != ExtensionOperation.UPDATE:
+            return
+
+        self._validate_and_resolve_version()
```
digimaun, line 1202: `validate_upgrade()` only runs for the ops extension at :945. On the table path, certManager and secretStore first hit this logic when `format_extension_row()` calls `get_patch()` at :495, where the broad handler renders `Unknown / Unknown / Check configuration`. With `--no-progress` there is no table, so the first call is :418 during apply, and that loop catches only `HttpResponseError`. A failed non-ops extension with no `properties.version` still reaches it, so the command can abort after the user confirmed, potentially with earlier operations already applied.

This is pre-existing and the PR removes the stale CLI trigger for it, so I do not think it should block. Now that `validate_upgrade()` returns early for CREATE and DELETE, a follow-up could call it for every extension state during analysis and move these failures ahead of the prompt.

The description should narrow the claim that this clears the `Unknown / Unknown` cascade for certManager and secretStore. It clears the stale CLI cause. A missing installed version still produces the same row.

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -480,17 +491,27 @@ def format_extension_row(ext: "ExtensionUpgradeState") -> Tuple[str, str, str]:
 
         # UPDATE operation
         current = format_version_with_train(ext.current_version[0], ext.current_version[1]) + status_indicator
-        desired = f"[cyan]{format_version_with_train(ext.desired_version[0], ext.desired_version[1])}[/cyan]"
 
         patch = ext.get_patch()
+        props = patch.get("properties", {})
+
+        desired = "[cyan]{}[/cyan]".format(
```
digimaun: The field by field fallback is right. On the reported config only path `can_upgrade()` is true because config changes, so this row does render, and it now shows Current `1.3.137 [stable]` against Desired `1.3.137 [stable]` with the desired side still cyan. Cyan is the change signal everywhere else in this table. Could we leave it unstyled or dim it when `props` carries neither `version` nor `releaseTrain`?

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -87,6 +87,17 @@ def upgrade_ops_instance(
 
     upgrade_state = upgrade_manager.analyze_cluster(**kwargs)
 
+    stale_monikers = upgrade_state.get_stale_cli_extensions()
+    if stale_monikers:
+        logger.warning(
+            "The version of %s built into this azure-iot-ops CLI extension is older than the "
+            "version deployed on the cluster, so this CLI has no newer version to offer. "
+            "Nothing will be downgraded: a lower version is only applied when explicitly "
```
digimaun: `--force` is hidden at `params.py:1538`, and the lower version this describes also needs that extension's version override, hidden for all three at `:1521`. Each CLI release pins and tests AIO alongside its dependencies as a set, `template.py:692` for certManager and secretStore and `:1478` for AIO. Pointing at that escape path from an ordinary warning can lead someone outside the version set this release was tested with.

Could we keep the outcome and drop the mechanism, something like "No deployed extension version will be changed"?

Minor on the same string. `", ".join(...)` reads oddly with two entries, as in "The version of certManager, secretStore ... is older than the version".

azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -1155,10 +1197,33 @@ def validate_upgrade(self) -> None:
...
+            return self.desired_version[0]
+
+        if self._has_non_success_state():
+            reconcile_version = self._get_reconcile_version()
+            if not reconcile_version:
+                raise ValidationError(
+                    f"Unable to determine installed version for {self.moniker} extension. "
+                    "Cannot validate upgrade path."
+                )
+            if self._has_delta_in_train():
```
digimaun, line 1219: Keeping target selection in `_validate_and_resolve_version()` matters because validation and patch construction now resolve the same version, and that is the property worth protecting in future cleanups. Could we add a short comment on why this call is conditional? A failed extension can have a version but no `releaseTrain`, and calling `_validate_version_upgrade()` unconditionally then raises `Unable to determine release train` even though the patch only reapplies the current version and sends no train.

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
```diff
@@ -2800,3 +2806,345 @@ def test_opcua_connector_version_matches_template_tag():
...
+        pytest.param({"provisioning_state": PROVISIONING_STATE_FAILED}, id="reconcile"),
+        pytest.param(
+            {"provisioning_state": PROVISIONING_STATE_FAILED, "built_in_train": "preview"}, id="train-guard"
+        ),
+        pytest.param(
+            {"provisioning_state": PROVISIONING_STATE_FAILED, "current_version": None}, id="no-installed-version"
+        ),
+    ],
+)
+def test_validate_upgrade_and_get_patch_cannot_diverge(ext_type, kwargs):
```
digimaun: `outcome()` compares only exception behavior, and the successful patch targets are already covered by `test_get_patch_version_selection` above. Could we rename this to something like `test_validate_upgrade_and_get_patch_have_same_validation_outcome` and adjust the docstring? `validate_upgrade()` is the early gate for ops at :945 but not for certManager.

zzcodework:

Reviewing `changes since b62de959` (head `7e78ce7c`)

One commit (`address tech lead review`), touching `format_extension_row`, the stale-CLI advisory, one comment in `_validate_and_resolve_version`, and three test edits. I re-read `format_extension_row` and the `format_extension_row` tests at this head before commenting.

**Blocking**

None.

**Suggestions**

1. `azext_edge/edge/providers/orchestration/upgrade2.py` — the new `desired_style` branch is the actual user-visible bug fix in this iteration (config-only patch no longer renders an unchanged version in cyan), but nothing asserts it. `test_desired_state_ignores_built_in_target_when_no_version_is_sent` is exactly the config-only case and already calls `format_extension_row`, yet it only checks which version string appears — it would still pass if the style regressed to cyan. One extra line there (`assert "[dim]" in desired` / `assert "cyan" not in desired`) pins the fix at zero cost. Without it the regression this commit fixes can silently come back.

**Nits**

1. `azext_edge/edge/providers/orchestration/upgrade2.py` — the comment says cyan is used "only when the patch actually moves the version", but the condition is key presence (`"version" in props or "releaseTrain" in props`), so the reconcile case — `props["version"] == ext.current_version[0]`, i.e. a same-version retry of a failed extension — still renders cyan. That looks intentional (the action line for that path is also cyan, "Reconcile at version X", and a reconcile *is* a change being applied), so I read this as a comment/code wording mismatch rather than a logic bug. Suggest wording it as "only when the patch sends a version or train" to match what the code checks.

**Summary**

Clean, well-scoped iteration. The stale-CLI advisory rewording is consistent with its test — the test now asserts the new phrasing and that `--force` is absent, and no other reference to the old wording survives in `upgrade2.py` or the unit test file. The `_has_delta_in_train()` comment accurately describes the failed-extension-with-no-train case it guards, and the test rename matches what the test actually asserts (validation outcome only). My only substantive point is the missing style assertion for the fix itself.

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: The dim and cyan split is right. I checked the four row types. Config only now renders dim, while a real upgrade, a reconcile, and a train only change stay cyan.

The test suite does not pin the style. I set `desired_style = "cyan"` unconditionally and the full `test_upgrade2_unit.py` still passed, with 262 tests. Could we add a style assertion to `test_desired_state_ignores_built_in_target_when_no_version_is_sent`? It already covers the config only patch and calls `format_extension_row()`.

Small thing on the comment above it. It says cyan is used when the patch moves the version, but the condition is key presence. A train only change and a same version reconcile are cyan too. Wording it as "sends a version or train" would match the code.

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: The internal flags are out of the warning. One thing on the reassurance, and it came from my suggested wording. `stale_monikers` can be a subset, so on the cluster shape from the report the advisory names iotOperations while certManager and secretStore still take real version upgrades in the same run, which makes "No deployed extension version will be changed" false for the command as a whole. Could we scope it to the named extensions, something like "No deployed version of the listed extensions will be changed"?

azext_edge/edge/providers/orchestration/upgrade2.py
digimaun: The narrowed wording in the description is accurate, including the note that the apply loop is unchanged.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: Yeah, both of those are in already. Style assertion is in `test_desired_state_ignores_built_in_target_when_no_version_is_sent`. On the comment, I ended up deleting it rather than rewording. It was describing the wrong condition, and the assertion pins the behavior now. Can put a corrected one back if you'd rather have it there. Thank you.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: You're right, thanks. I went with your wording.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: Added, pretty much your wording, thank you!

azext_edge/tests/edge/orchestration/test_upgrade2_unit.py
cheatsheet1999: Renamed to `test_validate_upgrade_and_get_patch_have_same_validation_outcome`, and the docstring now says it's comparing validation outcome only.

azext_edge/edge/providers/orchestration/upgrade2.py
cheatsheet1999: Cool, thanks for re-reading it. Really appreciate it.

And yeah, happy to take the follow-up, calling `validate_upgrade()` for every extension during analysis so those failures surface before the prompt. Agreed it’s better kept out of this PR since it changes behavior on a few paths. I can open a follow-up PR.

zzcodework:

Reviewing `changes since 7e78ce7c` (head `425bd5a`)

Note on method: `GET /compare/7e78ce7c...425bd5a3` returns a misleading delta here — `7e78ce7c` is no longer in this PR's commit list (the branch was rebased/amended), so the merge base falls back and the compare replays changes that were already reviewed. I diffed the two file blobs directly instead, so the increment below is the real one.

Actual increment — two commits (`code improvement`, `Adjust wording on messages`), 3 wording edits + 1 comment deletion in `azext_edge/edge/providers/orchestration/upgrade2.py`, 3 assertion edits in `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py`.

**Blocking**

None.

**Suggestions**

None. Both points from my previous pass are resolved:

- The missing style assertion is now pinned — `test_desired_state_ignores_built_in_target_when_no_version_is_sent` gained `assert "[dim]" in desired` / `assert "cyan" not in desired`, which is exactly the config-only row whose dim rendering this PR fixes.
- The comment/code mismatch I flagged ("only use it when the patch actually moves the version", while the condition is key presence) was resolved by deleting the comment rather than rewording it. That is a fine outcome — `desired_style = "cyan" if ("version" in props or "releaseTrain" in props) else "dim"` reads directly, and the reconcile-still-cyan behaviour no longer has a comment contradicting it.

**Nits**

None.

**Verification of the wording changes**

- Advisory: `"No deployed extension version will be changed"` → `"No deployed version of the listed extensions will be changed"`. This is the narrower and more accurate claim — `stale_monikers` can be a subset of the extensions in the run, so the old phrasing over-promised for extensions not in the `%s` list. The `%s` list appears in the first sentence, so "the listed extensions" resolves correctly. `test_stale_cli_warning_precedes_no_op_return` was updated in the same commit and no reference to the old string survives in either file.
- The rewritten comment above `if self._has_delta_in_train():` is factually correct: with a failed extension that has a version but no train, `_validate_version_upgrade()` hits `if not self.current_version[1]` and raises "Unable to determine release train for installed {moniker} extension", while the patch built on that path only reapplies `reconcile_version` and sends no `releaseTrain`.

**Summary**

Polish-only iteration: no behavioural change beyond the two operator-facing message strings, and every edit is matched by the corresponding test assertion. Nothing further from me on this pass.

digimaun:
The code is in good shape. Four spots in the description have drifted from the final state, worth fixing since the description outlives the review.

Changes 1 says the change signal is reserved for a version that is actually moving. The condition is key presence, so a same version reconcile and a train only patch are cyan too. "reserved for patches that send a version or release train" would match.

Changes 2 still carries the unscoped wording, "no deployed extension version will be changed". The code now says "No deployed version of the listed extensions will be changed", which is the phrasing that stays accurate when other extensions upgrade in the same run.

Changes 3 says the reconcile branch is reachable when there is no override and `desired <= current`. `_has_delta_in_version()` is also false when the current or desired version is missing, so "no positive version delta" covers all of it.

The implementation note says the invariant test pins that the two cannot disagree. After the rename it pins that they reach the same validation outcome, which is narrower.

digimaun (approved):

PR #927, opened 2026-09-01
feat(live-data): add az iot ops live-data command group

ketkimnaik:
- Adds the preview az iot ops live-data command group (`enable`,  `show`,  `disable`) to configure and inspect Live Data observability for an IoT Operations instance.

- Delivers a complete, idempotent lifecycle: enabling wires up the supporting Event Grid, Device Registry, and dataflow resources; `show` reports consolidated status; `disable` cleanly tears everything down.

- Extracts the shared Event Grid and identity setup into a common base so Live Data and Management Actions stay consistent and avoid duplicated behavior.

- Hardens the disable flow so an interrupted run can be safely re-run without leaving orphaned resources (applies to both Live Data and Management Actions).

- Includes full unit-test coverage and live integration tests for the new command group.

Enable:

image

image

image

Staged Enable:

image

Show:

image

image

Disable:

image

image

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_edge/edge/_help.py
+++ b/azext_edge/edge/_help.py
@@ -2526,9 +2526,10 @@ def load_iotops_help():
         type: command
         short-summary: Disable Live Data for an IoT Operations instance.
         long-summary: |
-            Removes this instance's observability endpoint entry from the Device Registry
-            namespace first (the per-instance disable signal), then tears down the dedicated
-            dataflow profile, the EG dataflow endpoint, and the observability topic space.
+            Tears down the dedicated dataflow profile, the EG dataflow endpoint, and the
+            observability topic space, then removes this instance's observability endpoint
+            entry from the Device Registry namespace last so an interrupted disable can be
+            safely re-run.
 
             Namespace-scoped role assignments are preserved; topic-space-scoped role
             assignments are removed together with the topic space.
--- a/azext_edge/edge/_help.py
+++ b/azext_edge/edge/_help.py
@@ -2473,6 +2473,93 @@ def load_iotops_help():
     """
 
     helps[
+        "iot ops live-data"
+    ] = """
+        type: group
+        short-summary: Live Data infrastructure configuration for an IoT Operations instance.
+    """
+
+    helps[
+        "iot ops live-data enable"
+    ] = """
+        type: command
+        short-summary: Enable Live Data for an IoT Operations instance.
+        long-summary: |
+            Provisions and configures the shared, instance-level infrastructure required for
+            Live Data across three domains:
+            - Event Grid Namespace: an observability topic space.
+            - IoT Operations Instance: a dedicated dataflow profile and an EG MQTT dataflow endpoint.
+            - Device Registry Namespace: the outbound identity and the per-instance observability
+              endpoint entry (keyed by the instance's custom location).
+
+            The command is idempotent. Re-running converges to the desired state without
+            overwriting unrelated entries. Enablement for an instance is expressed by the presence
+            of its observability endpoint entry.
+
+            By default, role assignments (Event Grid TopicSpaces Publisher for the instance identity
+            and Subscriber for the ADR namespace identity) are created at the Event Grid namespace
+            scope. Use --ra-scope topic-space for least-privilege assignments scoped to the topic
+            space, or --skip-ra to skip role assignment creation.
+
+        examples:
+        - name: Enable Live Data using a system-assigned managed identity and default roles.
+          text: >
+            az iot ops live-data enable --instance myinstance -g myresourcegroup
+            --eg-resource-id $EG_NAMESPACE_RESOURCE_ID
+        - name: Enable Live Data using a user-assigned managed identity for outbound auth.
+          text: >
+            az iot ops live-data enable --instance myinstance -g myresourcegroup
+            --eg-resource-id $EG_NAMESPACE_RESOURCE_ID --mi-user-assigned $UA_MI_RESOURCE_ID
+        - name: Enable Live Data with least-privilege, topic-space-scoped role assignments.
+          text: >
+            az iot ops live-data enable --instance myinstance -g myresourcegroup
+            --eg-resource-id $EG_NAMESPACE_RESOURCE_ID --ra-scope topic-space
+        - name: Enable Live Data and skip role assignments.
+          text: >
+            az iot ops live-data enable --instance myinstance -g myresourcegroup
+            --eg-resource-id $EG_NAMESPACE_RESOURCE_ID --skip-ra
+    """
+
+    helps[
+        "iot ops live-data disable"
+    ] = """
+        type: command
+        short-summary: Disable Live Data for an IoT Operations instance.
+        long-summary: |
+            Removes this instance's observability endpoint entry from the Device Registry
+            namespace first (the per-instance disable signal), then tears down the dedicated
+            dataflow profile, the EG dataflow endpoint, and the observability topic space.
+
+            Namespace-scoped role assignments are preserved; topic-space-scoped role
+            assignments are removed together with the topic space.
+
+        examples:
+        - name: Disable Live Data for an instance.
+          text: >
+            az iot ops live-data disable --instance myinstance -g myresourcegroup
+        - name: Disable Live Data without a confirmation prompt.
+          text: >
+            az iot ops live-data disable --instance myinstance -g myresourcegroup --yes
+    """
+
+    helps[
+        "iot ops live-data show"
+    ] = """
+        type: command
+        short-summary: Show Live Data configuration for an IoT Operations instance.
+        long-summary: |
+            Reports the Live Data configuration and resource status across the Device Registry
+            namespace, Event Grid, and IoT Operations dataflow resources. The top-level enabled
+            flag is true only when this instance's observability endpoint entry and all supporting
+            resources are present.
+
+        examples:
+        - name: Show Live Data configuration for an instance.
+          text: >
+            az iot ops live-data show --instance myinstance -g myresourcegroup
+    """
+
+    helps[
         "iot ops schema"
     ] = """
         type: group
--- a/azext_edge/edge/providers/orchestration/common.py
+++ b/azext_edge/edge/providers/orchestration/common.py
@@ -461,3 +461,16 @@ MGMT_ACTIONS_ADR_ENDPOINT_TYPE = "Microsoft.EventGrid/Namespaces"
 MGMT_ACTIONS_GRAPH_ARTIFACT = "azureiotoperations/graph-dataflow-map:1.0.0"
 MGMT_ACTIONS_GRAPH_RULES_VERSION = "1.0.0"
 MIN_EG_CLIENT_SESSIONS_PER_AUTH_NAME = 4
+
+# Live Data
+LIVE_DATA_ADR_API_VERSION = "2026-11-02-preview"
+LIVE_DATA_PROFILE_NAME = "live-data-profile"
+LIVE_DATA_ENDPOINT_NAME = "live-data-endpoint"
+LIVE_DATA_TOPICSPACE_PREFIX = "live-data-ts"
+LIVE_DATA_TOPIC_TEMPLATE = "aio/observabilitySessions/{scope_id}/#"
+LIVE_DATA_ADR_ENDPOINT_TYPE = "Microsoft.EventGrid/namespaces"
+
+
+class LiveDataRoleScope(Enum):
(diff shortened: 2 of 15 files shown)


azext_edge/edge/_help.py
```diff
@@ -2472,6 +2472,93 @@ def load_iotops_help():
...
+
+    helps[
+        "iot ops live-data disable"
+    ] = """
+        type: command
+        short-summary: Disable Live Data for an IoT Operations instance.
+        long-summary: |
+            Removes this instance's observability endpoint entry from the Device Registry
+            namespace first (the per-instance disable signal), then tears down the dedicated
+            dataflow profile, the EG dataflow endpoint, and the observability topic space.
```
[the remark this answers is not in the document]
ketkimnaik: Done

tox.ini
```diff
@@ -71,6 +71,7 @@ description =
     rpsaas: marked cloud
     upgrade: marked upgrade
     mgmtactions: marked mgmt-actions
+    livedata: marked live-data
```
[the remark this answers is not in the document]
ketkimnaik: Not needed

azext_edge/edge/providers/orchestration/live_data.py
```diff
@@ -0,0 +1,1043 @@
...
+        confirm_yes: Optional[bool] = None,
+        no_progress: Optional[bool] = None,
+        **kwargs,
+    ) -> None:
+        """Disable Live Data for an IoT Operations instance.
+
+        Removes this instance's observability endpoint entry first (per-instance disable),
+        then tears down the dedicated dataflow profile, EG dataflow endpoint, and topic
+        space. Namespace-scoped role assignments are preserved; topic-space-scoped roles
+        are removed together with the topic space.
```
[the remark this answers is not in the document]
ketkimnaik: Done

zzcodework:

Reviewing `full diff` (head `06095a27`)

New `az iot ops live-data` command group (enable/show/disable) plus extraction of the shared Event Grid logic into `EventGridProviderBase`. The refactor of `mgmt_actions.py` onto the new base looks faithful — `_validate_eg_namespace` keeps the mgmt-actions-only `MIN_EG_CLIENT_SESSIONS_PER_AUTH_NAME` check via the `min_client_sessions` override, `_setup_eg_dataflow_endpoint` keeps the hash-derived endpoint name, and the removed imports (`ConnectedCluster`, `MQTT_ENDPOINT_TYPE`, `CUSTOM_LOCATIONS_API_VERSION`, `MANAGED_IDENTITY_API_VERSION`, `InvalidArgumentValueError`) are all genuinely unused now. `--skip-ra`, `--yes` and `--no-progress` resolve from the global `iot ops` argument context, and `--instance/-i/-n` matches the `mgmt-actions` group. Unit coverage is broad.

**Blocking**

1. `azext_edge/edge/providers/orchestration/live_data.py` — `_setup_adr_observability()`, the early-return branch (`if endpoint_already_configured and (not needs_identity_enable):`) never converges `outboundIdentity`, so switching the outbound identity on an already-enabled instance silently no-ops on the ADR namespace.

   `desired_endpoint` contains only `endpointType`/`address`/`scopeId`/`resourceId` — nothing identity-related — so `endpoint_already_configured` stays `True` when only the identity changes. Concretely, for an instance already enabled with SAMI, re-running with `--mi-user-assigned <id>`:
   - `use_sami = False` → `needs_identity_enable = False`;
   - `endpoint_already_configured = True` (endpoint dict unchanged);
   - the method returns before any `_patch_namespace()` call, so `properties.outboundIdentity` remains `{"type": "SystemAssigned"}`;
   - but the returned dict reports `"outboundIdentity": outbound_identity` (the *desired* UAMI), so `enable` output claims a state that was never written.

   The same holds in reverse (enabled with UAMI, re-run without `--mi-user-assigned` while the namespace already has a SystemAssigned identity), and for swapping one UAMI for another. This is worse than a no-op because `_setup_eg_dataflow_endpoint()` *does* converge its `authentication` block, so the run ends with the dataflow endpoint on the new identity and the ADR namespace still on the old one — a half-migrated, non-working configuration reported as success. Suggested fix: include `properties.get("outboundIdentity") == outbound_identity` in the early-return condition (and drop into the normal write path otherwise). None of the unit tests cover an identity switch — `test_already_configured_early_return` and `test_uami_single_write` both start from a matching identity — so a regression test for the switch case would be worth adding alongside.

**Suggestions**

2. `azext_edge/edge/providers/orchestration/common.py` / `live_data.py` — the topic space *name* is uniquified per instance (`get_live_data_topic_space_name()` hashes the instance ARM resource ID), but the topic *template* it is created with is keyed only on the instance **name**: `LIVE_DATA_TOPIC_TEMPLATE = "aio/observabilitySessions/{scope_id}/#"` formatted with `instance_name` in `_setup_topic_space()`, and the ADR endpoint's `scopeId` is likewise `instance.get("name")`. Two instances with the same name in different resource groups/subscriptions sharing one Event Grid namespace therefore get two distinct topic spaces covering the *identical* topic filter. With the default `--ra-scope namespace` that is already broad, but it specifically undercuts `--ra-scope topic-space`, which the help sells as "least-privilege": instance A's publisher identity, scoped to A's topic space, still covers B's session topics. If DOE requires the plain instance name as `scopeId` this is worth an explicit comment; otherwise consider using the same hash for the topic segment.

3. `live_data.py` — `enable()` is documented (help long-summary and the `enable()` docstring) as idempotent and "converges to the desired state", but `_setup_topic_space()` and `_setup_dataflow_profile()` only test for *existence*: an existing topic space with drifted `topicTemplates`, or a profile with a changed `instanceCount`/diagnostics, is reported as `exists` and left untouched — while the returned `topicTemplates` are the desired ones, not the actual ones. `_setup_eg_dataflow_endpoint()` in the base class does compare-and-update, so the three are inconsistent. Either converge them the same way, or soften the wording to "created if missing" for those two resources.

4. `azext_edge/edge/providers/orchestration/live_data.py` — `enable()` wraps the ADR and role-assignment stages in `except Exception as exc: display.update_step(..., str(exc)[:40]); raise`. Truncating to 40 characters is fine for the progress cell, but for an ARM `HttpResponseError` the useful part (status/code) is usually past that cut and the exception is re-raised anyway, so it only affects the display — worth confirming the full error still surfaces to the user in `--debug`-less runs.

**Nits**

5. `HISTORY.rst` is not touched. The repo checklist requires an entry for user-facing changes, and a brand-new preview command group qualifies.

6. `azext_edge/edge/providers/adr/namespaces.py` — the `if tags:` guard is an unrelated drive-by fix in this PR. It is a behavior change (`--tags` no longer sends an explicit `null` on create); fine on its own, but it deserves its own mention in the changelog entry since it is not part of live-data.

Skipping the docstring/help ordering points already raised by Copilot and marked Done.

azext_edge/edge/providers/orchestration/live_data.py
```diff
@@ -0,0 +1,1044 @@
...
+        sami_already_enabled = current_identity_type == "systemassigned"
+
+        use_sami = mi_resource is None
+        needs_identity_enable = use_sami and not sami_already_enabled
+
+        base_payload: Dict = {"properties": {"outboundIdentity": outbound_identity}}
+        if needs_identity_enable:
+            base_payload["identity"] = {"type": "SystemAssigned"}
+
+        if endpoint_already_configured and (not needs_identity_enable):
```
HangyiWang: do we need to check the outbound_identity here before return?

azext_edge/edge/providers/orchestration/live_data.py
```diff
@@ -0,0 +1,1044 @@
...
+    outboundIdentity, management, and messaging from the original namespace.
+    """
+    properties = adr_namespace.get("properties", {})
+    existing_endpoints = properties.get("observability", {}).get("endpoints", {})
+    updated_endpoints = {k: v for k, v in existing_endpoints.items() if k != custom_location_id}
+
+    payload: Dict = {
+        "location": adr_namespace.get("location", ""),
+        "properties": {
+            "observability": {"endpoints": updated_endpoints},
```
HangyiWang: do we have other properties in `observability` needs to be preserved here since it's a PUT?

azext_edge/edge/providers/orchestration/common.py
```diff
@@ -461,3 +461,16 @@ class TrustedSigningKeyType(Enum):
...
 MGMT_ACTIONS_GRAPH_RULES_VERSION = "1.0.0"
 MIN_EG_CLIENT_SESSIONS_PER_AUTH_NAME = 4
+
+# Live Data
+LIVE_DATA_ADR_API_VERSION = "2026-11-02-preview"
+LIVE_DATA_PROFILE_NAME = "live-data-profile"
+LIVE_DATA_ENDPOINT_NAME = "live-data-endpoint"
+LIVE_DATA_TOPICSPACE_PREFIX = "live-data-ts"
+LIVE_DATA_TOPIC_TEMPLATE = "aio/observabilitySessions/{scope_id}/#"
+LIVE_DATA_ADR_ENDPOINT_TYPE = "Microsoft.EventGrid/namespaces"
```
HangyiWang, line 471: see line 460, `Namespaces` or `namespaces`?

azext_edge/edge/providers/orchestration/live_data.py
ketkimnaik: Good catch — added an outboundIdentity match to the early-return guard so a mismatched identity now falls through and gets corrected.

azext_edge/edge/providers/orchestration/live_data.py
ketkimnaik: Right — now copies the existing observability object and replaces only endpoints, preserving any sibling keys. Fixed in latest commit.

azext_edge/edge/providers/orchestration/common.py
ketkimnaik: Intentional — live-data uses the ARM-canonical lowercase Microsoft.EventGrid/namespaces (validated end-to-end); it's a separate endpoint map from mgmt-actions and ADR accepts both.

zzcodework:

Reviewing `changes since 06095a27` (head `f1af896`)

The increment is small and targeted: it addresses the two inline review points raised on the previous iteration (preserving sibling keys under `observability` in the full PUT, and checking `outboundIdentity` before the early return). Both fixes look correct.

**Blocking**

None.

**Suggestions**

- `azext_edge/edge/providers/orchestration/live_data.py` — the new `outbound_identity_already_configured` guard is the actual behavior change (a namespace whose endpoint matches but whose `outboundIdentity` differs now falls through and gets corrected), yet the only test change is adding `outbound_identity={"type": "SystemAssigned"}` to `test_already_configured_early_return` so that existing test keeps passing. There is no test covering the newly-reachable path. Worth adding a case with a matching endpoint but a mismatched `outboundIdentity` (e.g. stored `{"type": "UserAssigned", ...}` while `mi_resource is None`, or vice versa) asserting that a PATCH is issued and its body carries the corrected `properties.outboundIdentity`. Without it, a future refactor can silently restore the old early-return and no test fails.
- `azext_edge/edge/providers/orchestration/live_data.py` — the sibling-preservation fix in `_build_adr_observability_put_payload` is likewise untested. The two existing tests for that helper (`test_live_data_unit.py`, the `_build_adr_observability_put_payload` block) do not appear to feed an `observability` object containing keys other than `endpoints`, so `dict(observability)` vs the old `{"endpoints": ...}` is indistinguishable to the suite. A one-line addition — put e.g. `{"endpoints": {...}, "<sibling>": ...}` in the mocked namespace and assert the sibling survives into the payload — locks the fix in.

**Nits**

- `azext_edge/edge/providers/orchestration/live_data.py` — in the non-staged write path, `_setup_adr_observability` still builds `payload["properties"]["observability"] = {"endpoints": merged_endpoints}`, i.e. it does not do the same sibling copy the PUT helper now does. This is fine as long as `_patch_namespace` (`namespaces.begin_update`) is a genuine merge-PATCH, since the service merges the nested object; just flagging the asymmetry so it is a deliberate choice rather than an oversight, given the same shape was just fixed a few lines above.
- In the mismatched-`outboundIdentity` fall-through, the returned `endpoint_exists` is still `endpoint_already_configured` (`True`) even though a write was performed. That matches the "state before the call" reading of the flag, but if any caller uses it as "no write happened", it would now be misleading.

Summary: correct, minimal fixes for both prior review comments; the main gap is that neither fix has a test that would fail if it were reverted.

HangyiWang (approved):

PR #928, opened 2026-09-01
feat: add on-demand alpha build skill

ketkimnaik:
## Summary
- add a resumable on-demand alpha build skill covering release-input sync, PR review, integration tests, and private wheel publication
- add explicit safety gates before remote writes and production approval
- verify required tests-repository PRs/commits are included in the selected immutable ref and survive compilation
- add key-agnostic OPC UA connector tag verification to the template-sync skill

## Motivation
An active service-team PR is not part of its target release branch until merged. The new source-inclusion guard prevents such changes from being silently assumed present and distinguishes narrow release-policy exclusions from independent compiled behavior.

## Validation
- parsed skill YAML frontmatter with PyYAML
- checked both skill files with VS Code diagnostics
- ran `git diff --cached --check`

No generated templates or release constants are included in this PR.


.github/skills/on-demand-alpha-build/SKILL.md
```diff
@@ -0,0 +1,194 @@
+---
+name: on-demand-alpha-build
+description: "Orchestrate an on-demand Azure IoT Operations CLI alpha release from a chosen azure-iot-operations-tests ref: refresh generated release inputs, create a PR, run integration tests after merge, then start the build and wheel-publish workflow and notify approvers."
```
HangyiWang: oh it can also notify approvers now?

HangyiWang (dismissed):

zzcodework:

Reviewing `full diff` (head `9c1b9d8`)

Docs-only change: adds `.github/skills/on-demand-alpha-build/SKILL.md` and appends an OPC UA connector tag probe to the template-sync skill. No CLI command surface, params, help text, or runtime code is touched, so no `HISTORY.rst` entry is expected here. The dispatch policy in Phase 3 matches `release_workflow.yml` (`continue-on-error`, `github_release`, `upload_wheel` all exist, and `upload_wheel=true` alone is enough to hit the `production` approval job), and Phase 2's "omit `test-scenarios` to run all scenarios" matches the empty default in `int_test.yml`.

**Blocking**

_None._

**Suggestions**

1. `.github/skills/on-demand-alpha-build/SKILL.md`, Phase 1 step 5 — the delegation claim is stronger than what the sync skill actually does. It says "the sync workflow already extracts the tag key-agnostically from the unredacted instance compile, **treats a missing value as blocking**, and reconciles `OPCUA_CONNECTOR_VERSION`". The only thing this PR adds to `sync-aio-bicep-templates/SKILL.md` is an `echo` + `jq ... // "NOT FOUND"` inside the *step 4 developer verification block* (the "recompile in a scratch dir and inspect" snippet). That block prints `NOT FOUND` and keeps going — there is no assertion, and `OPCUA_CONNECTOR_VERSION` appears nowhere else in the sync skill except that new comment. Either add an explicit "a `NOT FOUND` result is blocking; stop and reconcile `OPCUA_CONNECTOR_VERSION` before reporting success" rule to the sync skill (ideally in the section 4/5 constant-reconciliation checks rather than the verification block), or soften the wording in the alpha-build skill so it does not promise a guard that is not written down.

2. `.github/skills/sync-aio-bicep-templates/SKILL.md`, added `jq` line — `first(.. | objects | select(.connectors?.image?.tag) | .connectors.image.tag)` silently takes whichever match the recursive descent hits first. If a compile ever carries more than one `connectors.image.tag` (e.g. an alias variable plus the `$fxv#N` blob, or a second embedded values payload), an inconsistency is invisible and you can stamp the wrong tag into `OPCUA_CONNECTOR_VERSION`. Collecting and de-duping makes a conflict loud, e.g.:

   ```shell
   jq -r '[.. | objects | select(.connectors?.image?.tag) | .connectors.image.tag] | unique | if length == 1 then .[0] elif length == 0 then "NOT FOUND" else "AMBIGUOUS: \(.)" end' temp-instance.json
   ```

   This matches the tone of the rest of the skill, which elsewhere insists on structural parsing and on failing loudly rather than picking a candidate.

**Nits**

3. `.github/skills/sync-aio-bicep-templates/SKILL.md` — the five added lines are indented 2 spaces while the surrounding lines in the same fenced block use 3 (the block is nested under a numbered list item). It still renders, but it is inconsistent with every other line in the snippet.

Summary: well-scoped, self-consistent skill document; the safety gates (never merge, never approve `production`, never create a GitHub release for an alpha, immutable tests ref, run-correlation instead of "newest run") are the right ones and line up with the actual workflow definitions. The one thing worth fixing before merge is the mismatch between the "treats a missing value as blocking" claim and the non-blocking `NOT FOUND` echo that this PR actually adds.

.github/skills/on-demand-alpha-build/SKILL.md
ketkimnaik: It should now. I have configured teams workflow in it.

zzcodework:

Reviewing `changes since 9c1b9d8a` (head `797fc38`)

The increment adds `.github/workflows/teams_notify.yml` (a `workflow_dispatch`-only Teams notifier) and rewires two SKILL.md steps to dispatch it instead of only printing a ready-to-send message. The workflow itself is well-constructed for a first pass: `permissions: {}`, no `actions/checkout`, all inputs passed via `env:` rather than `${{ }}` interpolation inside `run:` (so no script-injection surface), an allowlist regex on `target-url`, length caps, and the secret confined to the final step. Findings below are on top of what Copilot already flagged (empty `cli-version` in the title, missing curl timeouts) — I am not repeating those.

**Blocking**

None.

**Suggestions**

- `.github/workflows/teams_notify.yml` line 107 — `TEAMS_WEBHOOK_URL` as a plain repository secret is readable by any workflow, on any branch, dispatched by anyone with write access. Since the SKILL.md change makes an automated agent a routine caller, the "dispatch only on a trusted CLI base branch" instruction is the only thing standing between the secret and an arbitrary branch, and that is a convention rather than an enforced control. Scoping the secret to a GitHub Environment (`environment: teams-notify` on the job, secret defined on that environment, optionally with a deployment-branch rule limiting it to the base branch) makes the restriction enforced by the platform. This repo already uses environments for `release_workflow.yml`'s `production` gate, so the pattern is established.
- `.github/workflows/teams_notify.yml` line 118 — `--retry 2` combined with `--fail-with-body` will re-issue the POST on transient HTTP failures (408/429/5xx) and on connection errors. If the endpoint accepted the message but the response was lost, this posts the same card to the channel two or three times. For a non-idempotent POST, consider dropping `--retry` (a failed notification run is visible and re-dispatchable) or keeping it only after confirming the connector tolerates duplicates.
- `.github/skills/on-demand-alpha-build/SKILL.md` — the workflow defines four notification types but the skill only ever dispatches `pr-ready` and `approval-required`. The stop/abort paths still just report and halt without dispatching `failure`, so that template is unreachable from the skill. Either wire the `failure` type into those paths (presumably why it was added) or note in the skill that `failure` and `test` are for manual dispatch only.

**Nits**

- `.github/workflows/teams_notify.yml` line 97 — when `details` is empty the card still carries a `TextBlock` with `text: ""` and `isVisible: false`. `text` is a required property in the Adaptive Card schema and empty values are inconsistently handled across renderers; conditionally omitting the block (the same `if ... then [] else ...` pattern already used for `actions`) is more robust than relying on `isVisible`.
- `.github/workflows/teams_notify.yml` line 117 — `--show-error` prints curl's diagnostic on connection failures, which can include the webhook host. GitHub only masks the exact secret string, so a substring such as the hostname is not redacted. Low severity, but suppressing stderr and emitting an explicit generic failure message avoids leaking any part of the endpoint into public job logs.
- `.github/workflows/teams_notify.yml` line 36 — no `timeout-minutes` on the job; with the curl timeouts Copilot suggested this is mostly redundant, but a small job-level timeout is a cheap backstop.

Summary: a sensible, injection-safe notification workflow and a coherent SKILL.md wiring. The main thing I would reconsider before merge is the secret's blast radius (repository secret reachable from any branch) — everything else is polish.

zzcodework:

Reviewing `changes since 797fc382` (head `08b2309`)

The increment deletes `.github/workflows/teams_notify.yml` entirely and rewrites the notification wording in `.github/skills/on-demand-alpha-build/SKILL.md` to rely on the native **Microsoft Teams for GitHub** app instead of a custom webhook dispatch. It also drops the internal Teams channel deep link and the two approver email addresses from the required-inputs list.

This is a strictly better direction than the previous iteration and it resolves everything I raised on `797fc38`: there is no longer a `TEAMS_WEBHOOK_URL` repository secret reachable from any branch, no non-idempotent `curl --retry` POST, no unreachable `failure`/`test` notification templates, and no Adaptive Card construction to get wrong. Removing the `teams.microsoft.com` channel URL and the `@microsoft.com` approver aliases from a public repo is also the right call on its own. I verified the removal is clean: `teams_notify.yml` is gone from the tree at `08b2309`, no other workflow or doc (including `.github/workflows/README.md`) still references it, and no leftover mention of the webhook, the channel URL, or the individual contacts remains in the skill.

**Blocking**

None.

**Suggestions**

- `.github/skills/on-demand-alpha-build/SKILL.md` line 173 — the approval-gate fallback message still prints `Approvers: <APPROVER_HANDLES>`, but the placeholder no longer has a source. The bullet that previously named the approval contacts was removed from the "Required inputs" section (which opens with "Ask for these at the start. **Do not infer or invent them**"), and the safety boundary now says to "resolve approvers from the protected GitHub environment or user input at runtime" — yet nothing in phases 1–3 instructs the agent to actually collect or look them up. As written the agent has to either leave the placeholder literal or invent handles, both of which are bad. Suggest adding an explicit input (e.g. "approver GitHub handles, or confirmation to read them from the `production` environment's reviewers") so the placeholder is resolvable without guessing.

- `.github/skills/on-demand-alpha-build/SKILL.md` lines 114, 166 — the fallback path says to print a message "for manual posting in the configured channel", but the skill no longer knows which channel that is (by design). Worth one sentence making the handoff explicit: ask the user which channel to post in, or state plainly that the fallback text is handed back for the user to route. Otherwise "the configured channel" reads like a value the skill is expected to have.

- `.github/skills/on-demand-alpha-build/SKILL.md` lines 43–46 — the new boundary asserts the channel is subscribed with the `pulls`, `reviews`, and `workflows` features, but nothing in the skill verifies that, and a missing subscription is silent (no error, just no card). Since the skill correctly refuses to claim delivery, consider stating the expected-but-unverified nature once up front — e.g. treat native notification as best-effort and always print the fallback text, rather than only "if the native notification is absent", which the agent cannot determine on its own.

**Nits**

- `.github/skills/on-demand-alpha-build/SKILL.md` line 166 — 123 characters; the file otherwise wraps at ~120. Same for line 137, also touched by this iteration.
- `.github/skills/on-demand-alpha-build/SKILL.md` lines 44–50, 113–115, 141–142 — continuation lines under `-` and numbered items use a 3-space indent where the surrounding pre-existing bullets use 2 (and 4 for ordered items). Renders fine in GitHub, just inconsistent with the neighbouring lines.

Summary: the right simplification — deleting the custom notifier removes the secret blast radius, the duplicate-POST risk, and the public-repo leak of internal contact details in one go. The only substantive loose end is `<APPROVER_HANDLES>`, which is now referenced but never sourced.

zzcodework:

Reviewing `changes since 08b23091` (head `fcc7884`)

The increment is one documentation commit plus a merge of `dev`:

- `e167ae4` — `.github/skills/sync-aio-bicep-templates/SKILL.md`: the OPC UA connector tag comment now reads `common.OPCUA_CONNECTOR_VERSION` instead of `constants.OPCUA_CONNECTOR_VERSION (bug 39164109)`.
- `fcc7884` — merge of `dev` into `automate-alpha-build`, no PR-owned file changes.

**Blocking**

None.

**Suggestions**

None.

**Nits**

None. The corrected pointer is accurate: `OPCUA_CONNECTOR_VERSION` is defined in `azext_edge/edge/providers/orchestration/common.py` (line 447 on `dev`) and does not exist in `azext_edge/constants.py`, so the previous `constants.` reference would have sent a reader to the wrong file. The neighbouring echo line in the same block already said `common.py`, so the two are now consistent, and dropping the internal bug id from a public-repo skill file is the right call.

*This is an automated review and may be incomplete — please treat the findings as input, not as gospel.*

zzcodework:

Reviewing `changes since fcc78848` (head `e6c4f08`)

The increment is a single documentation commit (`e6c4f08` "Refine alpha release workflow") touching only `.github/skills/on-demand-alpha-build/SKILL.md`. It drops the "native Microsoft Teams for GitHub app" narrative and replaces it with a generic "print a ready-to-send status message at each human gate" instruction, and it removes the `dev` default for the CLI PR base branch in favour of asking the requester explicitly.

**Blocking**

None.

**Suggestions**

None.

**Nits**

None. Two consistency checks that the change passes:

- The Teams removal is complete, not partial. Grepping the file at `e6c4f08` for `teams` / `notif` / `webhook` returns nothing, so the frontmatter `description`, the safety-boundaries bullet, phase 1 step 10, phase 2 step 4, phase 3 steps 4-5, and the "A successful run reports" list are now all telling the same story: emit the URL, print a shareable message, do not claim delivery. This also lines up with the earlier `08b23091` commit that removed the webhook notification workflow — the PR no longer adds any notification machinery, which retires the earlier inline review points about `cli-version` validation and missing `curl --connect-timeout`/`--max-time` on that workflow.
- Removing the base-branch default is the safer choice for this repo specifically, since `azure-iot-ops-cli-extension` alpha work can legitimately target `dev` or `preview`, and the rest of the skill (preflight remote-existence check, `--ref <CLI_BASE_BRANCH>` dispatch, `baseRefName` verification after merge) consistently threads the chosen value rather than re-deriving it.

*This is an automated review and may be incomplete — please treat the findings as input, not as gospel.*

HangyiWang (approved):

PR #929, opened 2026-09-02
fix: skip OPC UA connector template when opcua feature is disabled

ketkimnaik:
## Overview

- Fixes an issue where creating or upgrading an Azure IoT Operations instance with the OPC UA feature disabled could hang until the operation timed out.
- Updates the bundled Azure IoT Operations deployment templates to the 2609 release.
- Adds automated test coverage for the OPC UA disabled scenario.

### DIFF
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -27,10 +27,6 @@ scenarios:
     description: "Cloud tests"
     tox_env: python-rpsaas-int
 
-  - name: opcuaoff
-    description: "Create with OPC UA disabled"
-    create_args: "--feature opcua.mode=Disabled"
-
   - name: upgrade
     description: "Upgrade tests"
     tox_env: python-upgrade-int
@@ -42,8 +38,9 @@ scenarios:
     parallel: false
 
   - name: redeploy
-    description: "Redeployment tests"
+    description: "Redeployment tests (OPC UA disabled)"
     test_redeploy: true
+    create_args: "--feature opcua.mode=Disabled"
 
   - name: trustbundle
     description: "Workload identity tests"
--- a/.github/test-scenarios.yml
+++ b/.github/test-scenarios.yml
@@ -27,6 +27,10 @@ scenarios:
     description: "Cloud tests"
     tox_env: python-rpsaas-int
 
+  - name: opcuaoff
+    description: "Create with OPC UA disabled"
+    create_args: "--feature opcua.mode=Disabled"
+
   - name: upgrade
     description: "Upgrade tests"
     tox_env: python-upgrade-int
--- a/azext_edge/edge/providers/orchestration/resources/connector_templates.py
+++ b/azext_edge/edge/providers/orchestration/resources/connector_templates.py
@@ -22,7 +22,7 @@ import tarfile
 import tempfile
 from contextlib import nullcontext
 import hashlib
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, List, Optional, Tuple
 
 from azure.cli.core.azclierror import (
     InvalidArgumentValueError,
@@ -31,6 +31,7 @@ from azure.cli.core.azclierror import (
     ValidationError,
     CLIInternalError,
 )
+from azure.core.exceptions import HttpResponseError
 from knack.log import get_logger
 from rich.console import Console
 
@@ -289,6 +290,35 @@ class ConnectorTemplates(Queryable):
                 headers=headers or {},
             )
             return wait_for_terminal_state(poller=poller, logger=logger)
+
+    def check_default_opcua_template_needed(
+        self,
+        instance_name: str,
+        resource_group_name: str,
+    ) -> Tuple[bool, Optional[str]]:
+        """Return (needed, repair_name) for the default OPC UA connector template.
+
+        A prefix-matching template in a terminal ``Failed`` state is repaired in place by reusing
+        its name (``repair_name``); a ``Succeeded`` or transient template is left untouched so user
+        customizations and in-flight provisioning are preserved. If none exists, a new default is
+        needed. Shared by the upgrade backfill and the update re-enable path.
+        """
+        from ..common import OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX, PROVISIONING_STATE_FAILED
+
+        try:
+            existing_templates = self.list(instance_name=instance_name, resource_group_name=resource_group_name)
+            for template in existing_templates:
+                if not (template.get("name") or "").lower().startswith(OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX):
+                    continue
+                if (template.get("provisioningState") or "").lower() == PROVISIONING_STATE_FAILED.lower():
+                    logger.debug("Default OPC UA connector template exists but failed; will repair in place.")
+                    return True, template.get("name")
+                logger.debug("Default OPC UA connector template already exists.")
+                return False, None
+            return True, None
+        except HttpResponseError as e:
+            logger.debug(f"Error checking OPC UA connector template: {e}")
+            return False, None
 
     @staticmethod
     def default_opcua_template_name(instance_name: str) -> str:
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -273,11 +273,17 @@ class Instances(Queryable):
         if description:
             instance["properties"]["description"] = description
 
+        opcua_needs_backfill = False
         if features:
             desired_features = parse_feature_kvp_nargs(features, strict=True)
-            current_features: dict = instance["properties"].get("features", {})
+            current_features: dict = instance["properties"].get("features", {}) or {}
+            prior_opcua_mode = (current_features.get("opcua") or {}).get("mode")
             current_features.update(desired_features)
             instance["properties"]["features"] = current_features
+            new_opcua_mode = (current_features.get("opcua") or {}).get("mode")
+            # Enabling OPC UA from Disabled needs the default connector template the create path skips
+            # while disabled; without it the supervisor has nothing to reconcile.
+            opcua_needs_backfill = prior_opcua_mode == "Disabled" and new_opcua_mode not in (None, "Disabled")
 
         if adr_namespace_resource_id:
             instance["properties"]["adrNamespaceRef"] = {"resourceId": adr_namespace_resource_id}
@@ -296,7 +302,19 @@ class Instances(Queryable):
                 resource=instance,
                 **operation_kwargs,
             )
-            return wait_for_terminal_state(poller, **kwargs)
+            result = wait_for_terminal_state(poller, **kwargs)
+            if opcua_needs_backfill:
+                from .connector_templates import ConnectorTemplates
+                from ..common import OPCUA_CONNECTOR_VERSION
+
+                ConnectorTemplates(self.cmd).create_default_opcua_template(
+                    resource_group_name=resource_group_name,
+                    instance_name=name,
+                    connector_version=OPCUA_CONNECTOR_VERSION,
+                    headers=headers,
+                    no_status=no_status,
+                )
+            return result
 
     def remove_mi_user_assigned(
         self,
--- a/azext_edge/edge/providers/orchestration/resources/instances.py
+++ b/azext_edge/edge/providers/orchestration/resources/instances.py
@@ -273,17 +273,16 @@ class Instances(Queryable):
         if description:
             instance["properties"]["description"] = description
 
-        opcua_needs_backfill = False
+        opcua_backfill_requested = False
         if features:
             desired_features = parse_feature_kvp_nargs(features, strict=True)
             current_features: dict = instance["properties"].get("features", {}) or {}
-            prior_opcua_mode = (current_features.get("opcua") or {}).get("mode")
             current_features.update(desired_features)
             instance["properties"]["features"] = current_features
-            new_opcua_mode = (current_features.get("opcua") or {}).get("mode")
-            # Enabling OPC UA from Disabled needs the default connector template the create path skips
-            # while disabled; without it the supervisor has nothing to reconcile.
-            opcua_needs_backfill = prior_opcua_mode == "Disabled" and new_opcua_mode not in (None, "Disabled")
+            requested_opcua_mode = (desired_features.get("opcua") or {}).get("mode")
+            # Enabling OPC UA needs the default connector template the create path skips while
+            # disabled; running on any enable request also lets a repeat repair a partial attempt.
+            opcua_backfill_requested = requested_opcua_mode not in (None, "Disabled")
 
         if adr_namespace_resource_id:
             instance["properties"]["adrNamespaceRef"] = {"resourceId": adr_namespace_resource_id}
@@ -303,17 +302,23 @@ class Instances(Queryable):
                 **operation_kwargs,
             )
             result = wait_for_terminal_state(poller, **kwargs)
-            if opcua_needs_backfill:
+            if opcua_backfill_requested:
                 from .connector_templates import ConnectorTemplates
                 from ..common import OPCUA_CONNECTOR_VERSION
 
-                ConnectorTemplates(self.cmd).create_default_opcua_template(
-                    resource_group_name=resource_group_name,
-                    instance_name=name,
-                    connector_version=OPCUA_CONNECTOR_VERSION,
-                    headers=headers,
-                    no_status=no_status,
+                connector_templates = ConnectorTemplates(self.cmd)
+                needed, repair_name = connector_templates.check_default_opcua_template_needed(
+                    instance_name=name, resource_group_name=resource_group_name
                 )
+                if needed:
+                    connector_templates.create_default_opcua_template(
+                        resource_group_name=resource_group_name,
+                        instance_name=name,
+                        connector_version=OPCUA_CONNECTOR_VERSION,
+                        template_name=repair_name,
+                        headers=headers,
+                        no_status=no_status,
+                    )
             return result
 
     def remove_mi_user_assigned(
--- a/azext_edge/edge/providers/orchestration/targets.py
+++ b/azext_edge/edge/providers/orchestration/targets.py
@@ -251,24 +251,19 @@ class InitTargets:
     ) -> Tuple[dict, dict]:
         if not cl_extension_ids:
             cl_extension_ids = []
-        param_to_target = {
-            "clusterName": self.cluster_name,
-            "clusterNamespace": self.cluster_namespace,
-            "clusterLocation": self.location,
-            "customLocationName": self.custom_location_name,
-            "clExtensionIds": cl_extension_ids,
-            "schemaRegistryId": self.schema_registry_resource_id,
-            "adrNamespaceId": self.adr_namespace_resource_id,
-            "defaultDataflowInstanceCount": self.dataflow_profile_instances,
-            "brokerConfig": self.broker_config,
-            "trustConfig": self.trust_config,
-        }
-        # Drive the template's OPC UA connector condition; the instance features are injected
-        # directly (not via the features parameter), so gate on the parameter the condition reads.
-        if self.instance_features and self.instance_features.get("opcua", {}).get("mode") == "Disabled":
-            param_to_target["disableOpcUaFeature"] = True
         template, parameters = self._handle_apply_targets(
-            param_to_target=param_to_target,
+            param_to_target={
+                "clusterName": self.cluster_name,
+                "clusterNamespace": self.cluster_namespace,
+                "clusterLocation": self.location,
+                "customLocationName": self.custom_location_name,
+                "clExtensionIds": cl_extension_ids,
+                "schemaRegistryId": self.schema_registry_resource_id,
+                "adrNamespaceId": self.adr_namespace_resource_id,
+                "defaultDataflowInstanceCount": self.dataflow_profile_instances,
+                "brokerConfig": self.broker_config,
+                "trustConfig": self.trust_config,
+            },
             template_blueprint=TEMPLATE_BLUEPRINT_INSTANCE,
         )
 
@@ -317,6 +312,10 @@ class InitTargets:
             )
 
         resources: Dict[str, Dict[str, dict]] = template.content.get("resources", {})
+        # OPC UA disabled: drop the connector template so it is never deployed; no supervisor
+        # reconciles it, so its provisioning would otherwise never reach a terminal state.
+        if self.instance_features and self.instance_features.get("opcua", {}).get("mode") == "Disabled":
+            resources.pop("opcUaConnectorTemplate", None)
         if phase == InstancePhase.EXT:
             del_if_not_in(resources, PHASE_KEY_MAP[InstancePhase.EXT])
             return template.content, parameters
--- a/azext_edge/edge/providers/orchestration/targets.py
+++ b/azext_edge/edge/providers/orchestration/targets.py
@@ -251,19 +251,24 @@ class InitTargets:
     ) -> Tuple[dict, dict]:
         if not cl_extension_ids:
             cl_extension_ids = []
+        param_to_target = {
+            "clusterName": self.cluster_name,
+            "clusterNamespace": self.cluster_namespace,
+            "clusterLocation": self.location,
+            "customLocationName": self.custom_location_name,
+            "clExtensionIds": cl_extension_ids,
+            "schemaRegistryId": self.schema_registry_resource_id,
+            "adrNamespaceId": self.adr_namespace_resource_id,
+            "defaultDataflowInstanceCount": self.dataflow_profile_instances,
+            "brokerConfig": self.broker_config,
+            "trustConfig": self.trust_config,
+        }
+        # Drive the template's OPC UA connector condition; the instance features are injected
+        # directly (not via the features parameter), so gate on the parameter the condition reads.
+        if self.instance_features and self.instance_features.get("opcua", {}).get("mode") == "Disabled":
+            param_to_target["disableOpcUaFeature"] = True
         template, parameters = self._handle_apply_targets(
-            param_to_target={
-                "clusterName": self.cluster_name,
-                "clusterNamespace": self.cluster_namespace,
-                "clusterLocation": self.location,
-                "customLocationName": self.custom_location_name,
-                "clExtensionIds": cl_extension_ids,
-                "schemaRegistryId": self.schema_registry_resource_id,
-                "adrNamespaceId": self.adr_namespace_resource_id,
-                "defaultDataflowInstanceCount": self.dataflow_profile_instances,
-                "brokerConfig": self.broker_config,
-                "trustConfig": self.trust_config,
-            },
+            param_to_target=param_to_target,
             template_blueprint=TEMPLATE_BLUEPRINT_INSTANCE,
         )
 
--- a/azext_edge/edge/providers/orchestration/template.py
+++ b/azext_edge/edge/providers/orchestration/template.py
@@ -49,12 +49,12 @@ class TemplateBlueprint(NamedTuple):
 
 
 TEMPLATE_BLUEPRINT_ENABLEMENT = TemplateBlueprint(
-    commit_id="51dff082e4e1fbdb156c650f79917774cb4c0062",
+    commit_id="f9e05bed5c435135403c5a32264afebf14a409ec",
     content={
         "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
         "languageVersion": "2.0",
         "contentVersion": "1.0.0.0",
-        "metadata": {"_generator": {"name": "bicep", "version": "0.46.1.21595", "templateHash": "6829492317291626106"}},
+        "metadata": {"_generator": {"name": "bicep", "version": "0.46.1.21595", "templateHash": "8485265132777030182"}},
         "definitions": {
             "_1.AdvancedConfig": {
                 "type": "object",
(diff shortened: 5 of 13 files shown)


azext_edge/edge/providers/orchestration/upgrade2.py
```diff
@@ -209,6 +209,12 @@ def _check_opcua_connector_template_needed(self) -> Tuple[bool, Optional[str]]:
         in-flight provisioning.
         """
         self._opcua_template_name_to_repair = None
+        opcua_mode = (
+            (self.instance_record.get("properties") or {}).get("features", {}).get("opcua", {}).get("mode")
+        )
```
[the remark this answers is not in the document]
ketkimnaik: Fixed in a78ab33: features is now coalesced to {} so a null features value no longer breaks upgrade.

azext_edge/tests/edge/orchestration/test_targets_unit.py
```diff
@@ -494,6 +494,63 @@ def test_init_targets_opcua_mode(target_scenario: dict):
...
+        instance_name=generate_random_string(),
+        instance_features=instance_features,
+    )
+    extension_ids = [generate_random_string()]
+
+    for phase in (None, InstancePhase.RESOURCES):
+        template, _ = targets.get_ops_instance_template(extension_ids, phase=phase)
+        connector_template = template["resources"].get("opcUaConnectorTemplate")
+        not_deployed = connector_template is None
+        gated = bool(connector_template and connector_template.get("condition"))
```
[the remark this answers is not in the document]
ketkimnaik: Fixed in a78ab33: the test now asserts the disableOpcUaFeature deployment parameter is set to drive the gate.

HangyiWang (dismissed):

zzcodework:

Reviewing `full diff` (head `b9b44e0a`)

The core fix looks right: `opcUaConnectorTemplate` now carries `"condition": "[not(equals(variables('opcUaFeatureMode'), 'Disabled'))]"`, and because `InitTargets` injects instance features directly into the resource body rather than through the `features` template parameter, driving the condition off a dedicated `disableOpcUaFeature` parameter (rather than off `parameters('features')`) is the correct call — the comment in `targets.py` explaining that is appreciated. The upgrade-side guard and both unit tests line up with the same failure mode.

Findings below.

**Blocking**

1. `azext_edge/edge/providers/orchestration/template.py` — `TEMPLATE_BLUEPRINT_INSTANCE.variables.TRAINS` changed from `{"iotOperations": "stable"}` to `{"iotOperations": "integration"}` (alongside `VERSIONS.iotOperations` `1.4.73` → `1.4.105`). This changes the release train the AIO extension is installed from for **every** `az iot ops create`/`init`, not just the OPC UA scenario. Two things make it worth calling out explicitly:
   - `TEMPLATE_BLUEPRINT_ENABLEMENT` in the same commit kept `stable` for certManager/secretStore, so the two blueprints now disagree on train.
   - The guardrail that would have caught this was updated in the same PR: `azext_edge/tests/edge/orchestration/test_template_unit.py` `EXTENSION_CONFIGS["instance"]` now asserts `(EXTENSION_TYPE_OPS, "1.4.105", "integration")`.

   Given `constants.py` moves to `VERSION = "2.10.0a1"` this may well be deliberate for an alpha build, but it must not reach a stable 2.10.0 tag. Please confirm the intent and, if it is alpha-only, note how/when it flips back to `stable` — otherwise this is a much larger behaviour change than the PR title suggests.

**Suggestions**

2. `azext_edge/edge/providers/orchestration/template.py` — the new `effectiveFeatures` variable and the `aioInstance` change from `"features": "[parameters('features')]"` to `"[variables('effectiveFeatures')]"` appear to be dead code on the CLI path. `InitTargets.get_ops_instance_template` replaces the instance body wholesale with `instance["properties"] = get_default_instance_config(description=..., features=self.instance_features)` (`targets.py`, and `get_default_instance_config` at the bottom of the same file returns a literal dict with `"features": features`), so the templated expression is discarded before deployment and `parameters('features')` is never supplied. Same for the `parameters('features')` fallback inside `opcUaFeature`. If these exist to keep the bundled template correct for consumers who deploy it directly with a `features` parameter, that is a reasonable reason to keep them — worth a one-line comment saying so, since otherwise the next reader will assume `effectiveFeatures` is what actually stamps the instance record.

3. `azext_edge/edge/providers/orchestration/upgrade2.py` — `_check_opcua_connector_template_needed` now returns `(False, None)` on `opcua.mode == "Disabled"` **before** listing existing connector templates. That is correct for the never-enabled case, but it also means an instance that was created with OPC UA enabled, has a real `akriConnectorTemplates` resource, and is *later* switched to `Disabled` will no longer be inspected for an in-flight/stuck template on upgrade. If that transition is possible, a short comment stating it is intentionally out of scope would help; if it is not, the early return is fine as-is.

**Nits**

4. `azext_edge/tests/edge/orchestration/test_targets_unit.py` — the `not_deployed or gated` assertion passes if `opcUaConnectorTemplate` is absent for *any* reason, so it would not catch a future refactor that drops the resource key entirely. The companion `test_init_targets_opcua_enabled_keeps_connector_template` mostly covers that gap, so this is minor; asserting `gated` directly (the resource is present **and** conditioned **and** the parameter is `True`) would be a tighter regression test for the exact bug being fixed.

Everything else reads cleanly: `OPCUA_CONNECTOR_VERSION` in `common.py` (`1.4.13`) matches the template variable default, which the pre-existing `test_opcua_connector_version_matches_template_tag` will keep honest; the new `opcuaoff` entry in `.github/test-scenarios.yml` omits `tox_env`, which is valid per the header comment and the existing `redeploy` scenario; and no secrets, injection, or unsafe subprocess/HTTP surfaces are introduced.

*This is an automated review and may be incomplete — please treat the findings as input, not as gospel.*

cheatsheet1999 (dismissed):
The PR looks good to me. The conflict is from my recent unit test changes in test_upgrade2_unit, so pulling the latest upstream should sort it out.

cheatsheet1999 (dismissed):

azext_edge/tests/edge/orchestration/test_targets_unit.py
```diff
@@ -494,6 +494,69 @@ def test_init_targets_opcua_mode(target_scenario: dict):
...
+    targets = InitTargets(
+        cluster_name=generate_random_string(),
+        resource_group_name=generate_random_string(),
+        schema_registry_resource_id=get_schema_registry_id(),
+        adr_namespace_resource_id=get_ns_resource_id(),
+        instance_name=generate_random_string(),
+        instance_features=instance_features,
+    )
+    template, _ = targets.get_ops_instance_template([generate_random_string()], phase=InstancePhase.RESOURCES)
+    assert "opcUaConnectorTemplate" in template["resources"]
```
digimaun: This assertion holds regardless of the feature value. The ARM condition leaves `opcUaConnectorTemplate` present for Disabled, Stable and the default alike.

Forcing the guard in `targets.py` to always take the disabled path leaves all three new cases passing. The seven failures come from pre-existing `test_init_targets` cases, so the regression is caught, but not by these.

If the resource is removed in Python as suggested on `targets.py`, these get simpler. Disabled should omit `opcUaConnectorTemplate` from the full and `RESOURCES` templates while Stable and the default keep it, and `EXT` and `INSTANCE` omit it in every mode because their phase filters drop child resources. Those fail if the guard goes unconditional or disappears.

azext_edge/edge/providers/orchestration/targets.py
```diff
@@ -251,19 +251,24 @@ def get_ops_instance_template(
...
+            "clExtensionIds": cl_extension_ids,
+            "schemaRegistryId": self.schema_registry_resource_id,
+            "adrNamespaceId": self.adr_namespace_resource_id,
+            "defaultDataflowInstanceCount": self.dataflow_profile_instances,
+            "brokerConfig": self.broker_config,
+            "trustConfig": self.trust_config,
+        }
+        # Drive the template's OPC UA connector condition; the instance features are injected
+        # directly (not via the features parameter), so gate on the parameter the condition reads.
+        if self.instance_features and self.instance_features.get("opcua", {}).get("mode") == "Disabled":
```
digimaun: Python already owns the normalized feature state and writes the instance properties directly at :286. Encoding that state into a new ARM parameter and reconstructing it through template variables works against that separation.

Removing the resource from the copied template is smaller and already the pattern here. `_handle_apply_targets` returns `template_blueprint.copy()`, :284 fetches this resource and :299 rewrites its name, and :321 already deletes resources through `del_if_not_in`. So alongside that:

```python
resources: Dict[str, Dict[str, dict]] = template.content.get("resources", {})
if self.instance_features and self.instance_features.get("opcua", {}).get("mode") == "Disabled":
    resources.pop("opcUaConnectorTemplate", None)
```

The rename at :299 runs first and the pop then removes that same dict, and if the resource is ever absent `get_resource_by_key` returns a detached fallback so the pop stays a no-op. `EXT` and `INSTANCE` already drop child resources, so this only changes the full and `RESOURCES` renders.

That removes `disableOpcUaFeature` and all three new variables. On our path `effectiveFeatures` is overwritten by the properties assignment at :286, `opcUaFeature` has no `features` parameter to read since we never pass one, and `opcUaFeatureMode` exists only to turn the boolean back into a condition. Running against the 2608 template with no template change at all, those three files pass.

zzcodework:

Reviewing `changes since b9b44e0a` (head `402ecbf`)

The increment is formatting only, plus a merge of `dev`:

- `9162a26` — merge of `dev` into `opcua-disable-tests`; no PR-owned file changes.
- `402ecbf` — `azext_edge/tests/edge/orchestration/test_upgrade2_unit.py`: two blank lines added after `test_check_opcua_connector_template_needed_disabled()` so `build_ext_upgrade_state` is separated per E302.

**Blocking**

None.

**Suggestions**

None.

**Nits**

None — this is exactly the flake8 fix the merged test needed, and it changes no test behaviour.

Note for readers: my earlier findings on head `b9b44e0a` are unaffected by this iteration, and the substantive design discussion now on the PR (removing `opcUaConnectorTemplate` from the copied template in `targets.py` rather than gating it with a new ARM parameter, and tightening the `test_targets_unit.py` assertions so they actually fail if the guard goes unconditional) is a stronger version of the same concern — that thread, not this comment, is the one to act on.

*This is an automated review and may be incomplete — please treat the findings as input, not as gospel.*

azext_edge/edge/providers/orchestration/template.py
```diff
@@ -1691,6 +1693,7 @@ def copy(self) -> "TemplateBlueprint":
                 "dependsOn": ["aioInstance", "customLocation"],
             },
             "opcUaConnectorTemplate": {
+                "condition": "[not(equals(variables('opcUaFeatureMode'), 'Disabled'))]",
```
digimaun, line 1696: Creating with opcua disabled now produces an instance with no default connector template. The re-enable path does not create one. `Instances.update` only writes the instance resource, and the two places that create the default template are the create path and the upgrade backfill. The error text in `adr/helpers.py` and `certs.py` points users at `az iot ops update --feature opcua.mode=Stable`, so following it leaves OPC UA enabled without the template the supervisor needs until an upgrade backfills it. Worth handling the disabled to enabled transition, since this PR is what makes that starting state reachable.

.github/test-scenarios.yml
```diff
@@ -27,6 +27,10 @@ scenarios:
     description: "Cloud tests"
     tox_env: python-rpsaas-int
 
+  - name: opcuaoff
```
digimaun: Live coverage for a disabled deployment is useful, but I do not think it needs another full cluster lifecycle. Each scenario provisions its own runner, k3s cluster, Arc connection, storage account, schema registry and ADR namespace, and the seven scenario jobs in the last successful nightly took 46 to 116 minutes each. The suite is already long enough that it runs nightly rather than per PR. Could this use `redeploy` instead, with both of its creates running under `opcua.mode=Disabled`?

The scenario also needs to prove the requested state. `python-init-int` does not assert the OPC UA feature or connector templates, so this job still passes if `--feature` is ignored and OPC UA stays enabled. Asserting that `features.opcua.mode` is Disabled and that no default connector template exists would make it a real check.

ketkimnaik: Done — dropped the standalone scenario and folded it into redeploy with --feature opcua.mode=Disabled, and the init int-test now asserts features.opcua.mode == Disabled and that no default connector template exists.

azext_edge/edge/providers/orchestration/template.py
ketkimnaik: It noe backfills the default OPC UA connector template on the Disabled → enabled transition, so following the "update --feature opcua.mode=Stable" guidance no longer leaves the supervisor without a template.

azext_edge/edge/providers/orchestration/targets.py
ketkimnaik: Done — removed the ARM parameter/variable approach and now pop opcUaConnectorTemplate from the copied template when opcua is disabled, matching the existing del_if_not_in pattern.

azext_edge/tests/edge/orchestration/test_targets_unit.py
ketkimnaik: Done — switched to the Python pop and the tests now assert opcUaConnectorTemplate is absent for Disabled and present for Stable/default across the full and RESOURCES phases.

azext_edge/edge/providers/orchestration/resources/instances.py
```diff
@@ -273,11 +273,17 @@ def update(
...
             desired_features = parse_feature_kvp_nargs(features, strict=True)
-            current_features: dict = instance["properties"].get("features", {})
+            current_features: dict = instance["properties"].get("features", {}) or {}
+            prior_opcua_mode = (current_features.get("opcua") or {}).get("mode")
             current_features.update(desired_features)
             instance["properties"]["features"] = current_features
+            new_opcua_mode = (current_features.get("opcua") or {}).get("mode")
+            # Enabling OPC UA from Disabled needs the default connector template the create path skips
+            # while disabled; without it the supervisor has nothing to reconcile.
+            opcua_needs_backfill = prior_opcua_mode == "Disabled" and new_opcua_mode not in (None, "Disabled")
```
digimaun: The backfill PUTs a freshly built default payload without checking what is already there. For an instance created by this CLI the create path already names the template with `default_opcua_template_name`, so this overwrites it and resets anything set through `az iot ops connector template update`, including replicas, log level and image pull settings. For an instance deployed from the product Bicep the existing template keeps the `uniqueString` suffix, so this adds a second prefix match instead. Either way it bypasses the succeeded, transient and failed handling `_check_opcua_connector_template_needed` already performs.

Extracting that prefix and provisioning state decision into `ConnectorTemplates` and calling it from both paths would cover this. Running it whenever the requested mode is enabled, rather than only when the prior mode was Disabled, also makes a repeat of the same command repair a partial attempt.

zzcodework:

Reviewing `changes since 402ecbf` (head `0d43b08`)

This iteration switches the fix from "gate `opcUaConnectorTemplate` with an ARM condition" to "pop it out of the copied template in Python", tightens the two `test_targets_unit.py` assertions, and adds a new `az iot ops update` backfill that creates the default OPC UA connector template when the feature is re-enabled. The template-side change is good — `_handle_apply_targets` returns `template_blueprint.copy()` (deepcopy in `TemplateBlueprint.copy`), so `resources.pop("opcUaConnectorTemplate", None)` mutates only the per-call copy, and the assertions now genuinely fail if the resource reappears in either the full or `RESOURCES` phase. The new backfill is where my concerns are.

**Blocking**

1. `azext_edge/edge/providers/orchestration/resources/instances.py` (`update`, the `if opcua_needs_backfill:` block) — the backfill PUTs unconditionally, with no check for an existing template, and the name it generates does **not** match the name the ARM path generates. `ConnectorTemplates.default_opcua_template_name` derives the suffix from `hashlib.sha256(instance_name)[:4]`, while the blueprint's `opcUaConnectorTemplateName` uses `substring(uniqueString(resourceId('Microsoft.IoTOperations/instances', ...)), 0, 4)` (`template.py`). Those are different values for the same instance. Consider the path this PR now makes reachable: instance created with OPC UA **enabled** (ARM creates `azureiotoperationsconnectorforopcua-<uniqueString>`) → `az iot ops update --feature opcua.mode=Disabled` → `az iot ops update --feature opcua.mode=Stable`. `prior_opcua_mode == "Disabled"` is true, so the backfill fires and creates `azureiotoperationsconnectorforopcua-<sha256>` — a **second** prefix-matching template alongside the existing one. Since the supervisor adopts by prefix (per the docstring on `default_opcua_template_name`), two adopted templates is not a benign state. `upgrade2._check_opcua_connector_template_needed` already solves exactly this: it lists templates, skips when a prefix-matching one exists in a good/transient state, and repairs a `Failed` one in place by reusing its name. The update path should reuse that logic rather than PUT blind.

**Suggestions**

2. `instances.py` — no error handling around the backfill. `upgrade2.apply_upgrades` wraps the identical `create_default_opcua_template` call in `try/except HttpResponseError` with a correlation-id log before re-raising; here any failure propagates raw out of `az iot ops update` *after* `begin_create_or_update` has already committed the feature flip, and the `result` that was just computed is discarded. The user then sees a failed command for an update that actually succeeded, with no instance record printed. At minimum log a warning and still `return result`, or mirror the upgrade2 error path so the correlation id is emitted.

3. `instances.py` — no minimum-version gate. `upgrade2` only creates the template when the target version is at least `MIN_INSTANCE_VERSION_FOR_OPCUA_CONNECTOR_TEMPLATE` (`1.4.73`, `common.py`). The backfill has no equivalent check, so re-enabling OPC UA on an instance older than that will attempt a PUT the service may not support. If the reasoning is that any instance reaching this code path is necessarily new enough, a one-line comment saying so would prevent the next reader from assuming the gap is an oversight.

4. `.github/test-scenarios.yml` — the dedicated `opcuaoff` scenario is deleted and `--feature opcua.mode=Disabled` is folded into `redeploy`. Two side effects worth confirming as intentional: there is no longer a plain (non-redeploy) create-with-OPC-UA-disabled job, and a failure in the `redeploy` job now conflates redeployment regressions with OPC-UA-disabled regressions — which is unfortunate for the one scenario this PR exists to protect. If this was to save CI capacity, saying so in the PR description would be enough. (The wiring itself is fine: `process_additional_args` turns `--feature opcua.mode=Disabled` into `{"feature": "opcua.mode=Disabled"}`, which is what the new `feature` kwarg on `assert_aio_instance` reads, so the new integration assertion will actually run.)

5. `azext_edge/tests/edge/orchestration/resources/test_instances_unit.py` — the backfill branch only asserts that *some* PUT hit `/akriConnectorTemplates/`. Given finding 1, asserting the resulting template **name** (and that the call count is exactly 3) would be the assertion that catches a duplicate-name regression.

**Nits**

6. `azext_edge/edge/providers/orchestration/template.py` — with `disableOpcUaFeature` no longer passed from `targets.py`, the parameter is now permanently `False`, which makes `opcUaFeatureMode`, the `condition` on `opcUaConnectorTemplate`, and the `effectiveFeatures` indirection all dead on the CLI path (`parameters('features')` is never supplied either — the instance body is replaced wholesale by `get_default_instance_config`). This is the same dead-code point I raised on `b9b44e0a`, but it is now unambiguous rather than arguable. Since `template.py` mirrors the product Bicep, hand-added parameters/variables that nothing drives will either be lost or conflict at the next regeneration — I would either revert the `template.py` additions from this PR and keep the Python-side pop as the sole mechanism, or keep the ARM condition and drop the pop, but not leave both half-wired.

Otherwise the increment reads cleanly: `re` is already imported in `test_instances_unit.py`, the local imports inside `update` match the module's existing style, `OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX` is correctly re-exported for the integration assertion, and no secrets, injection, or unsafe subprocess/HTTP surfaces are introduced.

*This is an automated review and may be incomplete — please treat the findings as input, not as gospel.*

azext_edge/edge/providers/orchestration/resources/instances.py
ketkimnaik: Done — the default OPC UA template check now lives in one shared place used by both upgrade and update, so re-enabling repairs a failed template in place and never overwrites a customized one or creates a duplicate.

zzcodework:

Reviewing `changes since 0d43b08` (head `b587d8f`)

The increment factors the "is the default OPC UA connector template needed?" decision out of `Upgrade2._check_opcua_connector_template_needed` into a shared `ConnectorTemplates.check_default_opcua_template_needed`, and makes `Instances.update` call that check before PUTting. That directly addresses the "backfill PUTs a default payload without checking what is already there" concern raised earlier on this PR — an existing `Succeeded` (or transient) prefix-matching template is now left alone, and a `Failed` one is repaired in place via `template_name=repair_name`. I verified the helper is behaviour-preserving for the upgrade path: `ConnectorTemplates.list` flattens `properties.provisioningState` into a top-level `provisioningState` key in `_extract_template_summary`, so the `template.get("provisioningState")` lookup inside the helper is correct against what `list()` actually returns, and `create_default_opcua_template` already accepts an optional `template_name`. `HttpResponseError` and `Tuple` remain used in `upgrade2.py` after the import trim, so no dead imports.

**Blocking**

None.

**Suggestions**

1. `azext_edge/edge/providers/orchestration/resources/connector_templates.py` — `check_default_opcua_template_needed` swallows `HttpResponseError` and returns `(False, None)` with only a `logger.debug`. That was defensible when the caller was `upgrade`, which prints its own resource status table, but `Instances.update` now inherits it: if the `akriConnectorTemplates` list call fails transiently (throttling, a propagating RBAC assignment), `az iot ops update --feature opcua.mode=Stable` reports success, creates no template, and says nothing the user can see at default verbosity. Before this change that same command unconditionally created the template, so this is a silent-failure regression on the update path specifically. A `logger.warning` (or returning the error to the caller so `update` can decide) would keep the "don't clobber" fix without making the failure invisible.

2. `azext_edge/tests/edge/orchestration/resources/test_instances_unit.py` — the update-path test only mocks the list as `{"value": []}`, i.e. only the "nothing exists, create a new one" branch. The two branches that motivated this refactor are untested from `Instances.update`: (a) a `Succeeded` prefix-matching template exists → assert **no** PUT is issued, and (b) a `Failed` one exists → assert the PUT reuses that exact ARM name. `test_upgrade2_unit.py` covers the helper well through the upgrade scenarios (`opcua_connector_template_exists` / `_name` / `_provisioning_state`), but not the new `needed` / `repair_name` wiring in `Instances.update` — a future edit there could drop the `if needed:` guard and every test would still pass.

**Nits**

3. `azext_edge/edge/providers/orchestration/resources/connector_templates.py` — the helper returns on the first prefix-matching template it sees, so with more than one adopt-prefix template the ARM list ordering decides whether a `Failed` one is repaired. This is pre-existing behaviour moved verbatim, so it is fine to leave; worth a short comment noting the first match wins if you expect multiples to be possible.

Summary: the refactor is correct and is the right shape — one shared check used by both the upgrade backfill and the update re-enable path. The two points worth acting on are the invisible skip when the list call fails on the `update` path, and update-path test coverage for the "already exists" / "repair in place" branches.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
```diff
@@ -290,6 +291,35 @@ def create_default_opcua_template(
...
+            for template in existing_templates:
+                if not (template.get("name") or "").lower().startswith(OPCUA_CONNECTOR_TEMPLATE_NAME_PREFIX):
+                    continue
+                if (template.get("provisioningState") or "").lower() == PROVISIONING_STATE_FAILED.lower():
+                    logger.debug("Default OPC UA connector template exists but failed; will repair in place.")
+                    return True, template.get("name")
+                logger.debug("Default OPC UA connector template already exists.")
+                return False, None
+            return True, None
+        except HttpResponseError as e:
```
digimaun: `check_default_opcua_template_needed` turns every `HttpResponseError` into `(False, None)`. In the update path that runs after the instance PUT has committed, so a transient list failure leaves `opcua.mode=Stable` with no connector template and the command still reports success. That is the state this change exists to prevent.

The same posture is fine in upgrade, which re-evaluates on every run and does not mutate the instance first. Update should surface the failure instead.

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
```diff
@@ -631,19 +631,49 @@ def test_instance_update_opcua_mode(
...
+    # Enabling OPC UA triggers a backfill: list existing templates, create the default if absent.
+    new_opcua_mode = features_scenario["expected"].get("opcua", {}).get("mode")
+    expects_backfill = new_opcua_mode not in (None, "Disabled")
+    if expects_backfill:
+        base_url = instance_endpoint.split("?")[0]
+        list_re = re.compile(re.escape(base_url) + r"/akriConnectorTemplates(\?|$)")
+        mocked_responses.add(
+            method=responses.GET,
+            url=list_re,
+            json={"value": []},
```
digimaun, line 645: The template list is mocked as empty here, so only the create branch is covered. Changing `if needed:` in `Instances.update` to unconditional creation leaves all 238 cases in this file passing, which means the existence check added for the overwrite and duplicate cases is not pinned.

Worth adding update path cases for a Succeeded template left untouched, a Bicep named template not duplicated, a Failed template repaired under its existing name, and a repeat enable request being idempotent.

azext_edge/edge/providers/orchestration/resources/connector_templates.py
ketkimnaik: Done — the shared check now lets the error propagate; upgrade keeps its graceful catch while update surfaces the failure so it can't report success with no template.

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
ketkimnaik: Done — added update-path cases: succeeded left untouched, Bicep-named not duplicated, failed repaired under its existing name, and repeat enable idempotent.

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
```diff
@@ -631,19 +631,168 @@ def test_instance_update_opcua_mode(
...
+        ([], True, None),
+        # Succeeded default template -> left untouched (no overwrite of customizations).
+        (
+            [{"name": "azureiotoperationsconnectorforopcua-abcd", "properties": {"provisioningState": "Succeeded"}}],
+            False,
+            None,
+        ),
+        # Bicep-named template (prefix + uniqueString suffix), Succeeded -> not duplicated.
+        (
+            [{"name": "azureiotoperationsconnectorforopcua-e5c0", "properties": {"provisioningState": "Succeeded"}}],
```
digimaun: This case distinguishes itself from the one above only by holding a suffix that differs from the one `default_opcua_template_name` derives for this instance. That derivation is a sha256 of the instance name, and `generate_random_string` is seeded, so the name here is stable for a given test ordering rather than arbitrary. If an ordering change ever lands the two on the same four characters, the case still passes but stops covering the Bicep named path it is there for.

Deriving the name in the test and perturbing it, or asserting up front that it differs from the derived default, would keep the intent pinned.

zzcodework:

Reviewing `changes since fdd649f0` (head `daf30bd`). The comparison range also replays the `dev` merge (GitHub Actions SHA pinning, `dependabot.yml` cooldown) — that is base-branch churn and is not reviewed here. The PR's own increment is commits `1e9edcf` (surface update backfill list errors), `606678e` (avoid nested status spinner) and the tests that came with them.

**Blocking**

None. Both of the previous round's inline findings are genuinely addressed:

- `azext_edge/edge/providers/orchestration/resources/connector_templates.py` — `check_default_opcua_template_needed` no longer collapses every `HttpResponseError` into `(False, None)`. The docstring now states the contract explicitly ("List failures propagate; the caller decides whether to surface them"), `upgrade2._check_opcua_connector_template_needed` keeps the tolerant catch with a comment explaining why (re-evaluated every run, instance not mutated), and `Instances.update` lets it raise after the instance PUT has committed. That is the right split — the update path can no longer report success with `opcua.mode=Stable` and no connector template.
- `azext_edge/tests/edge/orchestration/resources/test_instances_unit.py` — the new `test_instance_update_opcua_backfill` parametrization pins the branches that were previously unpinned: Succeeded default left untouched, Bicep-named (`...opcua-e5c0`) not duplicated, `Failed` repaired under its existing name, custom-only template still creating the default. Flipping `if needed:` to an unconditional create would now fail three of the five cases.

Also correct: `no_status=True` in `Instances.update`. The backfill runs inside `with status_context`, which is a live `console.status(...)` spinner whenever the caller did not pass `no_status`, so forwarding `no_status=no_status` could start a nested spinner. This now matches `upgrade2._create_default_opcua_connector_template`, which already hardcodes `no_status=True` for the same reason.

**Suggestions**

- `azext_edge/edge/providers/orchestration/resources/instances.py` (`update`, the `opcua_backfill_requested` block) — the failure now surfaces, which is the point, but it surfaces as the raw `HttpResponseError` from the `akriConnectorTemplates` list. At that point `begin_create_or_update` has already completed and `result` is discarded, so the user sees a generic service error and has no way to tell that the feature flip itself succeeded and only the template backfill is missing. Consider catching and re-raising with an actionable message — e.g. that instance features were updated but the default OPC UA connector template could not be verified or created, and that re-running `az iot ops update --feature opcua.mode=Stable` will repair it (the block is already written to be idempotent on repeat enable). Same applies to a failure from `create_default_opcua_template` on the line below.
- `azext_edge/tests/edge/orchestration/resources/test_instances_unit.py` (`test_instance_update_opcua_backfill_surfaces_list_error`) — the test asserts only the exception type. The property that actually matters is the ordering: the instance PUT committed and *then* the command failed. Asserting that a `PUT` to the instance endpoint is present in `mocked_responses.calls` before the raise would pin that, and would catch a future refactor that moves the backfill ahead of the instance update (which would change this from "committed but unverified" to "not committed at all").

**Nits**

- `mocked_responses.assert_all_requests_are_fired = False` in `test_instance_update_opcua_backfill` is needed for the `expects_create=False` cases (the `add_callback` PUT is registered but unused), but it applies to the whole test. The explicit `template_puts` assertions carry the weight, so this is fine as written — just worth knowing the safety net is off for the registered instance responses too.

**Summary** — the increment resolves both outstanding inline findings correctly and the error-handling split between `upgrade` and `update` is well reasoned and documented in code. No blockers from my side; the two suggestions are about making the now-surfaced failure actionable for the user and pinning the post-mutation ordering in the test.

digimaun (dismissed):

azext_edge/tests/edge/orchestration/resources/test_instances_unit.py
ketkimnaik: Updated the test

digimaun (approved):

PR #931, opened 2026-09-14
fix: Update topic space name

ketkimnaik:
Update topic space name.

---

### General Guidelines

Intent for Production

Basic expectations


zzcodework:

Reviewing `full diff` (head `3748a0c`, base `feature/live-data`). Two files: the `LIVE_DATA_TOPIC_TEMPLATE` constant changes from `aio/observabilitySessions/{scope_id}/#` to `aio/live-data/{scope_id}/#`, and `TestSetupTopicSpace::test_create_new` is updated to match.

**Blocking**

None.

**Suggestions**

- `azext_edge/edge/providers/orchestration/live_data.py` (`_setup_topic_space`) — the rename does not reach a topic space that already exists. `get_live_data_topic_space_name` hashes the instance ARM resource id, so the name is unchanged by this PR; the `topic_spaces.get(...)` in `_setup_topic_space` therefore succeeds on any environment already enabled, and the method returns early with `{"name": ..., "topicTemplates": topic_templates, "exists": True}` without issuing a PUT. Two consequences worth deciding on explicitly before this merges:
  - the server keeps `aio/observabilitySessions/{scope_id}/#` while the returned payload — which feeds the `Topic space` step and the enable result — reports the new `aio/live-data/...` value, so the CLI claims a template the namespace does not have;
  - anything publishing or subscribing on the new prefix will not match the stale topic space, and the only recovery is `live-data disable` + re-enable (or editing the topic space by hand).

  Since `feature/live-data` only landed in #927 and is unreleased, the blast radius is dev/test environments, which is why I am not calling this blocking — but if that is the reasoning, it is worth stating in the PR description. Otherwise, reconciling the existing topic space (compare `properties.topicTemplates` on the `get` result and PUT when it differs, rather than returning `exists: True` unconditionally) would make the rename actually take effect on upgrade. Note the status path at the `topic_space_section` build already reads `topicTemplates` off the live resource, so `live-data show` and `live-data enable` would disagree with each other in exactly this situation.

**Nits**

- `azext_edge/tests/edge/orchestration/test_live_data_unit.py` (`test_create_new`) — asserting the literal `f"aio/live-data/{scope_id}/#"` instead of re-deriving it from `LIVE_DATA_TOPIC_TEMPLATE` is the right call: the old assertion was tautological and would have passed for any value of the constant. The `instance_name` → `scope_id` rename in that test also reads better given the placeholder name. No change requested; flagging only because it is easy to "helpfully" revert this to the constant in review.
**Summary** — the constant change itself is correct and consistent with the test update. The one thing I would want an explicit answer on is the already-enabled path: `_setup_topic_space` short-circuits on existence, so this rename silently does not apply to environments enabled before it, while the command output says otherwise.

PR #935, opened 2026-09-17
ci: move integration tests to CentralUS

ketkimnaik:
## Summary
- Point integration/container workflows, the container-publisher caller, and scheduled cleanup at `ops-cli-int-test-centralus-rg`.
- Resolve the smoke-test asset-query location from the selected resource group, preserving custom-group inputs.
- Document regional defaults, prerequisites, and cleanup behavior. API versions and test/provider code are unchanged.

## Validation
- YAML/default-consistency checks, mocked smoke tests (default/custom group and failed lookup), and `git diff --check` passed.
- Actionlint adds no new findings; three existing diagnostics remain.
- The Central US test resource group is provisioned and pipeline RG permissions have been verified.
- All seven live integration scenarios: pending; draft until reviewed and validated.

## Operational notes
- Regional workaround for the reported ADR routing issue; this PR does not claim the service issue is fixed.
- Scheduled cleanup deletes resources inside the new group, not the group itself. Leftovers in the old group require separate cleanup.
- No user-facing CLI behavior or API-version change; changelog not applicable.


cheatsheet1999 (approved):

zzcodework:

Reviewing `full diff` (head `680dc7a6`)

Verified the rename is complete and that nothing else pins a region. On `dev`, `ops-cli-int-test-rg` appears in exactly 6 places (`int_test.yml` x3, `container_int_test.yml` x2, `cluster_cleanup.yml` x2, `publish_test_container_image.yml` x1 — 8 lines total), and this PR updates all of them; `grep -rn "ops-cli-int-test" .` finds no remaining reference outside `.github/workflows/`. The claim in the docs that everything inherits the group's location also checks out: `connect-arc/action.yml` calls `az connectedk8s connect` with only `-n/-g`, `setup-cluster-resources/action.yml` creates the storage account, schema registry and ADR namespace with only `-g`, and `deploy-aio/action.yml` only passes `--location` when explicitly supplied — so no hardcoded `westus` survives.

**Blocking**

None found.

**Suggestions**

1. `.github/workflows/publish_test_container_image.yml` line 24 — this passes `resource-group: ops-cli-int-test-centralus-rg` explicitly even though `container_int_test.yml` already declares the same default, so the literal now lives in two files and the next region move has to touch both. Worth considering whether the caller can rely on the callee's default instead (note `workflow_call` marks the input `required: true`, so dropping it needs a quick check that GitHub accepts the default for a required called-workflow input).

2. Operational follow-up for the old group: the PR repoints the scheduled cleanup (`cluster_cleanup.yml` cron `0 8 * * *`) at the new group, so anything still in `ops-cli-int-test-rg` — Arc connected clusters, AIO instances, storage accounts, schema registries — is now never swept and keeps costing. The docs correctly say "Remaining resources in the previous group need separate cleanup"; worth actually doing it once via `workflow_dispatch` on `cluster_cleanup.yml` with `resource_group: ops-cli-int-test-rg` before/after merge, and noting in the PR when it has been drained.

**Nits**

3. `.github/workflows/int_test.yml` line 295 — `resource_group_location=$(az group show --name "$RESOURCE_GROUP" --query location -o tsv)` runs inside the "Run smoke tests" step, i.e. per matrix job and only after the full init + scenario run. If the group does not exist (the workflows deliberately do not provision it) the failure surfaces as an `az group show` error at the very end of a long job rather than up front. Resolving the location once in the `setup` job and passing it down as an output — or asserting the group exists in an early step — would fail fast and save a run. Behaviour as written is otherwise correct: `steps.env_out.outputs.RESOURCE_GROUP` is just `env.RESOURCE_GROUP`, so the queried location matches the group the tests actually use.

4. `docs/integration-tests.md` — the removed blank line after `##### **Scenario Configuration**` is unrelated to the region move; harmless for GFM rendering, but it adds diff noise.

**Summary**

CI-only change: resource-group default rename plus replacing the hardcoded `--location westus` in the asset-query smoke test with the selected group's location. Rename coverage is complete, quoting of `$RESOURCE_GROUP` in the smoke test is an improvement over the previous `${{ env.RESOURCE_GROUP }}` interpolation, and no extension/provider code or API version is touched. No blocking issues; the two suggestions are about de-duplicating the group literal and not orphaning the old West US group.

PR #937, opened 2026-09-22
fix(ci): pin GA ADR integration calls to 2026-04-01

ketkimnaik:
- Pin generic ADR lookups and cleanup to 2026-04-01 to avoid cross-version resource-not-found errors.
- Apply cleanup pinning only to Device Registry resources, leaving other providers and regional defaults unchanged.
- Add 28 offline regression cases covering cleanup routing and the management-action namespace lookup.
- Document why explicit API versioning is required for the GA integration path.

---

### General Guidelines

Intent for Production

Basic expectations


digimaun (approved):

zzcodework:

Reviewing `full diff` (head `42ec520`)

**Blocking**

None.

The core change is sound and I verified the parts that could plausibly break:

- `azext_edge/tests/edge/conftest.py` — the provider extraction `resource_id.lower().rsplit("/providers/", 1)[-1].split("/", 1)[0]` takes the *last* `/providers/` segment, so an extension resource such as `.../Microsoft.DeviceRegistry/namespaces/ns/providers/Microsoft.Authorization/locks/x` resolves to `microsoft.authorization` and is correctly left unpinned. The bash form in `.github/workflows/cluster_cleanup.yml` (`${resource_path##*/providers/}`, longest-prefix removal) has the same semantics, so the two implementations agree — worth calling out because a subtle mismatch here is the usual bug in this pattern.
- `azext_edge/tests/edge/orchestration/test_mgmt_actions_int.py` — `ADR_API_VERSION` is already defined in that module (`ADR_API_VERSION = DEFAULT_DEVICEREGISTRY_MGMT_API_VERSION.value`), so the single-line change does not introduce a `NameError`.
- Pinning cleanup of *all* `Microsoft.DeviceRegistry` types to `2026-04-01` is safe for the legacy types the new tests assert on (`assets`, `assetEndpointProfiles`): the vendored client `azext_edge/edge/vendor/clients/deviceregistrymgmt/v20260401` still exposes those operations, so the GA version is not narrower than the preview surface for these resources. This was my main concern and it does not hold.
- The new file name ends in `_unit.py`, so it is picked up by the tox unit selector (`pytest -k "_unit.py" ./azext_edge/tests`) rather than silently never running.

**Suggestions**

1. `.github/workflows/int_test.yml`, `.github/workflows/container_int_test.yml`, `.github/workflows/cluster_cleanup.yml` — the literal `2026-04-01` is now hardcoded in four places across three workflows, duplicating `DEFAULT_DEVICEREGISTRY_MGMT_API_VERSION` in `azext_edge/edge/util/az_client.py`. The 28 new regression cases lock the Python side to the constant, but nothing ties the YAML to it, so the next bump of `DeviceRegistryMgmtApiVersion` will leave CI cleanup silently on the old version — exactly the version-skew failure mode this PR is fixing. Either derive it once (e.g. a workflow-level `env: ADR_API_VERSION` sourced from `python -c "from azext_edge.edge.util.az_client import DEFAULT_DEVICEREGISTRY_MGMT_API_VERSION; print(...)"` where the extension is already installed) or add an explicit comment at the constant telling the next bumper to grep the workflows.

2. `azext_edge/tests/edge/conftest.py` — the pinned fallback in `_build_delete_command` never executes in CI: the `tracked_resources` fixture returns early when `os.environ.get("CI")` is set ("CI always nukes the entire cluster — skip granular cleanup"). So in the environment this PR is fixing, the behaviour change comes entirely from the workflow edits; the conftest pin only affects local runs. Not a defect — just worth knowing when judging whether the fix covers the failure you saw, and worth a word in `docs/integration-tests.md` so the next reader does not assume conftest is the CI cleanup path.

**Nits**

3. `.github/workflows/int_test.yml` — the new comment ("Match the GA API used to create ADR resources; generic CLI selection may use v2.") sits directly above the `connectedClusters` delete, one line before the `if` block it actually explains. Moving it below that line would avoid implying the cluster delete is also version-sensitive.

4. `.github/workflows/container_int_test.yml` — the file still ends without a trailing newline (the diff shows `\ No newline at end of file` on both sides). Pre-existing, but this PR is already touching that last line.

**Summary**

CI-only change, no product-code or command-surface impact, so no `HISTORY.rst` entry is needed. The pinning logic is correct and consistently implemented in both Python and bash, the legacy ADR types survive the GA pin, and the regression coverage is meaningful (provider routing, casing, extension resources, and the namespace-lookup call sequence). My only substantive point is the hardcoded version literals in YAML drifting from the constant.

*This is an automated review and may be incomplete; please treat findings as input, not as a gate.*

